Issue #202, opened 2015-01-09
app crashes on open


xplat:
app version: 0.5.7-0.5.13 (F-droid, bisected)
platform info: Acer Iconia Tab 8
  Model #: A1-840
  Android version: 4.4.4 (x86)
  UEFI version: 1.0.5
  Kernel version: 3.10,20+
  Image version: Acer_AV0K1_A1-840_RV0ERC02_WW_GEN1
  Build number: Acer_AV0K1_A1-840_1.061.00_WW_GEN1
  CPU: Intel Atom (Bay Trail-T CR 3735F/3735G) 4 cores 1.8GHz
  Screen: 800x1280

If necessary I'll attach logs once I get somewhere I can use ADB.

AudriusButkevicius:
syncthing binary shipped with the app is an arm binary, hence you probably need two do your own packaging to include the right architecture binary, or replace the binary after you've installed the app.

xplat:
Ah, so the culprit for the hard crash would then be https://github.com/syncthing/syncthing-android/commit/df48d75127132f720fa99a78584e24f59bafe925 ?  It's true the syncthing binary always exits with error 1 even with other versions.  I guess that means the binary translation isn't working on syncthing.  Can I replace the binary on a non-rooted device?  If so how do I do it?  And why does the package with arm binary get listed as supporting x86 ABI?

AudriusButkevicius:
I dont think there is any translation happening. The exec call just fails.

Not sure if you are able to replace the binary, never ran stock android.

xplat:
What am I supposed to replace in here?  lib/x86/libsyncthing.so already says it's an x86 binary according to 'file'...

james@nanaka:~/Downloads$ aapt list ~/Downloads/com.nutomic.syncthingandroid_50.apk 
META-INF/MANIFEST.MF
META-INF/21B5F25B.SF
META-INF/21B5F25B.RSA
META-INF/buildserverid
META-INF/fdroidserverid
AndroidManifest.xml
res/anim/abc_fade_in.xml
res/anim/abc_fade_out.xml
res/anim/abc_slide_in_bottom.xml
res/anim/abc_slide_in_top.xml
res/anim/abc_slide_out_bottom.xml
res/anim/abc_slide_out_top.xml
res/color/abc_background_cache_hint_selector_material_dark.xml
res/color/abc_background_cache_hint_selector_material_light.xml
res/color/abc_primary_text_disable_only_material_dark.xml
[...]

PR #207, opened 2015-01-15
Fix i386

Zillode:
Fixes https://github.com/syncthing/syncthing-android/issues/202


Zillode:
updated

Nutomic:
Merged, thanks :)

PR #208, opened 2015-01-15
Fix certificate issue on travis build

Zillode:
We don't need https as we verify the checksum anyway


Zillode:
Cherry-picked by https://github.com/syncthing/syncthing-android/pull/207

PR #220, opened 2015-03-04
Project's owner changed, so the tip4commit link broke

Moter8:
The banner still has the correct image link.


Nutomic:
Thanks, completely missed that :)

PR #223, opened 2015-03-06
Link to the in-use Syncthing org

Moter8:


Nutomic:
Thanks, looks like this is the last incorrect link finally :D

Issue #239, opened 2015-03-14
Devices are not displayed


Nutomic:
JSON parsing is broken in `RestApi#getDevices()` because `Compression` is string instead of boolean now.

Requires settings update so this value can be changed properly.

schuft69:
can confirm this behavior in 0.5.20

PR #246, opened 2015-03-15
Compression is now a string instead of a boolean (fixes #245 and #239)

### LINKED ISSUE #245: version 0.5.0.20 unable to add server with lollipop 5.1
Hi,
I just did a fresh reset of my Nexus 5 with Lollipop 5.1. I installed Syncthing and wanted to add the server. I can scan the server ID and fill in the name but when it want's to restart it does nothing.
I tried to exit the app and try it again but it won't add the server. Then i tried to clear the cache and let it generates new keys but no luck. Still unable to add the server...
The server and Nexus 5 have the latest version of Syncthing.

I now tried Android Version 0.5.18 and that works...
So I manually installed version 0.5.20 and then the devices are disappeared (again)

Zillode:
Took me some time to get the correct preferences without loading the XML. If you load the XML it will crash because the original preference was a booleanpreference, which should become a listpreference.
Using ```SharedPreferences settings = activity.getPreferenceManager().getSharedPreferences();``` is the key ;)

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/fragments/DeviceSettingsFragment.java
+++ b/src/main/java/com/nutomic/syncthingandroid/fragments/DeviceSettingsFragment.java
@@ -8,6 +8,7 @@ import android.content.Intent;
 import android.os.Bundle;
 import android.preference.CheckBoxPreference;
 import android.preference.EditTextPreference;
+import android.preference.ListPreference;
 import android.preference.Preference;
 import android.support.v4.preference.PreferenceFragment;
 import android.view.Menu;
@@ -47,7 +48,7 @@ public class DeviceSettingsFragment extends PreferenceFragment implements
 
     private EditTextPreference mAddresses;
 
-    private CheckBoxPreference mCompression;
+    private ListPreference mCompression;
 
     private CheckBoxPreference mIntroducer;
 
@@ -78,7 +79,7 @@ public class DeviceSettingsFragment extends PreferenceFragment implements
         mName.setOnPreferenceChangeListener(this);
         mAddresses = (EditTextPreference) findPreference("addresses");
         mAddresses.setOnPreferenceChangeListener(this);
-        mCompression = (CheckBoxPreference) findPreference("compression");
+        mCompression = (ListPreference) findPreference("compression");
         mCompression.setOnPreferenceChangeListener(this);
         mIntroducer = (CheckBoxPreference) findPreference("introducer");
         mIntroducer.setOnPreferenceChangeListener(this);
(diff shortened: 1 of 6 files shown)


src/main/res/values/arrays.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string-array name="compress_values">
```
Nutomic: I'd mark this as `translatable=false` just to be sure. Also you forgot a newline at the file end ;)

src/main/java/com/nutomic/syncthingandroid/fragments/DeviceSettingsFragment.java
```diff
@@ -97,13 +103,43 @@ public void onCreate(Bundle savedInstanceState) {
                 mDevice.Name = "";
                 mDevice.DeviceID = "";
                 mDevice.Addresses = "dynamic";
-                mDevice.Compression = true;
+                mDevice.Compression = "always";
```
Nutomic: This is setting the default value. Wouldn't "metadata" be better?

src/main/java/com/nutomic/syncthingandroid/fragments/DeviceSettingsFragment.java
```diff
@@ -97,13 +103,43 @@ public void onCreate(Bundle savedInstanceState) {
...
                 ((EditTextPreference) mDeviceId).setText(mDevice.DeviceID);
             }
         }
     }
 
     /**
+     * Check for any upgrades of the preferences.
+     */
+    private void checkUpgradePreferences(PreferenceFragment activity) {
```
Nutomic: I don't understand why this is needed, the Rest API already returns the new values, not true/false.

And syncthing preferences aren't stored in SharedPreferences.

Nutomic:
Haven't tried if it works yet, will do that tomorrow.

Also, I closed #245 as a duplicate, so you can remove that from your title :)

src/main/java/com/nutomic/syncthingandroid/fragments/DeviceSettingsFragment.java
Zillode: I'm pretty sure reducing communication time saves more battery than reducing CPU load, so I suggest to keep this.

src/main/res/values/arrays.xml
Zillode: Right, I've set ```resources translatable="false"``` but I'm not sure if that works. Google didn't return much information on this:/

src/main/java/com/nutomic/syncthingandroid/fragments/DeviceSettingsFragment.java
Zillode: Hmm, that explains why it took me that long to read the preference value:)
Removed.

src/main/res/values/arrays.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources
```
Nutomic: I actually meant `<resources translatable="false">`, which means this will be ignored by Transifex (and should also ignore the `MissingTranslation` warning).

PR #250, opened 2015-03-16
Set announce color green if at least one is connected (fixes #249)

### LINKED ISSUE #249: Only 1 of 2 announce servers
Only one of two announce servers are available.  Is this related to issue #239?

screenshot_2015-03-16-08-58-10

Syncthing: 0.10.26
Syncthing-Android: 0.5.20

Zillode:
As implemented in upstream, see https://github.com/syncthing/syncthing/commit/f15c416e590607b654c7c7e6408e48af11b39985
Feel free to reject if you don't agree. I'm not sure what is the best option here, but it will at least not scare ppl:)

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/fragments/DrawerFragment.java
+++ b/src/main/java/com/nutomic/syncthingandroid/fragments/DrawerFragment.java
@@ -182,12 +182,11 @@ public class DrawerFragment extends Fragment implements RestApi.OnReceiveSystemI
         });
         mCpuUsage.setText(new DecimalFormat("0.00").format(info.cpuPercent) + "%");
         mRamUsage.setText(RestApi.readableFileSize(mActivity, info.sys));
-        if (info.extAnnounceConnected == info.extAnnounceTotal) {
-            mAnnounceServer.setText(android.R.string.ok);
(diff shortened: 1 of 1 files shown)


src/main/java/com/nutomic/syncthingandroid/fragments/DrawerFragment.java
```diff
@@ -182,12 +182,11 @@ public boolean onTouch(View view, MotionEvent motionEvent) {
         });
         mCpuUsage.setText(new DecimalFormat("0.00").format(info.cpuPercent) + "%");
         mRamUsage.setText(RestApi.readableFileSize(mActivity, info.sys));
-        if (info.extAnnounceConnected == info.extAnnounceTotal) {
-            mAnnounceServer.setText(android.R.string.ok);
+        mAnnounceServer.setText(Integer.toString(info.extAnnounceConnected) + "/" +
+                Integer.toString(info.extAnnounceTotal));
+        if (info.extAnnounceConnected > 1) {
```
Nutomic: Needs to be `>=` :P

Zillode: (omg)

Issue #256, opened 2015-03-18
crash when click "add device"


renyouguo:
Android 4.4.4
0.5.21

chucic:
Did you meant "Add device" ?

In case you did, its java.lang.NullPointerException in DeviceSettingsFragment.java line 77. Same thing will happen when you click on device name to change its settings.

Android 4.4.4
0.5.21

Zillode:
@chucic : both will reach him, but the github issue tracker is preferred I think.
Could you post a log? It works on my device.
Also, is it installed through the play store or f-droid?

chucic:
Through Google Play.
Which log do you mean? My phone is not rooted, so I can't get to normal log file. Or do you mean logcat? Last time I did logcat, I end up with 2k+ lines file... But preview of report to google contains system log, event log and stack trace which starts with java.lang.RuntimeException: Unable ... : java.lang.NullPointerException at ... (30 more lines), so I think its relevant. But seems I'm not able copy+paste from preview, so I send report.

Prestigio 5454
Stock Andriod 4.4.4
Without root

SteffenTux:
I have the same problem. Can't add devices. App promtly crashes, when trying it. Former versions worked without problems.

Nutomic:
Here's the logcat:
```
java.lang.RuntimeException: Unable to start activity ComponentInfo{com.nutomic.syncthingandroid/com.nutomic.syncthingandroid.activities.SettingsActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'void android.preference.Preference.setOnPreferenceChangeListener(android.preference.Preference$OnPreferenceChangeListener)' on a null object reference
at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2329)
at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2391)
at android.app.ActivityThread.access$900(ActivityThread.java:147)
at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1296)
at android.os.Handler.dispatchMessage(Handler.java:102)
at android.os.Looper.loop(Looper.java:135)
at android.app.ActivityThread.main(ActivityThread.java:5256)
at java.lang.reflect.Method.invoke(Native Method)
at java.lang.reflect.Method.invoke(Method.java:372)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:898)
[...]

PR #259, opened 2015-03-19
Upgrade preferences if needed

Zillode:
Seems like the [upgrade](https://github.com/syncthing/syncthing-android/pull/246) was needed after all, see https://github.com/syncthing/syncthing-android/issues/256


Nutomic:
This is *definitely* not the problem (as I said, syncthing settings aren't stored in the app).

I've linked a logcat in #256.

PR #269, opened 2015-03-21
Fix NPE due to invalid mDevice reference

Zillode:
ST updates mDevice in an asynchronous manner. This PR might be too defense, feel free to review...

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/fragments/DeviceSettingsFragment.java
+++ b/src/main/java/com/nutomic/syncthingandroid/fragments/DeviceSettingsFragment.java
@@ -11,6 +11,7 @@ import android.preference.EditTextPreference;
 import android.preference.ListPreference;
 import android.preference.Preference;
 import android.support.v4.preference.PreferenceFragment;
+import android.util.Log;
 import android.view.Menu;
 import android.view.MenuInflater;
 import android.view.MenuItem;
@@ -35,6 +36,8 @@ public class DeviceSettingsFragment extends PreferenceFragment implements
         SyncthingService.OnApiChangeListener, RestApi.OnDeviceIdNormalizedListener {
 
     public static final String EXTRA_NODE_ID = "device_id";
+
+    private static final String TAG = "DeviceSettingsFragment";
 
     private static final int SCAN_QR_REQUEST_CODE = 235;
 
@@ -93,7 +96,8 @@ public class DeviceSettingsFragment extends PreferenceFragment implements
         if (mIsCreate) {
             if (savedInstanceState != null) {
                 mDevice = (RestApi.Device) savedInstanceState.getSerializable("device");
-            } else {
+            }
+            if (mDevice == null) {
                 mDevice = new RestApi.Device();
                 mDevice.Name = "";
                 mDevice.DeviceID = "";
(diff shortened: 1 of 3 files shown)


src/main/java/com/nutomic/syncthingandroid/fragments/DeviceSettingsFragment.java
```diff
@@ -110,7 +115,9 @@ public void onCreate(Bundle savedInstanceState) {
     @Override
     public void onSaveInstanceState(Bundle outState) {
         super.onSaveInstanceState(outState);
-        outState.putSerializable("device", mDevice);
+        if (mIsCreate && mDevice != null) {
+            outState.putSerializable("device", mDevice);
```
Nutomic: I think `Device` needs to implement `Serializable` for this. 

Another option would be functions `Device.save(Bundle bundle)` and `Device.restore(Bundle bundle)` (because I don't know how well serializable works).

src/main/java/com/nutomic/syncthingandroid/fragments/DeviceSettingsFragment.java
```diff
@@ -91,9 +94,11 @@ public void onCreate(Bundle savedInstanceState) {
         }
 
         if (mIsCreate) {
+            mDevice = null;
```
Nutomic: This is entirely pointless, `mDevice` is null anyway (not initialized).

src/main/java/com/nutomic/syncthingandroid/fragments/DeviceSettingsFragment.java
```diff
@@ -110,7 +115,9 @@ public void onCreate(Bundle savedInstanceState) {
     @Override
     public void onSaveInstanceState(Bundle outState) {
         super.onSaveInstanceState(outState);
-        outState.putSerializable("device", mDevice);
+        if (mIsCreate && mDevice != null) {
```
Nutomic: This isn't needed (you can store null, and who cares if we save/restore a device and later replace it in `#onApiChange`?)

src/main/java/com/nutomic/syncthingandroid/fragments/DeviceSettingsFragment.java
```diff
@@ -142,6 +149,14 @@ else if (currentState != SyncthingService.State.ACTIVE) {
             }
             mDeviceId.setOnPreferenceClickListener(this);
         }
+
+        if (mDevice == null) {
+            // The device was removed from Syncthing and no reference was found
+            Log.d(TAG, "Empty mDevice");
+            getActivity().finish();
+            return;
+        }
```
Nutomic: Can you put all of these in a single function that you call everywhere?

Also, it should be `Log.w`, and I'd just move the comment into the log (no need to have comment and log here).

src/main/java/com/nutomic/syncthingandroid/fragments/DeviceSettingsFragment.java
Zillode: this won't restore the settings of a previously edited device, right?

src/main/java/com/nutomic/syncthingandroid/fragments/DeviceSettingsFragment.java
Zillode: This got me thinking, mDevice is updated by a postExecuted task, so I guess this can happen at any time, correct?
We should use a local reference (i.e. device = mDevice) and then check if it is null and use the device reference in the remaining code.

Nutomic: I guess that's right, we don't want a race condition. A lock might be more elegant though.

Nutomic:
We'll also have to apply the same changes in `FolderSettingsFragment`.

src/main/java/com/nutomic/syncthingandroid/fragments/DeviceSettingsFragment.java
Zillode: Also, moving this to a function doesn't help much. We still need to escape/return in the calling function, which involves a similar if test

Zillode:
:+1:

PR #272, opened 2015-03-21
Fix API access for devices < ICS

Zillode:
Found a warning about invalid API access on devices < ICS android versions. This patch fixes that.


Nutomic:
Yeah I saw that warning in Android Studio, but it doesn't make any sense.

[Docs say the function is API level 1](http://developer.android.com/reference/android/preference/TwoStatePreference.html#setChecked%28boolean%29). Also, it's not mentioned by `gradle lint`.

Issue #216, opened 2015-02-22
Shows "Generating keys. This may take a while." and gets stuck there.


digate:
Hi! I'm new here. I've installed Syncthing in my Samsung Galaxy S2 (rooted) today and it shows the message "Generating keys. This may take a while." and doesn't go forward from there, just shows the spinning circle. When I restart the phone and open Syncthing it shows "Waiting for API" and gets stuck again. Should I paste the log here or something ? (In that case, how do I get the log?) Thanks!

calmh:
That may be a long operation. On a slow CPU, minutes probably.

Zillode:
Yes, please post your logs. Read [this](https://github.com/syncthing/syncthing-android/blob/master/CONTRIBUTING.md#bug-reports) to obtain it.

digate:
@calmh Even more than 5 hours? Is it possible?
Here are my logs:

syncthing-android version: 0.5.17
Samsung Galaxy S2 GT-I9100 Android v4.1.2

logcat: http://pastebin.com/nGPK4Rt3
config.xml: http://pastebin.com/VmjNbvTu

AudriusButkevicius:
the fact that you have a config implies that you managed to get the keys generated.

Nutomic:
@digate There's nothing useful in the log. Can you reboot the device, then start the app and get a log?

chucic:
Hi, i run into this exactly same while ago, today found this thread and decided i would try to obtain log.

After some time (provided howto isn't much usefull anymore, since you have to be root to use an app and from terminal it didn't work either for me) I almost give up, I don't need syncthing on android that much. But you know, unresolved issue, so I got back to it. At the end I had to do it from PC with Android SDK.

But at the and, i got my 252313 lines of log, then i searched it and found that keys were generated, GUI started and everything seemed to be working, but syncthing still showed "generating...". Only some complains from proxy > https > adblock > false 127.0.0.1:8080, so i unset proxy (localhost:2020 used by adblock) from my current wifi connection and voila, syncthing app is working.

Even if i turn on proxy again, sycnting is still working, cpu usage changing... And when i shutdown syncthing and start it without disabling proxy (so i got endless "Waiting for API") web gui works normally. Since to adblock in android is quite hard to ad custom filters (and requires root anyway), i have no chance test if i could somehow bypass that.

And to web gui i had to connect to 127.0.0.1:37534, but log still says 127.0.0.1:8080...

Nutomic:
[...]

PR #283, opened 2015-03-22
Streamline SyncthingRunnable

Zillode:
Towards https://github.com/syncthing/syncthing-android/issues/216 ?

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
@@ -27,18 +27,52 @@ public class SyncthingRunnable implements Runnable {
 
     private static final String TAG_NATIVE = "SyncthingNativeCode";
 
+    public static final String UNIT_TEST_PATH = "was running";
+
     private final Context mContext;
 
+    private boolean mGenerate;
+
     private String mCommand;
+
+    public enum Command {
+        generate, // Generate keys, a config file and immediately exit.
+        main,     // Run the main Syncthing application.
+        reset,    // Reset Syncthing's indexes
+    }
 
     /**
      * Constructs instance.
      *
-     * @param command The exact command to be executed on the shell.
+     * @param command Which type of Syncthing command to execute.
      */
-    public SyncthingRunnable(Context context, String command) {
+    public SyncthingRunnable(Context context, Command command) {
         mContext = context;
-        mCommand = command;
+        String syncthing = mContext.getApplicationInfo().dataDir + "/" + SyncthingService.BINARY_NAME;
+        switch (command) {
+            case generate:
+                mCommand = syncthing + " -generate='" + mContext.getFilesDir() + "' -no-browser";
+                break;
+            case main:
+                mCommand = syncthing + " -home='" + mContext.getFilesDir() + "' -no-browser";
+                break;
+            case reset:
+                mCommand = syncthing + " -home='" + mContext.getFilesDir() + "' -reset";
+                break;
+            default:
+                Log.w(TAG, "Unknown command option");
+                mCommand = "";
+        }
+    }
+
+    /**
+     * Constructs instance.
+     *
+     * @param manualCommand The exact command to be executed on the shell. Used for tests only.
+     */
+    public SyncthingRunnable(Context context, String manualCommand) {
+        mContext = context;
+        mCommand = manualCommand;
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
```diff
@@ -24,16 +24,16 @@
...
+    private boolean mGenerate;
 
     /**
      * Constructs instance.
      *
-     * @param command The exact command to be executed on the shell.
+     * @param generate Whether to generate keys, a config file and immediately exit.
      */
-    public SyncthingRunnable(Context context, String command) {
+    public SyncthingRunnable(Context context, boolean generate) {
```
Nutomic: `command` is used in unit tests :/

src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -53,8 +54,8 @@ public ConfigXml(Context context) {
             generateKeysConfig(context);
         }
 
-        // This could cause an infinite loop, maybe we should add a counter, too.
-        do {
+        // Try hard to get the config right
+        for (int i = 0; i < 10; i++) {
```
Nutomic: This makes sense, but should go into a seperate commit/pr.

Zillode:
Bump

Zillode:
test this please

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
```diff
@@ -24,12 +24,24 @@
...
+
     private String mCommand;
 
     /**
      * Constructs instance.
      *
-     * @param command The exact command to be executed on the shell.
+     * @param generate Whether to generate keys, a config file and immediately exit.
+     */
+    public SyncthingRunnable(Context context, boolean generate) {
```
Nutomic: Maybe it would make more sense to pass an `enum command`. Then you could do a switch here and set `mCommand` to the correct one. 

Enum values would be something like generate, run, test.

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
```diff
@@ -22,39 +22,77 @@
...
+                    case main:
+                        dos.writeBytes(command + " -home='" + mContext.getFilesDir() + "' -no-browser");
+                        break;
+                    case reset:
+                        dos.writeBytes(command + " -home='" + mContext.getFilesDir() + "' -reset");
+                        break;
+                    case manual:
+                        dos.writeBytes(mManualCommand);
+                        break;
+                    default:
```
Nutomic: I would move the switch into a static method `SyncthingRunnable(Context, Command)`, then call `this(Context, f(Command))`. Makes everything simpler.

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
```diff
@@ -22,39 +22,77 @@
...
+
+    private Command mCommand;
+
+    private String mManualCommand;
 
+    public enum Command {
+        generate, // Generate keys, a config file and immediately exit.
+        main,     // Run the main Syncthing application.
+        reset,    // Clear the Syncthing index.
+        manual,   // Manual command
```
Nutomic: Do we need `reset`? Deleting app data should be good for that.

Zillode: Deleting app data resets your device id as well. reset only clears the indexes and moves all folders to a backup location

Nutomic: Okay that's right, but have you ever used that? (I haven't) And as long as there's no button for these, we might as well leave it out.

Zillode: Yeah I used it already and it happens sometimes when the db gets corrupted or when the sync never completes. I think we need a button for this in the near future.

Zillode:
@st-jenkins test this please

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
Nutomic: calmh doesn't share your opinion :p 
https://github.com/syncthing/syncthing/pull/1562

Nutomic:
Can you just remove the generate option for now? We can add it when there's an actual GUI item, but right now it's just dead code.

Zillode:
I added an option to reset the database.
Do not merge until upstream resolved https://github.com/syncthing/syncthing/pull/1562. The current behaviour is to also reinitialise the folders which is something we don't want on Android I think.

PR #286, opened 2015-03-22
Catch cases where GetTask returns null (fixes #270).

### LINKED ISSUE #270: NPE in org.json.JSONTokener.nextCleanInternal
```
java.lang.NullPointerException
at org.json.JSONTokener.nextCleanInternal(JSONTokener.java:116)
at org.json.JSONTokener.nextValue(JSONTokener.java:94)
at org.json.JSONObject.<init>(JSONObject.java:155)
at org.json.JSONObject.<init>(JSONObject.java:172)
at com.nutomic.syncthingandroid.syncthing.RestApi$2.onPostExecute(RestApi.java:213)
at com.nutomic.syncthingandroid.syncthing.RestApi$2.onPostExecute(RestApi.java:209)
at android.os.AsyncTask.finish(AsyncTask.java:632)
at android.os.AsyncTask.access$600(AsyncTask.java:177)
at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:645)
at android.os.Handler.dispatchMessage(Handler.java:102)
at android.os.Looper.loop(Looper.java:136)
at android.app.ActivityThread.main(ActivityThread.java:5140)
at java.lang.reflect.Method.invokeNative(Native Method)
at java.lang.reflect.Method.invoke(Method.java:515)
at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:795)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:611)
at dalvik.system.NativeStart.main(Native Method)
```
From Play logs

Zillode:
It sometimes returns [null](https://github.com/syncthing/syncthing-android/blob/fb2b6a21c4190a0ec702aa1f6424a3c6547284e5/src/main/java/com/nutomic/syncthingandroid/syncthing/GetTask.java#L81)

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/GetTask.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/GetTask.java
@@ -49,34 +49,42 @@ public class GetTask extends AsyncTask<String, Void, String> {
      */
     @Override
     protected String doInBackground(String... params) {
-        String fullUri = params[0] + params[1];
-        HttpClient httpclient = new DefaultHttpClient();
-        if (params.length == 5) {
-            LinkedList<NameValuePair> urlParams = new LinkedList<>();
-            urlParams.add(new BasicNameValuePair(params[3], params[4]));
-            fullUri += "?" + URLEncodedUtils.format(urlParams, HTTP.UTF_8);
-        }
-        HttpGet get = new HttpGet(fullUri);
-        get.addHeader(new BasicHeader(RestApi.HEADER_API_KEY, params[2]));
-
-        try {
-            HttpResponse response = httpclient.execute(get);
-            HttpEntity entity = response.getEntity();
-
-            if (entity != null) {
-                InputStream is = entity.getContent();
-
-                BufferedReader br = new BufferedReader(new InputStreamReader(is));
-                String line;
-                String result = "";
-                while ((line = br.readLine()) != null) {
-                    result += line;
-                }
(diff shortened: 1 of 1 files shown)


src/main/java/com/nutomic/syncthingandroid/syncthing/RestApi.java
```diff
@@ -204,6 +204,10 @@ public void onWebGuiAvailable() {
         new GetTask() {
             @Override
             protected void onPostExecute(String s) {
+                if (s == null) {
+                    Log.w(TAG, "Empty JSON string when retrieving version information");
```
Nutomic: This just postpones the crash. As soon as the value is accessed, there's an NPE again.

Zillode: Are you sure? This is an async task, so the current values should already be valid

Nutomic: It is only run once on first start, and before that, the values are all null.

Nutomic: So I'd say get rid of the changes in this file, and just change GetTask (which I haven't reviewed yet).

Zillode: ok, we can check again if there are still error reports coming in from this point

PR #303, opened 2015-03-25
Fix back arrow in actionbar for Android 4.1+

Zillode:


src/main/AndroidManifest.xml
```diff
@@ -31,12 +31,17 @@
         </activity>
         <activity
             android:name=".activities.WebGuiActivity"
-            android:label="@string/web_gui_title">
+            android:label="@string/web_gui_title"
+            android:parentActivityName=".activities.MainActivity">
+            <!-- Parent activity meta-data to support 4.0 and lower -->
```
Nutomic: I don't think the comment is really needed.

PR #304, opened 2015-03-25
Update welcoming text

Zillode:
Also, 'API' is too complex and 'securing connection' gives a good feeling (and it'll be fully true if we enable TLS):)

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -3,7 +3,7 @@
 
     <string name="app_name">Syncthing</string>
 
-    <string name="app_description">Replace proprietary services with something open, trustworthy and decentralized.</string>
+    <string name="app_description">An open, trustworthy and decentralized file synchronization application.</string>
 
     <!-- MainActivity -->
 
@@ -18,7 +18,7 @@
     <string name="send_device_id_to">Send Device ID to</string>
 
     <!-- Text for FoldersFragment and DevicesFragment loading view -->
-    <string name="api_loading">Waiting for API</string>
+    <string name="api_loading">Loading&#8230;</string>
 
     <!-- FoldersFragment -->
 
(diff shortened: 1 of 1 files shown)


src/main/res/values/strings.xml
```diff
@@ -18,7 +18,7 @@
     <string name="send_device_id_to">Send Device ID to</string>
 
     <!-- Text for FoldersFragment and DevicesFragment loading view -->
-    <string name="api_loading">Waiting for API</string>
+    <string name="api_loading">Securing Connection</string>
```
Nutomic: I don't like this, it's not really what we are doing.

src/main/res/values/strings.xml
```diff
@@ -3,7 +3,7 @@
 
     <string name="app_name">Syncthing</string>
 
-    <string name="app_description">Replace proprietary services with something open, trustworthy and decentralized.</string>
+    <string name="app_description">An open, trustworthy and decentralized file synchronization application.</string>
```
Nutomic: I took this from [syncthing.net](https://syncthing.net/), do you really think we should have a different text?

src/main/res/values/strings.xml
```diff
@@ -194,14 +194,14 @@ Please install a scanner or enter the device ID manually.</string>
     <string name="web_gui_loading">Waiting for GUI</string>
 
     <!-- Shown instead of web_gui_loading if the key does not exist and has to be created -->
-    <string name="web_gui_creating_key">Generating keys. This may take a while.</string>
+    <string name="web_gui_creating_key">Generating secure keys. This may take up to a few minutes.</string>
```
Nutomic: "This may take a few minutes" seems more concise.

src/main/res/values/strings.xml
Zillode: Yeah, I think we moved beyond the 'replace' and set our own goals

src/main/res/values/strings.xml
Zillode: Any suggestion? Should be as noob friendly as possible imo.
Loading?

Zillode:
Bumped

src/main/res/values/strings.xml
Nutomic: Yeah loading seems fine.

Nutomic:
So I'd also merge this for 0.6 (as is).

Zillode:
@st-jenkins test this please

src/main/res/values/strings.xml
```diff
@@ -18,7 +18,7 @@
     <string name="send_device_id_to">Send Device ID to</string>
 
     <!-- Text for FoldersFragment and DevicesFragment loading view -->
-    <string name="api_loading">Waiting for API</string>
+    <string name="api_loading">Loading...</string>
```
Nutomic: Can you use an [Ellipsis](https://en.wikipedia.org/wiki/Ellipsis) here to fix that lint warning?

Nutomic:
Merged in c3d4a1cfdc6a9219577d3479e575583b49dc8024

PR #310, opened 2015-03-26
Set user pw on install

Nutomic:
@Zillode Should we merge this in master, and make an 0.5 branch? Or wait with the merging until we almost have 0.6? We need to keep some stuff back so 0.6 is more impressive :D

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java
+++ b/src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java
@@ -1,5 +1,6 @@
 package com.nutomic.syncthingandroid.fragments;
 
+import android.content.SharedPreferences;
 import android.content.pm.PackageManager;
 import android.os.Bundle;
 import android.preference.CheckBoxPreference;
@@ -27,11 +28,16 @@ public class SettingsFragment extends PreferenceFragment
     private static final String TAG = "SettingsFragment";
 
     private static final String SYNCTHING_OPTIONS_KEY = "syncthing_options";
-    private static final String SYNCTHING_GUI_KEY = "syncthing_gui";
-    private static final String DEVICE_NAME_KEY = "DeviceName";
+    private static final String SYNCTHING_GUI_KEY     = "syncthing_gui";
+    private static final String DEVICE_NAME_KEY       = "DeviceName";
     private static final String USAGE_REPORT_ACCEPTED = "URAccepted";
-    private static final String EXPORT_CONFIG = "export_config";
-    private static final String IMPORT_CONFIG = "import_config";
+    private static final String ADDRESS               = "Address";
+    private static final String GUI_USER              = "gui_user";
+    private static final String GUI_PASSWORD          = "gui_password";
+    private static final String USER_TLS              = "UseTLS";
+    private static final String EXPORT_CONFIG         = "export_config";
+    private static final String IMPORT_CONFIG         = "import_config";
+    private static final String STTRACE               = "sttrace";
 
     private static final String SYNCTHING_VERSION_KEY = "syncthing_version";
 
@@ -72,18 +78,15 @@ public class SettingsFragment extends PreferenceFragment
                         break;
                     default:
                         value = api.getValue(RestApi.TYPE_OPTIONS, pref.getKey());
-
-
                 }
                 applyPreference(pref, value);
             }
 
-            for (int i = 0; i < mGuiScreen.getPreferenceCount(); i++) {
-                Preference pref = mGuiScreen.getPreference(i);
-                pref.setOnPreferenceChangeListener(SettingsFragment.this);
-                String value = api.getValue(RestApi.TYPE_GUI, pref.getKey());
-                applyPreference(pref, value);
-            }
+            Preference address = mGuiScreen.findPreference(ADDRESS);
+            applyPreference(address, api.getValue(RestApi.TYPE_GUI, ADDRESS));
+
+            Preference tls = mGuiScreen.findPreference(USER_TLS);
+            applyPreference(tls, api.getValue(RestApi.TYPE_GUI, USER_TLS));
         }
     }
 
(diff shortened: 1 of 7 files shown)


Zillode:
lol:)
I think if the TLS stuff is finished we have all the features we need for 0.6:P
Although a logging activity with EditText or something similar would be v useful as well

Nutomic:
I'm just asking because all the rebasing was a huge PITA for me for 0.5

Zillode:
Included in https://github.com/syncthing/syncthing-android/pull/317

Nutomic:
I pass the user:pw by environment variable now, this should be safe.

src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java
```diff
@@ -27,11 +28,16 @@
...
-    private static final String DEVICE_NAME_KEY = "DeviceName";
+    private static final String SYNCTHING_GUI_KEY     = "syncthing_gui";
+    private static final String DEVICE_NAME_KEY       = "DeviceName";
     private static final String USAGE_REPORT_ACCEPTED = "URAccepted";
-    private static final String EXPORT_CONFIG = "export_config";
-    private static final String IMPORT_CONFIG = "import_config";
+    private static final String ADDRESS               = "address";
+    private static final String GUI_USER              = "gui_user";
+    private static final String GUI_PASSWORD          = "gui_password";
+    private static final String USER_TLS              = "UseTLS";
```
Zillode: don't need this when #339 is applied

src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java
```diff
@@ -72,18 +78,15 @@ public void onApiChange(SyncthingService.State currentState) {
...
-            for (int i = 0; i < mGuiScreen.getPreferenceCount(); i++) {
-                Preference pref = mGuiScreen.getPreference(i);
-                pref.setOnPreferenceChangeListener(SettingsFragment.this);
-                String value = api.getValue(RestApi.TYPE_GUI, pref.getKey());
-                applyPreference(pref, value);
-            }
+            Preference address = mGuiScreen.findPreference(ADDRESS);
+            applyPreference(address, api.getValue(RestApi.TYPE_GUI, ADDRESS));
+
+            Preference tls = mGuiScreen.findPreference(USER_TLS);
```
Zillode: nor this

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
```diff
@@ -261,6 +264,25 @@ public boolean accept(File dir, String name) {
...
     @Override
     public void onCreate() {
+        PRNGFixes.apply();
+        SharedPreferences sp = PreferenceManager.getDefaultSharedPreferences(this);
+
+        // Make sure this is also done for existing installs. We can replace this check with
+        // {@link #isFirstStart()} after a while.
+        if (!sp.getBoolean("default_user_pw_set", false)) {
+            sp.edit().putBoolean("default_user_pw_set", true).commit();
+            char[] chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789".toCharArray();
```
Zillode: Should include lowercase and some special symbols

src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java
Nutomic: The `UseTLS` setting? If we take it out here, it's more conflicts again^^

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
Nutomic: I'm adding lowercase, but not sure which special characters we should use. It should be save enough anyway.

Nutomic:
@Zillode I've added a-z to the pw, should be good now.

Nutomic:
One more thing, don't merge yet :D

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
```diff
@@ -53,9 +53,11 @@ public void run() {
                 dos = new DataOutputStream(process.getOutputStream());
                 // Set home directory to data folder for syncthing to use.
                 dos.writeBytes("HOME=" + Environment.getExternalStorageDirectory() + " ");
-                dos.writeBytes("STTRACE=" + pm.getString("sttrace", "") + " ");
+                dos.writeBytes("STTRACE=" + sp.getString("sttrace", "") + " ");
                 dos.writeBytes("STNORESTART=1 ");
                 dos.writeBytes("STNOUPGRADE=1 ");
+                dos.writeBytes("STGUIAUTH='" + sp.getString("gui_user", "") + ":" +
```
Zillode: As this is sensitive information, it should be moved to the Runtime.exec function, see http://docs.oracle.com/javase/6/docs/api/java/lang/Runtime.html#exec%28java.lang.String,%20java.lang.String[]%29

Nutomic:
Finally :D

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
```diff
@@ -42,20 +43,25 @@ public SyncthingRunnable(Context context, String command) {
...
-        SharedPreferences pm = PreferenceManager.getDefaultSharedPreferences(mContext);
+        SharedPreferences sp = PreferenceManager.getDefaultSharedPreferences(mContext);
         DataOutputStream dos = null;
         int ret = 1;
         Process process = null;
         try {
             // Loop to handle syncthing restarts (these always have an error code of 3).
             do {
-                process = Runtime.getRuntime().exec("sh");
+                ProcessBuilder pb = new ProcessBuilder("sh");
```
Nutomic: @Zillode I tried to pass `mCommand` here, but that broke config generation. Maybe you can get this working in #283?

PR #314, opened 2015-03-27
Update default Syncthing port

Zillode:
Ref https://github.com/syncthing/syncthing/commit/c67e2c2a5a7b60ef727cca2f4f90bd558f559c3b


Nutomic:
That config_default file seems unused, you can just delete it.

Zillode:
:+1:

PR #317, opened 2015-03-27
Enable TLS

Zillode:
Works:
- All APIs (poller, get, post)

Broken:
- webview (I think we can fix it)

Question:
- Are user/password changes immediately effective or do we need to restart ST?


Nutomic:
WebView doesn't support certificate pinning [Android issue](https://code.google.com/p/android/issues/detail?id=76501).

User/password are passed via command line, so they only change on restart.

(haven't reviewed it yet)

PR #319, opened 2015-03-28
Add log window (fixes #311).

### LINKED ISSUE #311: Show Syncthing log
Let the user see the output of Syncthing in a window (with text that can be copy/pasted and shared).

Zillode:
Adds a Syncthing log and an ADB logcat window. There is also a share button.

Not sure about ```LogActivity.logger``` but referencing from LogActivity to a SyncthingRunnable was easier this way.

Feel free to comment

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -25,7 +25,6 @@
             android:launchMode="singleTop">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
-
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
@@ -43,6 +42,14 @@
             <meta-data
                 android:name="android.support.PARENT_ACTIVITY"
                 android:value=".activities.MainActivity" />
+        </activity>
+        <activity
+            android:name=".activities.LogActivity"
+            android:label="@string/log_title"
+            android:parentActivityName=".activities.SettingsActivity">
+            <meta-data
+                android:name="android.support.PARENT_ACTIVITY"
+                android:value=".activities.SettingsActivity" />
         </activity>
 
         <service android:name=".syncthing.SyncthingService" />
--- a/src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
+++ b/src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
@@ -1,0 +1,161 @@
+package com.nutomic.syncthingandroid.activities;
+
+import android.content.Intent;
+import android.os.AsyncTask;
+import android.os.Bundle;
+import android.support.v7.widget.ShareActionProvider;
+import android.util.Log;
+import android.view.Menu;
+import android.view.MenuInflater;
+import android.view.MenuItem;
+import android.widget.EditText;
+
+import android.support.v4.view.MenuItemCompat;
+import android.widget.ScrollView;
+import android.widget.TextView;
+
+import com.nutomic.syncthingandroid.R;
+
+import java.io.BufferedReader;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+/**
+ * Shows the log information from Syncthing.
+ */
+public class LogActivity extends SyncthingActivity {
+
+    private final static String TAG = "LogActivity";
+
+    private TextView mLog;
+    private boolean mSyncthingLog = true;
+    private AsyncTask mFetchLogTask;
+    private ScrollView mScrollView;
+    private Intent mShareIntent;
+
+    /**
+     * Initialize Log.
+     */
+    @Override
+    public void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        setContentView(R.layout.log_activity);
+        getSupportActionBar().setDisplayHomeAsUpEnabled(true);
+
+        if (savedInstanceState != null) {
+            mSyncthingLog = savedInstanceState.getBoolean("syncthingLog");
+            invalidateOptionsMenu();
+        }
+
+        mLog = (TextView) findViewById(R.id.log);
+        mScrollView = (ScrollView) findViewById(R.id.scroller);
+
+        updateLog();
+    }
+
+    @Override
+    protected void onSaveInstanceState(Bundle outState) {
+        super.onSaveInstanceState(outState);
+        outState.putBoolean("syncthingLog", mSyncthingLog);
+    }
+
+    @Override
+    public boolean onCreateOptionsMenu(Menu menu) {
+        MenuInflater inflater = getMenuInflater();
+        inflater.inflate(R.menu.log_list, menu);
+
+        // Add the share button
+        MenuItem shareItem = menu.findItem(R.id.menu_share);
+        shareItem.setTitle(mSyncthingLog ? R.string.log_android_title : R.string.log_syncthing_title);
+        ShareActionProvider actionProvider = (ShareActionProvider) MenuItemCompat.getActionProvider(shareItem);
+        mShareIntent = new Intent();
+        mShareIntent.setAction(Intent.ACTION_SEND);
+        mShareIntent.setType("text/plain");
+        mShareIntent.putExtra(android.content.Intent.EXTRA_TEXT, mLog.getText());
+        actionProvider.setShareIntent(mShareIntent);
+
+        return true;
+    }
+
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        switch (item.getItemId()) {
+            case R.id.switch_logs:
+                mSyncthingLog = !mSyncthingLog;
+                item.setTitle(mSyncthingLog ? R.string.log_android_title : R.string.log_syncthing_title);
+                updateLog();
+                return true;
+            default:
+                return super.onOptionsItemSelected(item);
+        }
+    }
+
+    private void scrollToBottom() {
+        mScrollView.post(new Runnable() {
+            @Override
+            public void run() {
+                mScrollView.scrollTo(0, mLog.getBottom());
+            }
+        });
+    }
+
+    private void updateLog() {
+        if (mFetchLogTask != null)
+            mFetchLogTask.cancel(true);
+        mLog.setText("Retrieving logs...");
+        mFetchLogTask = new AsyncTask<Void, Void, String>() {
+            @Override
+            protected String doInBackground(Void... params) {
+                return getLog(mSyncthingLog);
+            }
+            @Override
+            protected void onPostExecute(String log) {
+                mLog.setText(log);
+                if (mShareIntent != null)
+                    mShareIntent.putExtra(android.content.Intent.EXTRA_TEXT, log);
+                scrollToBottom();
+            }
+        }.execute();
+    }
+
+    private String getLog(final boolean syncthingLog) {
+        Process process = null;
+        DataOutputStream pOut = null;
+        try {
+            ProcessBuilder pb;
+            if (syncthingLog) {
+                pb = new ProcessBuilder("/system/bin/logcat", "-t", "300", "-s", "SyncthingNativeCode");
+            } else {
+                pb = new ProcessBuilder("/system/bin/logcat", "-t", "300", "'*'");
+            }
+            pb.redirectErrorStream(true);
+            process = pb.start();
+            BufferedReader bufferedReader = new BufferedReader(
+                    new InputStreamReader(process.getInputStream()), 8192);
+            StringBuilder log = new StringBuilder();
+            String line = "";
+            while ((line = bufferedReader.readLine()) != null) {
+                log.append(line);
+                log.append(System.getProperty("line.separator"));
+            }
+            return log.toString();
+        } catch (IOException e) {
+            Log.w(TAG, "Error reading Android log", e);
+        } finally {
+            try {
+                if (pOut != null) {
+                    pOut.close();
+                }
+            } catch (IOException e) {
+                Log.w(TAG, "Failed to close shell stream", e);
+            }
+            if (process != null) {
+                process.destroy();
+            }
+        }
+        return "";
+    }
+
+}
--- a/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
+++ b/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
@@ -43,7 +43,7 @@ public class SettingsActivity extends SyncthingActivity {
         if (savedInstanceState != null) {
             mFragment = fm.getFragment(savedInstanceState,
                     savedInstanceState.getString("fragment_name"));
-        } else {
+        } else if (getIntent().getAction() != null) {
             switch (getIntent().getAction()) {
                 case ACTION_APP_SETTINGS_FRAGMENT:
                     setTitle(R.string.settings_title);
@@ -65,6 +65,9 @@ public class SettingsActivity extends SyncthingActivity {
                     throw new IllegalArgumentException(
                             "You must provide the requested fragment type as an extra.");
             }
+        } else{
+            setTitle(R.string.settings_title);
+            mFragment = new SettingsFragment();
         }
 
         fm.beginTransaction()
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
@@ -9,7 +9,6 @@ import android.util.Log;
 
 import java.io.BufferedReader;
 import java.io.DataOutputStream;
-import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
@@ -21,7 +20,7 @@ public class SyncthingRunnable implements Runnable {
 
     private static final String TAG = "SyncthingRunnable";
 
-    private static final String TAG_IONICE = "SyncthingRunnable-IoNice";
+    private static final String TAG_IONICE = "SyncthingRunnableIoNice";
 
     private static final String TAG_NATIVE = "SyncthingNativeCode";
 
@@ -134,6 +133,7 @@ public class SyncthingRunnable implements Runnable {
      * Logs the outputs of a stream to logcat and mNativeLog.
      *
      * @param is The stream to log.
+     * @param priority The priority level.
      */
     private void log(final InputStream is, final int priority) {
         new Thread(new Runnable() {
@@ -149,7 +149,7 @@ public class SyncthingRunnable implements Runnable {
                 } catch (IOException e) {
                     // NOTE: This is sometimes called on shutdown, as
                     // Process.destroy() closes the stream.
-                    Log.w(TAG, "Failed to read syncthing command line output", e);
+                    Log.w(TAG, "Failed to read Syncthing's command line output", e);
                 }
             }
         }).start();
--- a/src/main/res/layout/log_activity.xml
+++ b/src/main/res/layout/log_activity.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <ScrollView
+        android:id="@+id/scroller"
+        android:layout_width="fill_parent"
+        android:layout_height="fill_parent"
+        android:scrollbars="vertical"
+        android:fillViewport="true">
+
+        <TextView
+            android:id="@+id/log"
+            android:gravity="top"
+            android:textSize="10sp"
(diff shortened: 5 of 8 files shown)


Zillode:
(also logging to a file would be hard as it is written and read at the same time).

src/main/res/layout/log_activity.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <EditText
```
Nutomic: Any reason you're using an EditText? This shows autocomplete errors and lets the user change stuff. TextView should be fine. Also, text is weirdly centered, you should probably use `gravity=top`.

src/main/res/menu/log_list.xml
```diff
@@ -0,0 +1,16 @@
...
+
+    <item
+        android:id="@+id/switch_logs"
+        android:title="@string/log_android_title"
+        app:showAsAction="always" />
+
+    <item
+        android:id="@+id/share"
+        android:title="@string/share_title"
+        app:showAsAction="always" />
```
Nutomic: Use the *share* icon ;)

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
```diff
@@ -0,0 +1,70 @@
...
+    public void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        setContentView(R.layout.log_activity);
+        getSupportActionBar().setDisplayHomeAsUpEnabled(true);
+
+        mLog = (EditText) findViewById(R.id.log);
+        mLog.setText(logger.getLog());
+    }
+
```
Nutomic: Needs to implement `onSaveInstanceState()` and save/restore android/native log state.

src/main/java/com/nutomic/syncthingandroid/util/Logger.java
```diff
@@ -0,0 +1,54 @@
...
+        while (it.hasNext()) {
+            sb.append(it.next());
+            sb.append("\n");
+        }
+        return sb.toString();
+    }
+
+    public String getAndroidLog() {
+        try {
+            Process process = Runtime.getRuntime().exec("logcat -d '*:d'");
```
Nutomic: I'd do `SyncthingNativeCode:s`, or we have that twice. Though it may also be useful, not sure.

Also, comment your class/methods :p

src/main/java/com/nutomic/syncthingandroid/util/Logger.java
```diff
@@ -0,0 +1,54 @@
...
+                    new InputStreamReader(process.getInputStream()));
+            StringBuilder log = new StringBuilder();
+            String line = "";
+            while ((line = bufferedReader.readLine()) != null) {
+                log.append(line);
+                log.append("\n");
+            }
+            return log.toString();
+        }
+        catch (IOException e) {}
```
Nutomic: Log this please :)

src/main/java/com/nutomic/syncthingandroid/util/Logger.java
```diff
@@ -0,0 +1,54 @@
...
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.util.Iterator;
+import java.util.LinkedList;
+
+public class Logger {
+
+    private final static String TAG = "Logger";
```
Nutomic: I'd put this in `SyncthingService`, pass it to `SyncthingRunnable` in constructor, and provide a getter method (probably just returning the entire log string).

`getAndroidLog()` can go into the Activity imo.

Nutomic:
Actually, this should be way easier: Just run `logcat -s SyncthingNativeCode:v` for the native logging, and let Android handle the storage etc. Maybe do a string.replace() on the log tags.

src/main/res/menu/log_list.xml
```diff
@@ -0,0 +1,17 @@
...
+    <item
+        android:id="@+id/switch_logs"
+        android:title="@string/log_android_title"
+        bwq:showAsAction="always" />
+
+      <item
+        android:id="@+id/menu_share"
+        android:title="@string/share_title"
+        bwq:actionProviderClass="android.support.v7.widget.ShareActionProvider"
+        bwq:showAsAction="always" />
```
Nutomic: What does `bwq` stand for? I've just used `app` in the other menus.

Also, what does the action provider do exactly?

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
```diff
@@ -7,6 +7,8 @@
 import android.preference.PreferenceManager;
 import android.util.Log;
 
+import com.nutomic.syncthingandroid.activities.LogActivity;
```
Nutomic: Ctrl + Alt + O to optimize imports ;)

Nutomic:
From testing, the toggle seems a bit slow, you should use an AsyncTask. Also, there's still no saveInstanceState. Scrollbar would be nice if possible. And parent activity is wrong, should be SettingsActivity.

I do like the idea :)

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
```diff
@@ -0,0 +1,205 @@
...
+    private void scrollToBottom() {
+        mScrollView.post(new Runnable() {
+            @Override
+            public void run() {
+                mScrollView.scrollTo(0, mLog.getBottom());
+            }
+        });
+    }
+
+    private void updateLog() {
```
Nutomic: No need for a function, just make an inner class `UpdateLogTask` (like in `SyncthinkService`). 

Also, `getSyncthingLog()` and `getAndroidLog()` seem to have quite some duplicate code. I would just put that in `doInBackground()` and make an if to choose the logcat command/parameters.

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
```diff
@@ -0,0 +1,205 @@
...
+     */
+    @Override
+    public void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        setContentView(R.layout.log_activity);
+        getSupportActionBar().setDisplayHomeAsUpEnabled(true);
+
+        if (savedInstanceState != null) {
+            syncthingLog = savedInstanceState.getBoolean("syncthingLog");
```
Nutomic: You also need to update the text of the menu button here.

src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -65,6 +65,9 @@ protected void onCreate(Bundle savedInstanceState) {
                     throw new IllegalArgumentException(
                             "You must provide the requested fragment type as an extra.");
             }
+        } else {
+            setTitle(R.string.settings_title);
+            mFragment = new SettingsFragment();
```
Nutomic: What are the changes in this file about? They're definitely not needed for the LogActivity.

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
```diff
@@ -0,0 +1,205 @@
...
+        if (fetchLogTask != null)
+            fetchLogTask.cancel(true);
+        mLog.setText("Retrieving logs...");
+        fetchLogTask = new AsyncTask<Void, Void, String>() {
+            @Override
+            protected String doInBackground(Void... params) {
+                if (syncthingLog) {
+                    return getSyncthingLog();
+                } else {
+                    return getAndroidLog();
```
Nutomic: `getAndroidLog()` seems to be stuck loading forever. Did you forget a parameter or something? (should be fixed by the above anyway).

Zillode:
@st-jenkins test this please

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
Zillode: - k
- Not sure, there are some subtle differences, like L135, L141-142 and L144-147, I don't think adding three if in one complex method would help here.

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
Zillode: Not sure why it is that slow sometimes

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
Zillode: Hmm, on a second thought, a function allows me to automatically cancel the running task. If I extend from AsyncTask I need to duplicate code which could be easily forgotten.

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
```diff
@@ -0,0 +1,205 @@
...
+            pOut = new DataOutputStream(process.getOutputStream());
+            BufferedReader bufferedReader = new BufferedReader(
+                    new InputStreamReader(process.getInputStream()));
+            pOut.writeBytes("logcat -d -t 500 '*:s' 'SyncthingNativeCode:d'\n");
+            pOut.writeBytes("exit\n");
+            pOut.flush();
+            process.waitFor();
+            StringBuilder log = new StringBuilder();
+            String line = "";
+            bufferedReader.readLine(); // Skip "--------- beginning of system"
```
Nutomic: I wouldn't even do this, no one cares.

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
```diff
@@ -0,0 +1,168 @@
...
+     */
+    @Override
+    public void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+
+        setContentView(R.layout.log_activity);
+        getSupportActionBar().setDisplayHomeAsUpEnabled(true);
+
+        if (savedInstanceState != null) {
+            mSyncthingLog = savedInstanceState.getBoolean("syncthingLog");
```
Nutomic: You still need to restore the correct button text.

Zillode: No, there is no access to the 'item' here and onCreateOptionsMenu is automatically called afterwards, which sets the correct title

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
```diff
@@ -0,0 +1,168 @@
...
+    private String getLog(final boolean syncthingLog) {
+        Process process = null;
+        DataOutputStream pOut = null;
+        try {
+            ProcessBuilder pb;
+            if (syncthingLog) {
+                pb = new ProcessBuilder("/system/bin/logcat", "-t", "300", "-s", "SyncthingNativeCode");
+            } else {
+                pb = new ProcessBuilder("/system/bin/logcat", "-t", "300", "'*'");
+            }
```
Nutomic: The logcat path should be a constant. Also, this is the perfect use for ternary operator imo.

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
```diff
@@ -0,0 +1,168 @@
...
+            protected void onPostExecute(String log) {
+                mLog.setText(log);
+                if (mShareIntent != null)
+                    mShareIntent.putExtra(android.content.Intent.EXTRA_TEXT, log);
+                scrollToBottom();
+            }
+        }.execute();
+    }
+
+    private String getLog(final boolean syncthingLog) {
```
Nutomic: Docstring please :p

src/main/res/values/strings.xml
```diff
@@ -314,6 +320,25 @@ Please report any problems you encounter via Github.</string>
...
+    <string name="log_title">Log</string>
+
+    <!-- Title of the "log android" menu button -->
+    <string name="log_android_title">View Android Log</string>
+
+    <!-- Title of the "log Syncthing" menu button -->
+    <string name="log_syncthing_title">View Syncthing Log</string>
+
+    <!-- Title of the "share log" menu action -->
+    <string name="share_log_title">Share log</string>
```
Nutomic: Log should be upper case :D

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
```diff
@@ -0,0 +1,168 @@
...
+
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        switch (item.getItemId()) {
+            case R.id.switch_logs:
+                mSyncthingLog = !mSyncthingLog;
+                if (mSyncthingLog) {
+                    item.setTitle(R.string.log_android_title);
+                } else {
+                    item.setTitle(R.string.log_syncthing_title);
```
Nutomic: Also works with a ternary operator: `setTitle((cond) ? title1 : title2)`

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
Nutomic: Have you tried that? On my phone, I definitely get the wrong text.

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
Zillode: Is this a no go without ternary? Golang devs decided not to support ternary operators because they (easily) introduce unreadable/broken code. After spending some time in Go, I actually like their reasoning, keep it super simple and stupid, without fancyness

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
Zillode: returns the log as a string?
This is simply duplicating the function definition: "String" "getLog"

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
Zillode: Yes, I added L79 for this

Nutomic: `Intent.putExtra()`? All you need to do is call `invalidateOptionsMenu()`.

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
Nutomic: Hm that's right, it's just that the method is a bit long at first glance. Can you make the AsyncTask an inner class instead?

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
Nutomic: Well it does help avoid some code duplication, but I can live without it,

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
Zillode: No because you couldn't cancel mFetchLogTask in that manner.

src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
Zillode: Yes they are, because Log is an activity and has a 'return' behavior to go to SettingsActivity. Therefore it will be started with an empty action intent.

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
Nutomic: You can still assign it to a variable (what I mean is `private class MyAsyncTask extends AsyncTask`).

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
```diff
@@ -0,0 +1,161 @@
...
+            }
+            pb.redirectErrorStream(true);
+            process = pb.start();
+            BufferedReader bufferedReader = new BufferedReader(
+                    new InputStreamReader(process.getInputStream()), 8192);
+            StringBuilder log = new StringBuilder();
+            String line = "";
+            while ((line = bufferedReader.readLine()) != null) {
+                log.append(line);
+                log.append(System.getProperty("line.separator"));
```
Nutomic: Is there any reason you can't use `\n`? :o

src/main/AndroidManifest.xml
```diff
@@ -44,6 +44,18 @@
...
                 android:value=".activities.MainActivity" />
         </activity>
+        <activity
+            android:name=".activities.LogActivity"
+            android:label="@string/log_title"
+            android:parentActivityName=".activities.SettingsActivity">
+            <intent-filter>
+                <action android:name=".activities.LogActivity" />
+                <category android:name="android.intent.category.DEFAULT" />
+            </intent-filter>
```
Nutomic: The intent filter shouldn't be needed.

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
Zillode: This whole class is about logging and you want to add another class?
This is overkill and adds an extra level of indirection (vtable).
You should prepare for many more code like this with the introduction of lambdas in Java 8.

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
Zillode: http://stackoverflow.com/questions/12987451/system-getpropertyline-separator-vs-n-for-android

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
Nutomic: We're optimizing for humans, not for the compiler. I'm pretty sure the compiler can make the same code from both, but one is much easier to read.

And lambdas are different because they have little to no boilerplate code.

src/main/res/xml/app_settings.xml
```diff
@@ -119,6 +119,13 @@
         android:title="@string/category_about">
 
         <Preference
+            android:title="@string/open_log"
+            android:summary="@string/open_log_summary">
+            <intent
+                android:action=".activities.LogActivity" />
```
Nutomic: Needs to be this:
```
            <intent
                android:action="android.intent.action.VIEW"
                android:targetPackage="com.nutomic.syncthingandroid"
                android:targetClass="com.nutomic.syncthingandroid.activities.LogActivity"
                android:data="open_source_licenses" />
```

Unfortunately, this doesn't work in debug builds, because we change the package name there. So the best solution is probably to send the intent from Java.

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
```diff
@@ -0,0 +1,161 @@
...
+            }
+            return log.toString();
+        } catch (IOException e) {
+            Log.w(TAG, "Error reading Android log", e);
+        } finally {
+            try {
+                if (pOut != null) {
+                    pOut.close();
```
Nutomic: pOut seems to be unused.

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
```diff
@@ -0,0 +1,161 @@
...
+
+/**
+ * Shows the log information from Syncthing.
+ */
+public class LogActivity extends SyncthingActivity {
+
+    private final static String TAG = "LogActivity";
+
+    private TextView mLog;
+    private boolean mSyncthingLog = true;
```
Nutomic: I'm not sure what the problem is, but this still doesn't work right with device rotation.

It doesn't help that `mSyncthingLog == true` indnicates "View Android Log" and It doesn't help that `mSyncthingLog == false` indnicates "View Syncthing Log", you should rename this.

Zillode: No because it shows the syncthing log if it equals true. The button is meant to switch to the other state.

Nutomic:
Why did you close this?

src/main/res/xml/app_settings.xml
Zillode: If you keep the intent filter I added before it works as intended.

Zillode:
Because there were too many changes in other branches. I like to have one build with all features enabled to properly test it. My [master branch](https://github.com/zillode/syncthing-android/) is stable for v0.11

src/main/res/xml/app_settings.xml
Nutomic: Okay I guess this is alright. I just don't completely understand the `DEFAULT` intent category.

Issue #96, opened 2014-08-21
FR : Properly killed syncthing binary


blurmonk:
Hi

One of the main issue I have with this tool is that it never kills its own process when exiting or restarting it. Restarting it via the gui is almost useless because it never restarts properly and at times I have to just restart my tablet which is an inconvenience. Not a life threatening issue for sure but it would be nice if restarting is done properly so we could test things out much easier. 

Sometimes I kill the process using android`s running apps but then it is not a solution because recurrent restarts of this app ends up in crash at start with some message 

"
Syncthing Binary Crashed
The syncthing binary has exited with error code 2.
If this error persists, try reinstalling the app and restarting your devic

 It seems like when a health ST restart is needed, the best way is to restart the tablet.

I hope that the dev could address this issue so that we could test out versions with a nicer setup

thanks

PR #321, opened 2015-03-28
Ensure Syncthing is not running (fixes #92 and #96)

### LINKED ISSUE #92: Syncthing Binary Crashed  - error code 2
Hi,
I just updated syncthing to version 0.4.8 using Google Play (I suppose this is the version due to the comments... "disallow adding the same node multiple times" ...)
When running, this popup comes:

Syncthing Binary Crashed
The syncthing binary has exited with error code 2.
If this error persists, try reinstalling the app and restarting your device.

I've tried both but still happens...

any suggestions ?
(I am using android-2.3.7 - CyanogenMode-7.2.0-blade, on ZTE-Blade)

Zillode:
Ensure Syncthing is not running
- Fixes a bug: api.shutdown() is called asynchronously (line 278)
- Kills Syncthing processes on shutdown (e.g. when the /rest/shutdown call could not be applied because ST was too busy).

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
@@ -21,7 +21,9 @@ public class SyncthingRunnable implements Runnable {
 
     private static final String TAG = "SyncthingRunnable";
 
-    private static final String TAG_IONICE = "SyncthingRunnable-IoNice";
+    private static final String TAG_NICE = "SyncthingRunnable-Nice";
+
+    private static final String TAG_KILL = "SyncthingRunnable-Kill";
 
     private static final String TAG_NATIVE = "SyncthingNativeCode";
 
@@ -99,7 +101,7 @@ public class SyncthingRunnable implements Runnable {
                 DataOutputStream niceOut = null;
                 int ret = 1;
                 try {
-                    Thread.sleep(1000); // Wait a second before getting the Pid
+                    Thread.sleep(1000); // Wait a second before getting the pid
                     nice = Runtime.getRuntime().exec("sh");
                     niceOut = new DataOutputStream(nice.getOutputStream());
                     niceOut.writeBytes("set `ps |grep libsyncthing.so`\n");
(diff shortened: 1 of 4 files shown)


Nutomic:
I would move the logging into the loop, or into `killProcessId()`, so it only logs actual kills. Do you think one second is enough time on slower devices?

Please add a docstring to `killProcessId()` and log exceptions. And a unit test would be very nice ;)

Nutomic:
Oh and Github only understands `fixes #x, fixes #y` apparently.

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
```diff
@@ -131,6 +133,82 @@ public void run() {
...
+                psOut.writeBytes("ps | grep libsyncthing.so\n");
+                psOut.writeBytes("exit\n");
+                psOut.flush();
+                ps.waitFor();
+                InputStreamReader isr = new InputStreamReader(ps.getInputStream());
+                BufferedReader br = new BufferedReader(isr);
+                String id;
+                while ((id = br.readLine()) != null) {
+                    killProcessId(id, i > 0);
+                }
```
Nutomic: You don't seem to check if the kill succeeds, so you run it 3 times in any case. Which gives us the same problem if someone stops and immediately starts the app.

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
```diff
@@ -131,6 +133,82 @@ public void run() {
...
+        DataOutputStream killOut = null;
+        try {
+            kill = Runtime.getRuntime().exec("sh");
+            killOut = new DataOutputStream(kill.getOutputStream());
+            if (!force) {
+                killOut.writeBytes("kill " + id + "\n");
+                killOut.writeBytes("sleep 1\n");
+            } else {
+                killOut.writeBytes("sleep 3\n");
+                killOut.writeBytes("kill -9 " + id + "\n");
```
Nutomic: What's the logic with the sleep here? (different values and different command order) I think it should go into the end of the loop rather than here.

Nutomic:
Oh and log shows `E/SyncthingNativeCode﹕ Syncthing binary crashed with error code 143`, is there any way we can avoid that? (without removing that logging entirely)

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
Zillode: 1) perform a nice kill and wait one second
2) check which processes exist
3) if it still exists, wait 3 extra seconds and then kill it hard

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
Zillode: This is done synchronously

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
Nutomic: Yeah but I don't think we really need the SIGKILL, I've never had trouble to kill a command line program with CTRL-C. Like I said, this could be a problem if a new syncthing instance is started after the call, because it would then be killed.

Zillode: It won't because I changed the async ```PostTask().execute(mUrl, PostTask.URI_SHUTDOWN, mApiKey)``` to a blocking operation. I cannot see a path that would kill it when ST should be starting.
I could remove the sigkill, but then there is no guarantee it will be stopped.

PR #334, opened 2015-04-05
Make ARMv5 and 387 builds

Zillode:


Nutomic:
`GO386` is already set in line 45. `GOARM` should go in line 53 imo.

Sucks that I missed this -.-

PR #337, opened 2015-04-06
Notify Syncthing when folders are deleted

Zillode:
Fixes https://discourse.syncthing.net/t/app-not-triggering-sync-when-deleting-content/2335

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/util/FolderObserver.java
+++ b/src/main/java/com/nutomic/syncthingandroid/util/FolderObserver.java
@@ -107,7 +107,7 @@ public class FolderObserver extends FileObserver {
                         break;
                     }
                 }
-                break;
+                // fall through
             case MOVED_TO:
                 // fall through
             case CREATE:


Nutomic:
Damn, this is embarrasing.

Any chance you could write a test case for this?

Zillode:
I think inotify unit tests are covered by the Android framework quite well. This is a simple bugfix and the remaining code looks fine

Nutomic:
Let me quote the Syncthing Wiki here:

"Perhaps there isn't a test, and your code fixes a bug with the code. That just highlights that there should have been a test to start with, and this is the optimal time to add one."

src/main/java/com/nutomic/syncthingandroid/util/FolderObserver.java
```diff
@@ -107,7 +107,7 @@ public void onEvent(int event, String path) {
                         break;
                     }
                 }
-                break;
+                // fall through
```
Nutomic: This is actually wrong, as it would add observers if `fullPath` is a directory.

PR #341, opened 2015-04-06
Add option to run syncthing when tethering is enabled

grimpy:
Listen to event wifi ap changed to start syncthing

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/DeviceStateHolder.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/DeviceStateHolder.java
@@ -18,6 +18,10 @@ public class DeviceStateHolder extends BroadcastReceiver {
      * Intent extra containing a boolean saying whether wifi is connected or not.
      */
     public static final String EXTRA_HAS_WIFI = "has_wifi";
+    /**
+     * Intent extra containg a boolean saying whether tethering is enabled or not.
+     */
+    public static final String EXTRA_AP_ENABLED = "is_ap_enabled";
 
     /**
      * Intent extra containging a boolean saying whether the device is
@@ -26,6 +30,7 @@ public class DeviceStateHolder extends BroadcastReceiver {
     public static final String EXTRA_IS_CHARGING = "is_charging";
 
     private boolean mIsWifiConnected = false;
+    private boolean mIsApEnabled = false;
 
     private boolean mIsCharging = false;
 
@@ -54,8 +59,13 @@ public class DeviceStateHolder extends BroadcastReceiver {
         return mIsWifiConnected;
     }
 
+    public boolean isApEnabled() {
+        return mIsApEnabled;
+    }
+
     public void update(Intent intent) {
         mIsWifiConnected = intent.getBooleanExtra(EXTRA_HAS_WIFI, mIsWifiConnected);
         mIsCharging = intent.getBooleanExtra(EXTRA_IS_CHARGING, mIsCharging);
+        mIsApEnabled = intent.getBooleanExtra(EXTRA_AP_ENABLED, mIsApEnabled);
     }
 }
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/NetworkReceiver.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/NetworkReceiver.java
@@ -5,6 +5,7 @@ import android.content.Context;
 import android.content.Intent;
 import android.net.ConnectivityManager;
 import android.net.NetworkInfo;
+import android.net.wifi.WifiManager;
 import android.util.Log;
 
 /**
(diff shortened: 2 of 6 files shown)


src/main/java/com/nutomic/syncthingandroid/syncthing/NetworkReceiver.java
```diff
@@ -13,19 +14,27 @@
...
-        ConnectivityManager cm =
-                (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
-        NetworkInfo wifiInfo = cm.getNetworkInfo(ConnectivityManager.TYPE_WIFI);
-        boolean isWifiConnected = (wifiInfo != null && wifiInfo.isConnected());
-        Log.v(TAG, "Received wifi " + (isWifiConnected ? "connected" : "disconnected") + " event");
         Intent i = new Intent(context, SyncthingService.class);
-        i.putExtra(DeviceStateHolder.EXTRA_HAS_WIFI, isWifiConnected);
+        String action = intent.getAction();
+        if (action.equals("android.net.wifi.WIFI_AP_STATE_CHANGED")) {
+            int state = intent.getIntExtra("wifi_state", 0);
```
Zillode: 0 = WIFI_AP_STATE_DISABLED?

Zillode:
Cool:)

Zillode:
@st-jenkins build this please

Nutomic:
So the purpose of this is to sync to other mobile devices that are connected to the hotspot? Just curious, how do you avoid transfers from all devices over mobile data?

Also, I'm kinda worried that the settings are getting too full with this. maybe we should move these into a ListPreference (though not necessarily in this PR).

src/main/java/com/nutomic/syncthingandroid/syncthing/DeviceStateHolder.java
```diff
@@ -54,8 +59,13 @@ public boolean isWifiConnected() {
         return mIsWifiConnected;
     }
 
+    public boolean isAPOn() {
```
Nutomic: This should be `isApOn` for better readability (I like `isApEnabled` more). Same for other occurences.

src/main/java/com/nutomic/syncthingandroid/syncthing/NetworkReceiver.java
```diff
@@ -13,19 +14,27 @@
 public class NetworkReceiver extends BroadcastReceiver {
 
     private static final String TAG = "NetworkReceiver";
+    private static final int WIFI_AP_STATE_ENABLED = 13;
```
Nutomic: There should be a constant in the Android API somewhere?

Nutomic:
`NetworkReceiverTest` is failing, can you fix that? And ideally add one or more test cases for this functionality.

src/main/java/com/nutomic/syncthingandroid/syncthing/NetworkReceiver.java
grimpy: https://github.com/android/platform_frameworks_base/blob/master/wifi/java/android/net/wifi/WifiManager.java#L216

Hidden.

src/main/java/com/nutomic/syncthingandroid/syncthing/NetworkReceiver.java
grimpy: Something that is not enabled... disabled is 11 see https://github.com/android/platform_frameworks_base/blob/master/wifi/java/android/net/wifi/WifiManager.java#L206

Zillode:
@st-jenkins add to whitelist

src/main/java/com/nutomic/syncthingandroid/syncthing/NetworkReceiver.java
Nutomic: Ah that sucks. We should verify that this works accross Android versions. [These](http://stackoverflow.com/a/12531759) [links](http://stackoverflow.com/a/14405339) suggest that the values are different before ICS.

src/main/res/xml/app_settings.xml
```diff
@@ -20,6 +20,12 @@
             android:defaultValue="false" />
 
         <CheckBoxPreference
+            android:key="sync_only_ap"
+            android:title="@string/sync_only_ap"
+            android:defaultValue="false" />
+
+
```
Nutomic: Only one empty line please :)

Nutomic:
Sorry I didn't notice your update. Next time, could you post a comment so I get a notification?

[This](https://github.com/syncthing/syncthing-android/pull/341#discussion_r27831491) seems to be still relevant. Also, the option is not disabled if "Always run in Background" is unchecked.

`NetworkReceiverTest#testOnReceive()` is failing. You should add a test case for the new code anyway ;)

Additionally, I just noticed that the wording isn't entirely accurate ("sync only on x"), but that's a seperate problem.

Nutomic:
Closing this, feel free to reopen if you wanna work on merging it.

PR #345, opened 2015-04-09
Notify Syncthing when folders are deleted

Zillode:
'fall through' is obviously error prone.
This makes it explicit and clear.


Nutomic:
Sorry I had already fixed this in 1bc49354637137bd35701b0086c48ecb5906ec4a, thought you'd see the ref :s

PR #346, opened 2015-04-09
Download the latest Syncthing binaries when compiling on windows

Zillode:
Could be improved by rewriting `build-syncthing.sh` in groovy such that it builds on linux and windows.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,6 +12,8 @@ buildscript {
 }
 
 import java.util.regex.Pattern
+import groovy.json.*
+import org.apache.tools.ant.taskdefs.condition.Os
 
 apply plugin: 'com.android.application'
 apply plugin: 'com.github.ben-manes.versions'
@@ -32,7 +34,11 @@ dependencies {
 }
 
 preBuild {
-    dependsOn 'buildNative'
+    if (Os.isFamily(Os.FAMILY_WINDOWS)) {
+        dependsOn 'downloadNative'
+    } else {
+        dependsOn 'buildNative'
+    }
 }
 
 project.archivesBaseName = 'syncthing'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -111,6 +117,35 @@ def getVersionCodeFromManifest() {
     return Integer.parseInt(matcher.group(1))
 }
 
+task downloadNative(type: Copy) {
+    def release = new JsonSlurper().parseText("https://api.github.com/repos/syncthing/syncthing/releases/latest".toURL().text)
+
+    // Download ARM
+    new File('bin').mkdirs()
+    copy {
+        ant.get(src: "https://github.com/syncthing/syncthing/releases/download/v0.10.30/syncthing-linux-arm-"+release.tag_name+".tar.gz", dest: "bin/syncthing-arm.tar.gz")
```
Nutomic: This doesn't make sense, you're hardcoding the version in one place and reading it from the API in another?

Nutomic:
Is there a way to read the version from the git submodule? Anything else will break if the syncthing API changes.

Zillode:
- I don't think this is a big issue because we're always up to date.
- It would need a `git submodule init` to fetch the tagged Syncthing version, which means relying on git, which could also not be installed.

But sure, closed.

Nutomic:
I'm not saying we can't do this, it's just that I don't really like it.

Looks like we don't have any other option, so we'll have to do this. But we should have some kind of disclaimer that windows builds might randomly break.

Nutomic:
@Zillode As I said, I would merge this if you include a disclaimer in the readme.

Nutomic:
What's the status on this, can we merge it? As I said, all I want is a note in the readme that builds might break ;)

Nutomic:
Closing because this is outdated (feel free to reopen).

PR #369, opened 2015-04-23
TLS and v0.11.0 Support

Zillode:


Zillode:
@st-jenkins build this

Zillode:
(todo: add a patch to block the webgui for devices < 3.x because https://github.com/Zillode/syncthing-android/commit/8f2fe9955a862f7ae763f4a8237cd144ff9f4940#diff-705d504eb766ed22410cbd4084361666R63 does not exist). They need to manually accept the cert in their default browser.

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/RestApiTest.java
```diff
@@ -1,145 +0,0 @@
-package com.nutomic.syncthingandroid.test.syncthing;
```
Nutomic: Why are you deleting the tests? :(

Nutomic:
You can block < 3.x in `DrawerFragment#initMenu()`.

Would be nice if you could squash TLS commits into one commit, v0.11 API commits into another (so there are those two plus syncthing v0.11 update and port change).

Zillode:
(updated btw)
TLS verification of the webgui is omitted for devices < ICS

spinza:
Is there an schedule to get this to the play store?  Sync is broken for me as my main server auto updated via apt-get before i realised that 0.11 is incompatible.   If not I may downgrade everything to 0.10 again.

PR #370, opened 2015-04-23
Only copy device ID when finger lifts

Zillode:


Nutomic:
What problem is this fixing, exactly? Was the device ID copied twice?

Also, according to the [docs](http://developer.android.com/reference/android/view/View.OnTouchListener.html#onTouch%28android.view.View,%20android.view.MotionEvent%29) this should only return true if the event was consumed.

Zillode:
True but multiple MotionEvent are triggered because the finger touches down, moves and lifts up again. On my phone the toast is repeated at least 3 times.

Nutomic:
I see. Still, the function shouldn't always return true.

Nutomic:
LGTM!

PR #371, opened 2015-04-23
Log

Zillode:


src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    @Override
+    public boolean onCreateOptionsMenu(Menu menu) {
+        MenuInflater inflater = getMenuInflater();
+        inflater.inflate(R.menu.log_list, menu);
+
+        // Add the share button
+        MenuItem shareItem = menu.findItem(R.id.menu_share);
+        shareItem.setTitle(mSyncthingLog ? R.string.log_android_title : R.string.log_syncthing_title);
```
Nutomic: You need to do the `setTitle()` on `R.id.switch_logs`.

src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -65,6 +65,9 @@ protected void onCreate(Bundle savedInstanceState) {
                     throw new IllegalArgumentException(
                             "You must provide the requested fragment type as an extra.");
             }
+        } else{
+            setTitle(R.string.settings_title);
+            mFragment = new SettingsFragment();
```
Nutomic: I still don't know what this is good for.

src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
```diff
@@ -0,0 +1,161 @@
...
+            protected void onPostExecute(String log) {
+                mLog.setText(log);
+                if (mShareIntent != null)
+                    mShareIntent.putExtra(android.content.Intent.EXTRA_TEXT, log);
+                scrollToBottom();
+            }
+        }.execute();
+    }
+
+    private String getLog(final boolean syncthingLog) {
```
Nutomic: Add javadoc please. Also, any chance you can shorten this method? (eg using try with resources)

src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
Zillode: As stated in the previous PR, this is needed.
Log is an activity with a 'return' behavior to SettingsActivity. Therefore the SettingsActivity will be started with an empty action intent.

src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -65,6 +65,9 @@ protected void onCreate(Bundle savedInstanceState) {
                     throw new IllegalArgumentException(
                             "You must provide the requested fragment type as an extra.");
             }
+        } else {
```
Nutomic: I still don't see the point of this. The only way it could happen if someone forgets to set the intent action in new code, and in that case, we should throw an AssertionError.

Zillode: Remove it and try the back button from the log window.

Nutomic: Ah I see. In that case, I would document this with a comment.

PR #394, opened 2015-05-17
Added option to run syncthing as root (fixes #48).

### LINKED ISSUE #48: An option to be able to sync files on rooted phones
Hi,
I hope it is already possible, but I guess it's not.
It would be really useful to be able to sync files in protected directories (like /data ) for rooted phones.
To make the app ask for sudo authorization when the option is activated.

Nutomic:
Right now, it breaks when switching from `root` to `non-root`, because index files are then owned by root. Is there a better option than manually running chown on setting change?

I haven't tested without root, we should check that the setting is really hidden in that case.

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
@@ -5,6 +5,7 @@ import android.content.Intent;
 import android.content.SharedPreferences;
 import android.os.Environment;
 import android.preference.PreferenceManager;
+import android.text.TextUtils;
 import android.util.Log;
 
 import java.io.BufferedReader;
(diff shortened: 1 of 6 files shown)


src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
```diff
@@ -88,8 +91,7 @@ public void run() {
         try {
             ProcessBuilder pb = new ProcessBuilder("chmod", "+x", mSyncthingBinary);
             Process p = pb.start();
-            if (p != null)
-                p.waitFor();
+            p.waitFor();
```
Zillode: Are you sure this isn't possible? I think I added it because of a crash submitted to the play store, but I'm not sure.

Nutomic: Android Studio said that p can never be null. Can you find the crash log again?

Zillode:
@st-jenkins test this please

Zillode:
@st-jenkins run integration tests please

Zillode:
@st-jenkins run integration tests please

Zillode:
@st-jenkins run integration tests please

Nutomic:
Eh, I didn't intend to merge this yet (should have been more clear on this).

As I wrote, the problem is that syncthing breaks after disabling the setting, because index files are owned by root.

Issue #422, opened 2015-07-07
Tablet Screenshots


Nutomic:
Google Play says I should add those, for 7 inch and 10 inch tablets.

Can someone with such a device take some screenshots? Ideally as a pull request, but anything works really ;)

Edit: Updated phone screenshots are also welcome!

wweich:
The pull request would only consist of the 3 image files in  `/market/`, right?
I took some current shots from my Nexus 4: https://goo.gl/photos/XtSbQCLiwT8wqAVPA
Its a fresh install which will be removed, so the ID and such don't have to by obfuscated ;)
I will upload Screenshots from my old Nexus 7 once the battery is loaded enough to power on :)

If you want just the current 3 shots updated / in tablet form, I can make a PR once I have the tablet shots.

wweich:
Here are the 7 inch tablet shots: https://goo.gl/photos/gXPzjE7nszpYZMd27
Again,  it's a fresh install which will be removed, so the ID and such don't have to by obfuscated ;)

Nutomic:
Yeah exactly, with the tablet screenshots named something like `screenshot_tablet_7_inch_x`.

Also, there's a limit of 8 pictures per type, so I'd just leave the "loading" and "restart needed" pictures out. They don't make for good advertising anyway :D

wweich:
My PR with updated phone and 7inch screenshots is in #423.

PR #423, opened 2015-07-09
Update phone screenshots + new 7 inch tablet screenshots

wweich:
Here are updated screenshots for phone and 7 inch tablet with current Syncthing Adnroid version.


Nutomic:
There seems to be something wrong with your pictures, they are named .png but contain html code? :o 

Also, I'd prefer you reorder the pictures so that you don't have to change the readme (that way, we can easily have the same ordering on Google Play).

Once you've done that, could you squash your commits and add "(ref #422)" to the commit?

PR #430, opened 2015-07-14
Bugfix: Request parameters are duplicated over and over

FlashSystems:
If the request is repeated within the for loop the request parameters get built over and over again causing the "?xxx=yyy" part to be repeated over and over.


Nutomic:
Makes sense, thank you :)

PR #431, opened 2015-07-15
Asynchronous event API

FlashSystems:
I had a go at implementing the event API. It works asynchronous from the other REST calls because synchting is blocking if no events are available. Specific events (currently ItemFinished, FolderCompletion, DeviceConnected, DeviceDiscovered, StateChanged) are converted to broadcast intents with the base URI com.nutomic.syncthingandroid.event. A itemfinished event is sent as a broadcast intent of com.nutomic.syncthingandroid.event.itemfinished. This allows the service to implement a broadcast receiver and act according to the event it got (eg. update the media database, show a message, etc.). More than this, it enables other applications to register for the broadcast intents and act on them (eg. a widget showing the last synced file).

This code is meant as a first try and base for discussion. Some things are currently missing:
- If syncthing is restarted, previous events are lost. Before stopping syncthing remaining events should be processed.
- No tests.
- The broadcast intents must be used and converted into useful actions.
- When restarting syncthing the event processor must be correctly restarted also. Currently this is not always working.

Before I go any further I'd like to know what you think about this idea. Is converting the events into broadcast intents a way to implement this or does it collide with your design ideas for this app?

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
@@ -1,0 +1,190 @@
+package com.nutomic.syncthingandroid.syncthing;
+
+import android.content.Context;
+import android.content.Intent;
+import android.content.SharedPreferences;
+import android.os.Bundle;
+import android.os.Handler;
+import android.os.Looper;
+import android.preference.PreferenceManager;
+import android.support.v4.content.LocalBroadcastManager;
+import android.util.Log;
+
+import java.io.File;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Run by the syncthing service to convert syncthing events into local broadcasts.
+ * It uses SyncthingService.GetEvents to read the pending events and wait for new events.
+ */
+public class EventProcessor implements SyncthingService.OnWebGuiAvailableListener, Runnable,
+        RestApi.OnReceiveEventListener {
+
+    private static final String TAG = "EventProcessor";
+    private static final String PREF_LAST_SYNC_ID = "last_sync_id";
+
+    private static final String EVENT_BASE_ACTION = "com.nutomic.syncthingandroid.event";
+
+    /**
+     * Minimum interval in seconds at which the events are polled from syncthing and processed.
+     * This intervall will not wake up the device to save battery power.
+     */
+    public static final long EVENT_UPDATE_INTERVAL = TimeUnit.SECONDS.toMillis(15);
+
+    // Use the MainThread for all callbacks and message handling
+    // or we have to track down nasty threading problems.
+    private final Handler mMainThreadHandler = new Handler(Looper.getMainLooper());
+
+    private volatile long mLastEventId = 0;
+    private volatile boolean mShutdown = true;
+
+    private final Context mContext;
+    private final RestApi mApi;
+    private final LocalBroadcastManager mLocalBM;
+    private final Map<String, String> mFolderToPath = new HashMap<>();
+
+    /**
+     * Returns the action used by notification Intents fired for the given Syncthing event.
+     * @param eventName Name of the Syncthing event.
+     * @return Returns the full intent action used for local broadcasts.
+     */
+    public static String getEventIntentAction(String eventName) {
+        return EVENT_BASE_ACTION + "." + eventName.toUpperCase();
+    }
+
+    /**
+     * C'tor
+     * @param context Context of the service using this event processor.
+     * @param api Reference to the RestApi-Instance used for all API calls by this instance of the
+     *            Event processor.
+     */
+    public EventProcessor(Context context, RestApi api) {
+        mContext = context;
+        mApi = api;
+        mLocalBM = LocalBroadcastManager.getInstance(mContext);
+    }
+
+    private void updateFolderMap()
+    {
+        synchronized(mFolderToPath) {
+            mFolderToPath.clear();
+            for (RestApi.Folder folder: mApi.getFolders()) {
+                mFolderToPath.put(folder.id, folder.path);
+            }
+        }
+    }
+
+    /**
+     * @see Runnable
+     */
+    @Override
+    public void run() {
+        // Restore the last event id if the event processor may have been restartet.
+        if (mLastEventId == 0) {
+            mLastEventId = PreferenceManager.getDefaultSharedPreferences(mContext)
+                    .getLong(PREF_LAST_SYNC_ID, 0);
+        }
+
+        // First check if the event number ran backwards.
+        // If that's the case we've to start at zero because syncthing was restartet.
+        mApi.getEvents(0, 1, new RestApi.OnReceiveEventListener() {
+            @Override
+            public void onEvent(long id, String eventType, Bundle eventData) {
+            }
+
+            @Override
+            public void onDone(long lastId) {
+                if (lastId < mLastEventId) mLastEventId = 0;
+
+                Log.d(TAG, "Reading events starting with id " + mLastEventId);
+
+                mApi.getEvents(mLastEventId, 0, EventProcessor.this);
+            }
+        });
+    }
+
+    /**
+     * @see RestApi.OnReceiveEventListener
+     */
+    @Override
+    public void onEvent(final long id, final String eventType, final Bundle eventData) {
+        // If a folder item is contained within the event. Resolve the local path.
+        if (eventData.containsKey("folder")) {
+            String folderPath = null;
+            synchronized (mFolderToPath) {
+                if (mFolderToPath.size() == 0) updateFolderMap();
+                folderPath = mFolderToPath.get(eventData.getString("folder"));
+            }
+
+            if (folderPath != null) {
+                eventData.putString("_localFolderPath",folderPath);
+
+                if (eventData.containsKey("item")) {
+                    final File file = new File(new File(folderPath), eventData.getString("item"));
+
+                    eventData.putString("_localItemPath", file.getPath());
+                }
+            }
+        }
+
+        final Intent eventBroadcastItent = new Intent(EVENT_BASE_ACTION + "." + eventType.toUpperCase());
+        eventBroadcastItent.putExtras(eventData);
+        mLocalBM.sendBroadcast(eventBroadcastItent);
+
+        Log.d(TAG,
+                "Sent local event broadcast " + eventBroadcastItent.getAction() +
+                " including " + eventType.length() + " extra data items."
+        );
+    }
+
+    /**
+     * @see RestApi.OnReceiveEventListener
+     */
+    @Override
+    public void onDone(long id) {
+        if (mLastEventId < id) {
+            mLastEventId = id;
+
+            // Store the last EventId in case we get killed
+            final SharedPreferences sp = PreferenceManager.getDefaultSharedPreferences(mContext);
+
+            //noinspection CommitPrefEdits
+            sp.edit().putLong(PREF_LAST_SYNC_ID, mLastEventId).commit();
+        }
+
+        synchronized (mMainThreadHandler) {
+            if (!mShutdown) {
+                mMainThreadHandler.removeCallbacks(this);
+                mMainThreadHandler.postDelayed(this, EVENT_UPDATE_INTERVAL);
+            }
+        }
+    }
+
+    /**
+     * @see SyncthingService.OnWebGuiAvailableListener
+     */
+    @Override
+    public void onWebGuiAvailable() {
+        Log.d(TAG, "WebGUI available. Starting event processor.");
+
+        updateFolderMap();
+
+        // Remove all pending callbacks and add a new one. This makes sure that only one
+        // event poller is running at any given time.
+        synchronized (mMainThreadHandler) {
+            mShutdown = false;
+            mMainThreadHandler.removeCallbacks(this);
+            mMainThreadHandler.postDelayed(this, EVENT_UPDATE_INTERVAL);
+        }
+    }
+
+    public void shutdown() {
+        Log.d(TAG, "Shutdown event processor.");
+        synchronized (mMainThreadHandler) {
+            mShutdown = true;
+            mMainThreadHandler.removeCallbacks(this);
+        }
+    }
+}
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/GetTask.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/GetTask.java
@@ -36,6 +36,7 @@ public class GetTask extends AsyncTask<String, Void, String> {
     public static final String URI_MODEL       = "/rest/db/status";
     public static final String URI_DEVICEID    = "/rest/svc/deviceid";
     public static final String URI_REPORT      = "/rest/svc/report";
+    public static final String URI_EVENTS      = "/rest/events";
 
     private String mHttpsCertPath;
 
@@ -55,9 +56,11 @@ public class GetTask extends AsyncTask<String, Void, String> {
         String fullUri = params[0] + params[1];
         Log.v(TAG, "Calling Rest API at " + fullUri);
 
-        if (params.length == 5) {
+        if (params.length >= 5) {
             LinkedList<NameValuePair> urlParams = new LinkedList<>();
-            urlParams.add(new BasicNameValuePair(params[3], params[4]));
+            for (int paramCounter = 3; paramCounter + 1 < params.length; ) {
+                urlParams.add(new BasicNameValuePair(params[paramCounter++], params[paramCounter++]));
+            }
             fullUri += "?" + URLEncodedUtils.format(urlParams, HTTP.UTF_8);
         }
 
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/RestApi.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/RestApi.java
@@ -9,6 +9,8 @@ import android.content.ClipboardManager;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
+import android.os.AsyncTask;
+import android.os.Bundle;
 import android.util.Log;
 import android.widget.Toast;
 
@@ -173,6 +175,11 @@ public class RestApi implements SyncthingService.OnWebGuiAvailableListener,
      * for calculating device percentage in {@link #getConnections(OnReceiveConnectionsListener)}.
      */
     private HashMap<String, Model> mCachedModelInfo = new HashMap<>();
+
+    /**
+     * Stores a hash map to resolve folders to paths for events.
+     */
+    private final Map<String, String> mCacheFolderPathLookup = new HashMap<String, String>();
 
     public RestApi(Context context, String url, String apiKey, String guiUser, String guiPassword,
                    OnApiAvailableListener listener) {
@@ -644,7 +651,6 @@ public class RestApi implements SyncthingService.OnWebGuiAvailableListener,
                 : 100;
     }
 
-
     /**
      * Listener for {@link #getModel}.
      */
@@ -652,6 +658,25 @@ public class RestApi implements SyncthingService.OnWebGuiAvailableListener,
         public void onReceiveModel(String folderId, Model model);
     }
 
+    /**
+     * Listener for {@link #getEvents}.
+     */
+    public interface OnReceiveEventListener {
+        /**
+         * Called for each event.
+         * @param id ID of the event. Monotonously increasing.
+         * @param eventType Name of the event. (See Syncthing documentation)
+         * @param eventData Bundle containing the data fields of the event as data elements.
+         */
+        void onEvent(final long id, final String eventType, final Bundle eventData);
+
+        /**
+         * Called after all available events have been processed.
+         * @param lastId The id of the last event processed. Should be used as a starting point for
+         *               the next round of event processing.
+         */
+        void onDone(long lastId);
+    }
     /**
      * Returns status information about the folder with the given id.
      */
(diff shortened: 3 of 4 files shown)


src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
```diff
@@ -0,0 +1,90 @@
...
+                        }
+
+                        @Override
+                        public void onDone(long id) {
+                            if (mLastEventId < id) mLastEventId = id;
+                        }
+                    });
+                }
+            }
+        }, EVENT_UPDATE_INTERVAL, EVENT_UPDATE_INTERVAL, TimeUnit.MILLISECONDS);
```
Nutomic: Shouldn't `initialDelay` be 0, so it starts immediately?

src/main/java/com/nutomic/syncthingandroid/util/AsyncState.java
```diff
@@ -0,0 +1,30 @@
...
+import android.os.AsyncTask;
+
+import java.lang.ref.WeakReference;
+
+/**
+ * Tracks the current state of a running AsyncTask without keeping a reference to it.
+ * This allows the object to be GCed as soon as it is finished. Releasing the resources
+ * allocated by the object.
+ */
+public class AsyncState {
```
Nutomic: Are you sure this works, and the AsyncTask won't be GC'd while it is being executed?

Nutomic:
Very cool!

I think it would be better to move `getEvents()` into `EventProcessor`, as `RestApi` is already quite big (but I'm not quite sure about this, depends how well it works).

Not sure about the broadcasts (because of security and privacy). I'd prefer to use `LocalBroadcastManager`, then we can still replace it with normal broadcasts if there's an concrete use case.

I didn't compile/run the code yet, but it seems the events aren't consumed anywhere. To make this PR complete, all GUI classes should use this interface, and all the get* functions in `RestApi` should be removed (although I'm not sure where we can get data at app start).

I had already started work in this (on the [event-api branch](https://github.com/syncthing/syncthing-android/tree/event-api)), but I tried to do it in Scala, and haven't even finished the event parsing yet. Seeing as your code is further ahead (and I don't have much time/motiviation right now), it's definitely the best option. And your approach is good, just needs some more things handled. There are also some formatting problems (eg missing newlines at the end of files), but that doesn't really matter at this point.

src/main/java/com/nutomic/syncthingandroid/util/AsyncState.java
FlashSystems: While the AsyncTask is running there will be a reference to the object on the stack of the currently executing thread. This reference will be part of the root set used by the garbage collector.  The instance can not be collected before doInBackground returns because the this reference on the threads stack will keep it alive.

See also http://www.dynatrace.com/en/javabook/how-garbage-collection-works.html

src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
FlashSystems: I was a little unsure if starting to process events asap would interfere with the other start-up calls. So I backed off a little.

src/main/java/com/nutomic/syncthingandroid/util/AsyncState.java
FlashSystems: Regarding the "premature optimization". It's no problem to replace the WeakReference by a real object reference. It just didn't feel right to return the full AsyncTask because this would allow the caller to cancel the operation or interfere with the inner workings of getEvents in some way. So I wanted an interface class that hides the inner workings of getEvents and just allows to poll the state of the operation. So I reused what I had lying around (I wrote this for an other project that used much more worker threads).

FlashSystems:
Hello Nutomic,

using the "LocalBroadcastManager" is a great idea. Maybe at a later stage some (non critical) or all broadcasts can be exposed to other apps. You are right, security issues have to be considered before information about synced files and devices is made public. I'll implement it that way.
 
The broadcast intent isn't currently consumed anywhere. I tested it via the debugger and adb.
In my opinion, the first use case would be notifying the media manager about new files. I'm currently unsure where to put the broadcast receiver for this action. [Separate Broadcast Receiver] What is your recommendation?

Later the activities code can be migrated to using the itents. I don't think that the get* methods within RestApi can be completely removed because the events only fire if something changes. If you just start up the interface, you still need a way to populate the information. But the regular polling for new data can be replaced by processing the event intents. The two interfaces can coexist while migrating from the polling to the event approach. So this can be done step by step.
 
Shouldn't this option be configurable per folder? Currently syncthing for android seems not to store per folder configuration information (besides the options that are passed to syncthing itself). How should this be implemented?
As an alternative the media manager notification could be a global option. Than the current settings dialog can be used.
I don't know if an option is at all necessary. Maybe the media manager should always be notified and the user should not be confused with more options.

One thing I've to figure out is how to do a last round of event processing before stopping syncthing. Or some events that fired just before syncthing stopped might be lost...

FlashSystems:
Oh dear. I found a really big problem with my implementation. If the service gets killed and you restart the synchting app all events since 0 will get reprocessed. This is bad. Putting the syncthing is running notification into foreground via "this.startForeground(NOTIFICATION_ACTIVE, n);" helps keeping the service alive. But it's not a perfect solution. The user can disable the notification and then the service will get killed if you swipe to close. 

See http://stackoverflow.com/questions/18612880/android-service-crashes-after-app-is-swiped-out-of-the-recent-apps-list

I have to think about this. Maybe i can save the timestamp of the last event I processed. If the event handler gets called, only events with a newer timestamp are reprocessed. Just saving the event id won't work because if syncthing itself gets restartet too, then we will wait forever for the message ids to reach the saved value. So starting from 0 when the service is restarted is a safe bet.

Really sorry for flooding you with text but I want to make this work and you've got more experience with this app than me...

I've implemented a broadcast listener that sends new files to the media scanner. Watching pictures appear and disappear from the gallary while they are synced is like magic :)

src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
Nutomic: The API should be available after `onWebGuiAvailable()`, so there shouldn't be any problems. I'd say just set it to 0, and if it works, that's good ;)

src/main/java/com/nutomic/syncthingandroid/util/AsyncState.java
Nutomic: > The instance can not be collected before doInBackground returns because the this reference on the threads stack will keep it alive.
I don't quite understand this, do you mean a reference inside the `AsyncTask` class? Anyway, I think it would be better to make `getEvents()` a private method in `EventProcessor`, so that kind of abstraction won't be needed.

Nutomic:
Isn't the MediaScanner API really simple, like 2 lines of code? If so, I'd just handle it directly in the `eventType` switch (maybe extracting that into a seperate function).

If a complete replacement doesn't work, that's fine, although I'd love to reduce the amount of code there (`RestApi` and `EventProcessor` could at least share the JSON parsing functionality).

I assume the option you mean is for the MediaScanner? I don't see why anyone would disable that, it should really always be enabled.

For the shutdown, maybe we could wait for timeout on the last call?

For the service killing, how bad is the reprocessing? Is there a noticable delay? (more than a few seconds). @calmh @AudriusButkevicius can you say how the web gui handles that?

Generally, it might be a good idea to look at [Syncthing's web gui](https://github.com/syncthing/syncthing/tree/master/gui/scripts/syncthing) for an example implementation.

AudriusButkevicius:
We just skip stuff https://github.com/syncthing/syncthing/blob/master/gui/scripts/syncthing/core/services/events.js

FlashSystems:
Ok, I've now created a Version that is working most of the time. But it is fragile as hell. Android is killing the service from time to time. This happens on my test device and on my normal phone as well.
If the service is killed the events do not get processed any more. As soon as the service is restarted it kills the libsyncthing.so (that was still running and syncing happily) and all events that happened in the meanwhile are lost. This leads to images not appearing within the gallery.
Another thing I noticed while testing: Sometimes syncthing is not restarted when the service is. Then all the events are reprocessed and the media service is re-indexing all the files that were transferred since the last restart. If you move many pictures this may hurt battery life considerably.
This also explains why the inotify support is failing on my phone from time to time. As soon as I start the app to check what's going on the problem vanishes. -> When the service is killed, the ContentObserver won't work as soon as syncthing is restarted it's re-indexing all folders and the problem vanishes.

The last two days I've digged into the source and the syncthing API. In my opinion the event system is not usable for the service part of syncthing for android. It is made for the web interface (thank you for the link AudriusButkevicius). In this use-case it's not a problem if events get lost if nobody is looking (if the website is closed). But syncthing for android is trying to synchronize operating system state (the media scanner or a user notification) with the syncthing executable. And as soon as events are lost glitches appear and everything falls apart. Having tried it many times I think keeping a service running on android is a futile approach. Android isn't made for that and google is changing the service life-cycle with every release.

I'm sorry, but it seems I've wasted your time with the events approach :(

Nutomic:
What is your Android version, and are you using AOSP? Maybe older versions handle services differently, or RAM is a problem on your device.

The [silk client](https://github.com/OpenSilk/SyncthingAndroid) also uses the event API and seems to work fine, so there should be a way. One idea would be using SharedPreferences to store the event id (and maybe restart state). Also, we could wait with the MediaScanner for a later point.

I'm a bit busy with exams and will go on vacation soon, so I don't really have time to dig into this. But I'll try and help with this as soon as I can.

FlashSystems:
I'm using Android x86 4.4.2 as my development device and Android 5.1 on a Moto X.
I'll dig a little bit more into as my time permits and give you feedback. I've got an idea that I want to try.

Good luck with your exams.

Nutomic:
Any progress on this?

FlashSystems:
I've to admit that my time for working on the event interface was rather limited. I still trying to get this to work reliably. I'll notify you as soon as I've got a fully working version.

Out of curiosity I tinkered around with gomobile and go 1.5 to create a real syncthing library for android. I think real callbacks instead of the event interface would be great. But my go knowledge is between very limited and non existent (but growing) and the syncthing code is rather complex. ;(

Nutomic:
Better late than never :)

Gomobile seems to be meant to write Android apps in Go. That won't really work i think, because you can't access the Android APIs, and only render OpenGL.

FlashSystems:
Regarding Gomobile: It can build libraries from go programs that can be used by android applications:
https://godoc.org/golang.org/x/mobile/cmd/gomobile ("Build a library for Android and iOS")
https://godoc.org/golang.org/x/mobile/example/bind/hello
Being able to create a library from syncthing would be great as it removes the fragile dependency on the background process. Syncthing is very modular and I think it should be possible. Maybe someone with more syncthing and go knowledge can give some insights.

Nutomic:
That might be pretty useful. The problem is it would probably require major changes to both syncthing and syncthing-android. But you could start with a very basic integration to see if it's feasible. The only disadvantage I can see is that root wouldn't work any more.

@AudriusButkevicius @calmh Any thought on gomobile?

AudriusButkevicius:
I think it's possible, we've done a bit of work to separate stuff out recently.

FlashSystems:
Hello Nutomic. I finally made the event interface work. There are some corner cases (with being killed and syncthing restarting, etc.) that may not work reliably. But it's the best I could come up with till now.

Events are converted to local broadcasts. This way they can be used for updating the GUI, etc. I added a handler for itemfinished to call the media manager.

I'll investigate my library idea further as soon as time permits.

Nutomic:
Great! I'll give it a try later.

Just so I understand it right, events are only used to update the MediaScanner, right?

FlashSystems:
Currently: yes
But using local broadcasts enables other parts of the app (e.g. the UI) to register for the events they need. Every event is converted to an intend with the action set to EventProcessor.EVENT_BASE_URL + "." + <Event-Name in Uppercase>. For eg. if an item is completed a local broadcast intent with the action set to "com.nutomic.syncthingandroid.event.ITEMCOMPLETED" is served. The intend contains all fields of the original event as extra data.
To make the intend easier to use: If the event contains a "folder" value a string-extra with the name "_localFolderPath" is added containing the local path for the folder referenced. And if an "item" value is found it is resolved to a "_localItemPath" extra.

Currently only the following events are converted:
itemfinished
foldercompletion
deviceconnected
devicediscovered
statechanged

More can easily be added within getEvents in RestApi.java.

Nutomic:
One problem I noticed while testing is the following:
- add a toast to `mItemFinishedBroadcastReceiver`
- set app to run only while charging
- disconnect charger, make sure the app is disabled
- add a file on the other device
- reconnect the charger, see the file sync to the phone

Here, "Unhandled event Starting" is shown, but the toast we added previously is not shown.

src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * Run by the syncthing service to convert syncthing events into local broadcasts.
+ * It uses SyncthingService.GetEvents to read the pending events and wait for new events.
+ */
+public class EventProcessor implements SyncthingService.OnWebGuiAvailableListener {
+
+    private static final String TAG = "EventProcessor";
+    private static final String PREF_LASTSYNCID = "last_sync_id";
+
+    public static final String EVENT_BASE_URL = "com.nutomic.syncthingandroid.event";
```
Nutomic: Please remove this. It might be useful if/when we add system wide broadcasts, but right now, it's just extra complexity.

src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
```diff
@@ -0,0 +1,146 @@
...
+
+    /**
+     * Minimum interval in ms at which the events are polled from syncthing and processed.
+     * This intervall will not wake up the device to save battery power.
+     */
+    public static final int EVENT_UPDATE_INTERVAL = 60000;
```
Nutomic: Replace this with `TimeUnit.SECONDS.toMillis(60)`.

src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
```diff
@@ -0,0 +1,146 @@
...
+            @Override
+            public void run() {
+                // Restore the last event id if the event processor may have been restartet.
+                if (mLastEventId == 0) {
+                    mLastEventId = PreferenceManager.getDefaultSharedPreferences(mContext).getLong(PREF_LASTSYNCID, 0);
+                }
+
+
+                // First check if the event number ran backwards. If that's the case we've to start at zero because syncthing was restartet.
+                mApi.getEvents(0, 1, new RestApi.OnReceiveEventListener() {
```
Nutomic: This stuff is pretty deeply wrapper. Can you move the inner listeners to EventProcessor?

src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
```diff
@@ -0,0 +1,146 @@
...
+                    @Override
+                    public void onEvent(long id, String eventType, Bundle eventData) {
+                    }
+
+                    @Override
+                    public void onDone(long lastId) {
+                        // Now we know the last event ID. We check if it ran backwards and then start loading events.
+                        if (lastId < mLastEventId) mLastEventId = 0;
+
+                        if (Log.isLoggable(TAG, Log.DEBUG)) Log.d(TAG, "Reading events starting with id " + mLastEventId);
```
Nutomic: I don't think this condition is useful, and it's too much clutter.

src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
```diff
@@ -0,0 +1,146 @@
...
+                            }
+
+                            @Override
+                            public void onDone(long id) {
+                                if (mLastEventId < id) {
+                                    mLastEventId = id;
+
+                                    // Store the last EventId in case we get killed
+                                    final SharedPreferences sp = PreferenceManager.getDefaultSharedPreferences(mContext);
+                                    sp.edit().putLong(PREF_LASTSYNCID, mLastEventId).commit();
```
Nutomic: Should be `apply()` instead of `commit()`.

src/main/java/com/nutomic/syncthingandroid/syncthing/RestApi.java
```diff
@@ -173,6 +177,11 @@ public void setParams(int keep) {
      */
     private HashMap<String, Model> mCachedModelInfo = new HashMap<>();
 
+    /**
+     * Stores a hash map to resolve folders to paths for events.
+     */
+    private Map<String, String> mCacheFolderPathLookup = new HashMap<String, String>();
```
Nutomic: Should be final.

src/main/java/com/nutomic/syncthingandroid/syncthing/RestApi.java
```diff
@@ -687,6 +702,111 @@ protected void onPostExecute(String s) {
...
+                }
+            }
+        };
+
+        // Within honeycomb execute was fully parallel. Since honeycomb the AsyncTasks are serialized and if the event interface
+        // blocks (no new events), all other tasks will block also. To prevent this, starting with HONEYCOMB the thread pool executor must be used.
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.HONEYCOMB)
+            return new AsyncState(eventGetTask.execute(mUrl, GetTask.URI_EVENTS, mApiKey, "since", String.valueOf(sinceId), "limit", String.valueOf(limit)));
+        else
+            return new AsyncState(eventGetTask.executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR, mUrl, GetTask.URI_EVENTS, mApiKey, "since", String.valueOf(sinceId), "limit", String.valueOf(limit)));
```
Nutomic: The post-honeycode option also works on older Android versions, right? So I'd only use that, and remove the conditional.

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
```diff
@@ -100,11 +107,26 @@
...
     /**
+     * Processes the local broadcast message if an item was finished.
+     * Launches the media scanner to update the media library.
+     */
+    private final BroadcastReceiver mItemFinishedBroadcastReceiver = new BroadcastReceiver() {
+
+        @Override
+        public void onReceive(Context context, Intent intent) {
+            final File updatedFile = new File(intent.getStringExtra("_localItemPath"));
+            context.sendBroadcast(new Intent(Intent.ACTION_MEDIA_SCANNER_SCAN_FILE, Uri.fromFile(updatedFile)));
```
Nutomic: There should be logging here.

src/main/java/com/nutomic/syncthingandroid/util/AsyncState.java
Nutomic: Now you don't even use the class anymore. So remove it ;)

src/main/java/com/nutomic/syncthingandroid/syncthing/RestApi.java
veniosg: To be honest you don't even care about pre-Honeycomb now that the minSdk is 11.

Nutomic: [API 11 is Android 3.0](http://developer.android.com/guide/topics/manifest/uses-sdk-element.html#ApiLevels), so we still support it.

veniosg: I must be missing something. minSdk is 11 (3.0, Honeycomb), meaning there's no need to cater for anything lower. The else clause here works on versions >= 11, not just > 11, which we indeed support.

Nutomic: Ah sorry, I completely misread your code/comment. You are correct.

src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
FlashSystems: Which of the conditions in this code snippet do you mean?
The "ran backwards" check is absolutely necessary to make it work because we save the last event id. If syncthing is restartet then the event ids will run from zero again. For eg. if the event id was 1000 then no events will be delivered until the id crosses the 1000 mark again.

src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
FlashSystems: Commit was choosen to save the event id as processed immediately. If apply is used we might get killed before the id is written and then we'll reprocess the event. We're already on a background thread so using commit() does not have any performance impact.

src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
FlashSystems: Why and how should it be removed? Actions for intents have to follow the naming convention specified within http://developer.android.com/guide/components/intents-filters.html (section Action). I don't see an exception for local itents within the documentation.
Maybe it won't matter. But I'd prefer to stick to the rules.

src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
Nutomic: Oh I mean the `isLoggable()` on the line I commented ;)

src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
Nutomic: Ah that makes sense. But you should disable the warning then, with `//noinspection CommitPrefEdits`.

src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
Nutomic: Ok it seems you are right. But you should make a static method to construct these event URIs, instead of doing it inline.

src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
FlashSystems: Google uses this version in their code. They mandate for it as making debugging easier (less cluttered log) and being more battery friendly (less calls to the logging service).But I can't find the source of this citation currently. I'll remove it to be consistent with the rest of the code.

FlashSystems:
I found the cause of the problems with processing events after being shut down while unwinding the deeply wrapped parts of the code. Creating a new pull request asap.

src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
```diff
@@ -0,0 +1,171 @@
...
+    private static final String TAG = "EventProcessor";
+    private static final String PREF_LASTSYNCID = "last_sync_id";
+
+    private static final String EVENT_BASE_URL = "com.nutomic.syncthingandroid.event";
+
+    /**
+     * Minimum interval in seconds at which the events are polled from syncthing and processed.
+     * This intervall will not wake up the device to save battery power.
+     */
+    public static final long EVENT_UPDATE_INTERVAL = 15;
```
Nutomic: I meant that you should put the `TimeUnit.SECONDS.toMillis(15)` directly here, so you don't have to convert it later. Because this way, it's not clear what unit is used.

src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
```diff
@@ -0,0 +1,171 @@
...
+        // First check if the event number ran backwards. If that's the case we've to start at zero because syncthing was restartet.
+        mApi.getEvents(0, 1, new RestApi.OnReceiveEventListener() {
+            @Override
+            public void onEvent(long id, String eventType, Bundle eventData) {
+            }
+
+            @Override
+            public void onDone(long lastId) {
+                // Now we know the last event ID. We check if it ran backwards and then start loading events.
+                if (lastId < mLastEventId) mLastEventId = 0;
```
Nutomic: You basically have the same comment twice, in line 83 and 91 :p 

Also, can you make the code after the if in a new line for clarity/consistency? (same in line 107).

src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
```diff
@@ -0,0 +1,171 @@
...
+    @Override
+    public void onWebGuiAvailable() {
+        Log.d(TAG, "WebGUI available. Starting event processor.");
+
+        updateFolderMap();
+
+        // Remove all pending callbacks and add a new one. This makes sure that only one
+        // event poller is running at any given time.
+        synchronized (mMainThreadHandler) {
+            mShutdown=false;
```
Nutomic: Can you put spaces around the = sign? This looks really weird :D

src/main/java/com/nutomic/syncthingandroid/syncthing/GetTask.java
```diff
@@ -55,9 +56,12 @@ protected String doInBackground(String... params) {
...
         Log.v(TAG, "Calling Rest API at " + fullUri);
 
-        if (params.length == 5) {
+        if (params.length >= 5) {
             LinkedList<NameValuePair> urlParams = new LinkedList<>();
-            urlParams.add(new BasicNameValuePair(params[3], params[4]));
+            int paramCounter = 3;
+            while (paramCounter + 1 < params.length) {
+                urlParams.add(new BasicNameValuePair(params[paramCounter++], params[paramCounter++]));
+            }
```
Nutomic: You could do `for (int paramCounter = 3; paramCounter + 1 < params.length; )` here.

Nutomic:
Some small nitpicks on the code, I'll also test it again later.

FlashSystems:
I cleaned the code up a little according to your remarks.

src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
```diff
@@ -0,0 +1,170 @@
...
+import eu.chainfire.libsuperuser.Shell;
+
+/**
+ * Run by the syncthing service to convert syncthing events into local broadcasts.
+ * It uses SyncthingService.GetEvents to read the pending events and wait for new events.
+ */
+public class EventProcessor implements SyncthingService.OnWebGuiAvailableListener, Runnable, RestApi.OnReceiveEventListener {
+
+    private static final String TAG = "EventProcessor";
+    private static final String PREF_LASTSYNCID = "last_sync_id";
```
Nutomic: Should be `PREF_LAST_SYNC_ID` for readability.

src/main/java/com/nutomic/syncthingandroid/syncthing/EventProcessor.java
```diff
@@ -0,0 +1,170 @@
...
+    // Use the MainThread for all callbacks and message handling or we have to track down nasty threading problems.
+    private final Handler mMainThreadHandler = new Handler(Looper.getMainLooper());
+
+    private volatile long mLastEventId = 0;
+    private volatile boolean mShutdown = true;
+
+    private final Context mContext;
+    private final RestApi mApi;
+    private final LocalBroadcastManager mLocalBM;
+    private final Map<String, String> mFolderToPath = new HashMap<String, String>();
```
Nutomic: You can just write `new HashMap<>()` here.

Nutomic:
Yep, seems to work fine now :+1: 

A few more things, can you adjust the code so that it uses at most 100 columns? And the functions in `EventProcessor` should have some Javadoc explaining what they do.

Also, events are a bit annoying to read being all uppercase (`DEVICEDISCOVERED`), maybe we should just leave it as lowercase?

Sorry for being so slow with reviewing!

FlashSystems:
I'd like to keep the event names uppercase because they are used as Intent-actions and Google mandates for uppercase actions in https://developer.android.com/intl/zh-tw/reference/android/content/Intent.html.

Squashed the changes after my merge from master into one change. Still trying to figure out how to put the changes before the merge into the squashed change.

Nutomic:
Looks like you did `git merge origin master` instead of `git rebase origin master`? Can you just revert the merge commit, then rebase and squash?

Nutomic:
Please comment when you update your PR ;)

Unfortunately there's no notifications for pushes.

FlashSystems:
Sorry, I was in a hurry and forgot to leave you a comment.

PR #436, opened 2015-07-24
Fix 'bad mode' error when chmodding with +x

Zillode:
Ref http://stackoverflow.com/questions/9285903/adb-shell-giving-bad-mode-when-executing-chmod-under-su


Nutomic:
Ah that's right. 

We should check why Jenkins is failing.

PR #446, opened 2015-08-20
First crack at implementing sharing functionality

anderspitman:
Implements handling "share button" intents from other apps.

Should resolve [this request](https://forum.syncthing.net/t/are-there-plans-for-a-share-feature-in-the-android-app/860).

This is still quite rough, but the basic functionality works. I wanted to get some feedback before starting to clean it up and optimize it since I'm new to android.

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
+++ b/src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
@@ -1,0 +1,78 @@
+package com.nutomic.syncthingandroid.fragments;
+
+import android.database.Cursor;
+import android.net.Uri;
+import android.provider.MediaStore;
+import android.util.Log;
+import android.view.View;
+import android.widget.AdapterView;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.ArrayList;
+
+public class SelectFolderFragment extends FoldersFragment {
+
+    private static final String TAG = "SelectFolderFragment";
+
+    @Override
+    public void onItemClick(AdapterView<?> adapterView, View view, int i, long l) {
+        for (Uri uri : files) {
+            File sourceFile = new File(getRealPathFromURI(uri));
+            Log.d(TAG, sourceFile.getPath());
+
+            String path = mAdapter.getItem(i).path;
+            File destFile = new File(path, sourceFile.getName());
+            Log.d(TAG, destFile.getPath());
+            try {
+                copy(sourceFile, destFile);
+            }
+            catch (IOException e) {
+                Log.e(TAG, "Failed copying file", e);
+            }
+        }
+
+        getActivity().finish();
+    }
+
+    public void setFiles(ArrayList<Uri> files) {
+        this.files = files;
+    }
+
+    // Copied from http://stackoverflow.com/a/9293885/943814
+    public void copy(File src, File dst) throws IOException {
+        InputStream in = new FileInputStream(src);
+        OutputStream out = new FileOutputStream(dst);
+
+        // Transfer bytes from in to out
+        byte[] buf = new byte[1024];
+        int len;
+        while ((len = in.read(buf)) > 0) {
+            out.write(buf, 0, len);
+        }
+        in.close();
+        out.close();
+    }
+
+    // Copied from http://stackoverflow.com/a/9989900/943814
+    private String getRealPathFromURI(Uri contentURI) {
+        String result;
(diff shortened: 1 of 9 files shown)


src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
```diff
@@ -0,0 +1,78 @@
...
+        int len;
+        while ((len = in.read(buf)) > 0) {
+            out.write(buf, 0, len);
+        }
+        in.close();
+        out.close();
+    }
+
+    // Copied from http://stackoverflow.com/a/9989900/943814
+    private String getRealPathFromURI(Uri contentURI) {
```
veniosg: If I'm not mistaken this way of loading cursors has been deprecated in favor of using a CursorLoader that will not block the UI but run asynchronously.

veniosg:
I don't have any authority on whether this goes in but can I suggest not autoformatting such large portions of existing files as it tends to pollute the history :-)

src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
anderspitman: I'll look in to that.

anderspitman:
Good suggestion. I'll see if I can back up and only change what I added.

anderspitman:
@veniosg Although once it's autoformatted the first time won't it pretty much be consistent after that?

veniosg:
Yeah, and we can always check the history if really needed; I just mentioned it as the actual change was a bit buried in the unnecessarily large different.

anderspitman:
Agreed it puts extra work on the reviewer. I'll see if @Nutomic is ok with this since it would be a pain to redo it, but in the future I'll avoid autoformatting or do it in a separate commit.

src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
```diff
@@ -0,0 +1,78 @@
...
+        this.files = files;
+    }
+
+    // Copied from http://stackoverflow.com/a/9293885/943814
+    public void copy(File src, File dst) throws IOException {
+        InputStream in = new FileInputStream(src);
+        OutputStream out = new FileOutputStream(dst);
+
+        // Transfer bytes from in to out
+        byte[] buf = new byte[1024];
```
Nutomic: You should change this to 8192 for better performance.

http://stackoverflow.com/a/237495
http://stackoverflow.com/a/11394667

src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
Nutomic: As the first answer on that stackoverflow question says, why do you even need the real path? I'd much prefer to avoid this hack. For the filename, have you checked if that is passed in the intent?

src/main/res/values/dimens.xml
```diff
@@ -0,0 +1,5 @@
+<resources>
+    <!-- Default screen margins, per the Android Design guidelines. -->
+    <dimen name="activity_horizontal_margin">16dp</dimen>
+    <dimen name="activity_vertical_margin">16dp</dimen>
+</resources>
```
Nutomic: Those margins seem way too big on my Nexus 5. I suggest you just get rid of them (and also the w820 ones).

src/main/res/menu/menu_share_to_folder.xml
```diff
@@ -0,0 +1,7 @@
+<menu xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    tools:context="com.nutomic.syncthingandroid.activities.ShareToFolderActivity">
+    <item android:id="@+id/action_settings" android:title="@string/action_settings"
+        android:orderInCategory="100" app:showAsAction="never" />
+</menu>
```
Nutomic: File is unused?

Nutomic:
As @veniosg said, could you please remove the autoformatted stuff from the manifest, as it makes it really difficult to review the actual changes.

anderspitman:
@Nutomic, thanks for the review. I'll address your comments and probably do a new PR rather than try and undo the autoformatting.

anderspitman:
Closed in favor of #447

src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
anderspitman: Copying the data works fine without the hack. However, so far I haven't been able to find a way to figure out the filename from what's included in the intent.

PR #447, opened 2015-08-21
Implement intent handler for sharing

anderspitman:
When other apps share files there's now an option to share
to syncthing


src/main/res/layout/activity_share_to_folder.xml
```diff
@@ -0,0 +1,10 @@
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools" android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    tools:context="com.nutomic.syncthingandroid.activities.ShareToFolderActivity">
+
+    <fragment android:name="com.nutomic.syncthingandroid.fragments.SelectFolderFragment"
+        android:id="@+id/share_to_folder_fragment"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent" />
+</RelativeLayout>
```
Nutomic: PLease add a newline at the end ;)

src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
```diff
@@ -0,0 +1,75 @@
...
+        int len;
+        while ((len = in.read(buf)) > 0) {
+            out.write(buf, 0, len);
+        }
+        in.close();
+        out.close();
+    }
+
+    // Copied from http://stackoverflow.com/a/9989900/943814
+    private String getRealPathFromURI(Uri contentURI) {
```
Nutomic: Like  Isaid, I'd prefer if you don't use this hack.

src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
```diff
@@ -0,0 +1,75 @@
...
+        } else {
+            cursor.moveToFirst();
+            int idx = cursor.getColumnIndex(MediaStore.Images.ImageColumns.DATA);
+            result = cursor.getString(idx);
+            cursor.close();
+        }
+        return result;
+    }
+
+    private ArrayList<Uri> files;
```
Nutomic: PLease put members at the beginning of the class, and name should be `mFiles`.

src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
```diff
@@ -0,0 +1,75 @@
...
+            }
+            catch (IOException e) {
+                Log.e(TAG, "Failed copying file", e);
+            }
+        }
+
+        getActivity().finish();
+    }
+
+    public void setFiles(ArrayList<Uri> files) {
```
Nutomic: This should be done in the constructor. Then you can also make the member final.

src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
```diff
@@ -0,0 +1,75 @@
...
+import java.io.OutputStream;
+import java.util.ArrayList;
+
+public class SelectFolderFragment extends FoldersFragment {
+
+    private static final String TAG = "SelectFolderFragment";
+
+    @Override
+    public void onItemClick(AdapterView<?> adapterView, View view, int i, long l) {
+        for (Uri uri : files) {
```
Nutomic: You are currently doing this on the GUI thread, but it should be done in a background thread (probably with an AsyncTask).

src/main/java/com/nutomic/syncthingandroid/fragments/FoldersFragment.java
```diff
@@ -26,7 +26,7 @@
 public class FoldersFragment extends ListFragment implements SyncthingService.OnApiChangeListener,
         AdapterView.OnItemClickListener, AdapterView.OnItemLongClickListener {
 
-    private FoldersAdapter mAdapter;
+    protected FoldersAdapter mAdapter;
```
Nutomic: A method `getItemPath(int item)` would be better for encapsulation.

Nutomic:
Please don't open a new pr when you change something, instead, just `git push -f` on your existing branch.

Also, it would be good if you call [RestApi#onFolderFileChange()](https://github.com/syncthing/syncthing-android/blob/master/src/main/java/com/nutomic/syncthingandroid/syncthing/RestApi.java#L943) after a file has been added.

anderspitman:
Yeah `git push -f` didn't work for some reason when I tried it.

src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
anderspitman: The files are not available until `ShareToActivityFolder`'s `onCreate` method is called. I have the `SelectFolderFragment` declared in XML so I don't have direct access to it's constructor. I'm assuming there's syntax for this but I don't know it?

src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
anderspitman: So far I haven't been able to find a way to determine the filename from what's contained in the intent.

src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
anderspitman: The files are not available until `ShareToFolderActivity`'s `onCreate` method is called. I'm not sure how to pass `SelectFolderFragment`'s constructor different parameters since it's declared in XML. What's the best way to handle this? Should I call the constructor programatically and infalte the XML?

anderspitman:
@Nutomic I believe I've addressed everything I didn't have questions about. Please see concerns [here](https://github.com/syncthing/syncthing-android/pull/447#discussion_r37672928) and [here](https://github.com/syncthing/syncthing-android/pull/447#discussion_r37678420).

src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
veniosg: Use of non-default constructors for Fragments is discouraged and bound to cause trouble. The official way to pass data and have them persisted for the lifetime of your fragment is by getting a reference to the Fragment and adding stuff to the arguments field. Use setArguments and getArguments and update the fields as needed. 
You should probably wrap the specifics inside the fragment so that callers don't have to worry about argument keys.

To get a reference fragment, check the fragment manager's find methods. There are many ways.

src/main/java/com/nutomic/syncthingandroid/activities/ShareToFolderActivity.java
```diff
@@ -0,0 +1,52 @@
...
+
+    private SelectFolderFragment mSelectFolderFragment;
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_share_to_folder);
+        mSelectFolderFragment = (SelectFolderFragment) getSupportFragmentManager()
+                .findFragmentById(R.id.share_to_folder_fragment);
+        mSelectFolderFragment .setHasOptionsMenu(false);
```
Nutomic: What's that space doing there before the dot? :p

src/main/java/com/nutomic/syncthingandroid/activities/ShareToFolderActivity.java
```diff
@@ -0,0 +1,52 @@
...
+
+        Intent intent = getIntent();
+        String action = intent.getAction();
+        String type = intent.getType();
+
+        ArrayList<Uri> uris = new ArrayList<>();
+        if (Intent.ACTION_SEND.equals(action) && type != null) {
```
Nutomic: Why are you checking the type? And if that's really needed, you should do something (at least logging) if it is null.

src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
Nutomic: Ah sorry, I had missed the fact that it's a Fragment.

src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
```diff
@@ -0,0 +1,108 @@
...
+            activity.getApi().onFolderFileChange(destFolder.id, destFolder.path);
+            activity.finish();
+            return null;
+        }
+
+        // Copied from http://stackoverflow.com/a/9989900/943814
+        private String getRealPathFromURI(Uri contentURI) {
+            String result;
+            Cursor cursor = getActivity().getContentResolver()
+                    .query(contentURI, null, null, null, null);
```
Nutomic: I just dumped this cursor for an image from the gallery, and got all kinds of data (`DatabaseUtils.dumpCursorToString(cursor)`)
```:
08-26 20:20:01.061    5899-6488/com.nutomic.syncthingandroid.debug D/SelectFolderFragment﹕ >>>>> Dumping cursor android.content.ContentResolver$CursorWrapperInner@3ab483f7
    0 {
    _id=20502
    _data=/storage/emulated/0/DCIM/Camera/IMG_20150824_164618.jpg
    _size=2480077
    _display_name=IMG_20150824_164618.jpg
    mime_type=image/jpeg
    title=IMG_20150824_164618
    date_added=xxx
    date_modified=xxx
    description=null
    picasa_id=null
    isprivate=null
    latitude=xxx
    longitude=xxx
    datetaken=xxx
    orientation=0
    mini_thumb_magic=1245638041573583436
    bucket_id=-1739773001
    bucket_display_name=Camera
    width=3200
    height=2368
    }
    <<<<<
```
So I don't think the hack is necessary. And you should probably pass a selection (instead of null).

src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private static final String TAG = "SelectFolderFragment";
+    private ArrayList<Uri> mFiles;
+    private boolean copying = false;
+
+    @Override
+    public void onItemClick(AdapterView<?> adapterView, View view, int i, long l) {
+        // Currently only handle one copy request per intent. So the UI will remain responsive
+        // but won't really do anything after the first click.
+        if (!copying) {
```
Nutomic: This is in case someone manages to do a second click before the activity is closed? Your comment isn't so clear to me ;)

src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
```diff
@@ -0,0 +1,108 @@
...
+                try {
+                    copyFile(sourceFile, destFile);
+                }
+                catch (IOException e) {
+                    Log.e(TAG, "Failed copying file", e);
+                }
+            }
+
+            SyncthingActivity activity = (SyncthingActivity) getActivity();
+            activity.getApi().onFolderFileChange(destFolder.id, destFolder.path);
```
Nutomic: This will force a rescan of the entire repo, it's probably better to request a rescan for each file individually (@Zillode @calmh please tell me if I'm wrong).

src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
```diff
@@ -0,0 +1,108 @@
...
+            activity.finish();
+            return null;
```
Nutomic: There's currently no indication that the file has been shared successfully. You should at least add a toast.

Also, if you don't need onPostExecute() of the AsyncTask, you could just use a Runnable. But you'll probably need it for the toast.

src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
Zillode: correct, use `&sub=folder/filename`

src/main/java/com/nutomic/syncthingandroid/fragments/SelectFolderFragment.java
anderspitman: I thought "the hack" was querying the database period. Whereas not using a hack would be somehow getting the filename information from the intent. What are you referring to?

Nutomic: Ah sorry, I didn't read your code right. Still, I think it's better to use the display name field instead of the path. For your cursor == null check, it might be cleaner if you do `Uri.getScheme().equals("file)`.

Nutomic:
Are you still working on this?

anderspitman:
Yeah. Been a bit behind lately. I'll try to look at it this weekend.

fti7:
Any Progress here? :-) important feature

anderspitman:
@fti7 I've gotten quite busy with school as of late. I don't think it would be too much work for me to get the current functionality merged in, but honestly I think it's use in the current state is fairly limited, because you can only share files to the root of a syncthing share. I really need to integrate a directory explorer to let the user choose where inside the share to copy the file before it's what we want I think.

Nutomic:
Closing this. Pleaase reopen when you update it.

PR #455, opened 2015-09-06
Material design 

### LINKED ISSUE #404: Material design?
When will you transition to material design?

veniosg:
Tried to keep this to a minimum but unfortunately it's become quite big already. 

Stuff included:
* Migration of folder editing to views.
* Updating and addition of icons. Now following latest guidelines. Also compressed all of them.
* Migration to Toolbar.
* Update of side nav to match guidelines and work on landscape.
* General styling/metrics/consistency updates all over the app.
* Minimum SDK is now 11.

Stuff not included (needed to close #404 IMHO):
* Move device editing layout to views.
* Update preference/settings styling.


Nutomic:
Looks newlines are fine in the other files (just checked manually because I couldn't find a good bash command). But maybe you can set your editor so it adds them automatically?

There are some lint warnings (`gradle lint`). I haven't checked which ones are from your code, but some definitely are :p 

It seems like orange as accent color doesn't really work. Now, some dialogs have two different colors (EditText and buttons), which looks really ugly. We should really pick a different accent color.

Btw it would be nice if we could find a material icon for the QR code (in DeviceSettingsFragment). Unfortunately, the Android page doesn't seem to have one.

For the future, it would be great if you could make more, smaller pull requests instead of these huge ones, makes everything easier. All that said, I really appreciate your contribution, design and code look really good :+1: 

@Zillode This changes min sdk to 11. Should we handle this somehow? We could make a release before merging that would show a warning to users below api version 11.

veniosg:
I updated this with the stuff you talked about in the comments.

I'll have a look at the warnings, some do seem new. Regarding the accent, I can keep the default blue/green accent the system provides for dialogs and use orange for everything else as you're right that mixing the two looks weird. I do think that orange works for the rest of the app though, since the accent colour is meant to pop and contrast the primary/primaryDark and as far as I can tell, this hue doesn't really clash with the primary blue we use :) How does this sound?

About the QR icon, I was supposed to check that when I restyle the device screen to match the folder one - wanted to make sure that one looks fine before moving further :)

The plan was to only PR the first commit or so but then I got carried away while on vacation and messed around with more stuff than I should have :P 

Thanks for the quick review, glad I could help!

Nutomic:
Honestly, I think it looks ugly and wrong to have different accent colors in the app and in dialogs. I think the green accent we had before worked better in this regard (but another color might also be fine).

The qr icon was just something I noticed, and wanted to mention ;)

veniosg:
Having different accent colours in the app and dialogs is what Google effectively points to. Additionally, by default, the system does not apply the app's accent to the dialogs, we do that on purpose (and as far as I remember, this doesn't happen on the released versions of the app - the dialog default is used). 
That being said, I will try to find a different accent that also fits dialogs without clashing with the "don't use red/orange for the dialog's buttons" guideline as I do agree that the current version looks off.

Nutomic:
I don't understand that you should have different colors. More that red/orange isn't good as an accent color.

veniosg:
Hmm, that's what I understood but just for the dialog, as they do use red accent in examples and apps (Inbox comes to mind). Maybe I'm misinterpreting though, I'm sure we can find a good looking non red/orange accent to use on everything :)

veniosg:
I've just tried a ton of combinations for the accent colour. Assuming we are to keep the primary colour as it is, the colours that work with it as accents (namely reds and yellows), don't fit being an accent for the dialog theme. Other colours just look plain weird (e.g. purples). 
To keep colour consistency within the app we could use the app's primary as the dialog's accent and use orange as the app's accent. 
I still think that greens make the screen look muddy due to their lack of contrast with the blue.
Let me know what you think about this.

P.S. A nice tool to quickly test combinations is https://www.materialpalette.com/.

Nutomic:
I think the blue fits well with the Syncthing logo and [website](https://syncthing.net/). How about blue with green/yellow/amber as accent?

But I'm also open for other suggestions.

veniosg:
Amber could work actually. Yellow I find to be too close to white to stand as a text colour for actions (talking about the dialog). I'll try them out this evening and possibly push. I agree on the blue :)

Zillode:
>@Zillode This changes min sdk to 11. Should we handle this somehow? We could make a release before merging that would show a warning to users below api version 11.

Nah, I think it is clearer to let the stores handle this for us, e.g. in F-droid the app will simply be grayed out.

Nutomic:
I meant as a warning for existing users that there won't be any new updates. But I prefer having the stores handle this.

veniosg:
accent
primarydark

Here are screenshots with using the app's accent and using the primaryDark respectively for the dialogs accent. From a usability and consistency with the rest of the system point of view I propose we use the second option. I can't see anything apart from blue/green/gray tones working for dialog text, especially considering the guideline mentioning colour for actions implying action state/severity.

Nutomic:
Yeah the second looks good, assuming the contrast is good enough.

veniosg:
I think it does in the app, yeah, plus we keep consistent in dialogs :) 
Am I forgetting anything for this pull request? I could include the device editor redesign but that can take a week or two.

Nutomic:
Better don't, it's way too big already :p 

I would ask you to squash your commits together somewhat, but that's probably more effort than it's worth. Will review your new changes this evening.

veniosg:
Fixed the comments on menu/folder_list and FolderPickerActivity. Also rebased on top of the upstream master to make merging easier.

Nutomic:
Thanks. But one more thing, could you put the suppress warnings in your last commit in the line before the cast, instead of on the function? Preferably in an amend. That way it's less likely we miss an actual problem.

And I promise this is my last complaint on this PR ;)

veniosg:
Oh sure, amended :)

Nutomic:
Merged, thanks a lot for your effort :)

veniosg:
Thanks for merging, glad I could contribute!

PR #459, opened 2015-09-20
Update Polish translation

konradmb:
Polish translation for syncthing_active and syncthing_disabled_title seems unnatural.


Nutomic:
Thanks, but we use [Transifex](https://www.transifex.com/syncthing-android/syncthing-android/) for translations. Can you please update it there? ;)

konradmb:
Ok, sorry :)

PR #461, opened 2015-09-28
Update device details styling

veniosg:
Updated the device details layout, style, icons and code to match the new folder details ones. 
This migrates FolderSettingsFragment to views.
There's also a fix for the actionmode bar, which would previously push the toolbar down when text was selected.


Nutomic:
Looks very good! 

I feel like the compression dialog should have a dismiss button, but I'm not sure (I remember a cancel button was recommended somewhere in the docs, but I couldn't find it).

Would it be possible to make the device id always multiline? Not so important, but should make editing easier.

The design of the EditTexts makes it somewhat hard to tell the difference from TextViews. Especially when adding a device, "dynamic" looks like a TextView (and it doesn't help that there's no label for it). Maybe we could leave Addresses empty by default, which would imply "dynamic".

veniosg:
Glad you like this :) 

I omitted the negative action button from the dialog as there are already two ways to cancel.

I agree on the EditText issue. I like this solution a lot, sounds intuitive from a user point of view and it shouldn't be too hard to implement. I'll add this one along with the rest of the suggestions.

veniosg:
Updated :)

src/main/java/com/nutomic/syncthingandroid/fragments/FolderFragment.java
```diff
@@ -76,7 +77,7 @@
 
     private TextView mVersioningKeepView;
 
-    private boolean mIsCreate;
+    private boolean mCreateMode;
```
Nutomic: I think `mIsCreateMode` is better for a boolean ("is" showing that it's either true or false). But not really important.

src/main/java/com/nutomic/syncthingandroid/util/TextWatcherAdapter.java
```diff
@@ -0,0 +1,16 @@
...
+
+    @Override
+    public void beforeTextChanged(CharSequence s, int start, int count, int after) {}
+
+    @Override
+    public void onTextChanged(CharSequence s, int start, int before, int count) {}
+
+    @Override
+    public void afterTextChanged(Editable s) {}
+}
```
Nutomic: Newline. There should be a setting in Android Studio ;)

src/main/res/layout/item_device_form.xml
```diff
@@ -1,12 +1,10 @@
...
+    style="@style/Widget.Syncthing.TextView.Label.Details.DeviceList"
     android:layout_width="match_parent"
-    android:layout_height="48dp"
+    android:layout_height="wrap_content"
     android:layout_marginLeft="@dimen/material_divider_inset"
     android:layout_marginStart="@dimen/material_divider_inset"
-    android:layout_marginRight="@dimen/abc_action_bar_content_inset_material"
-    android:layout_marginEnd="@dimen/abc_action_bar_content_inset_material"
     android:textAppearance="@style/TextAppearance.AppCompat.Caption"
     tools:ignore="RtlHardcoded,RtlSymmetry" />
```
Nutomic: Here too :D

veniosg:
Amended the last commit with the latest suggestions. 
Also, finally found and enabled the newline thing :D

src/main/java/com/nutomic/syncthingandroid/fragments/DeviceFragment.java
```diff
@@ -0,0 +1,405 @@
...
+        }
+    }
+
+    @Override
+    public void onStop() {
+        super.onStop();
+
+        // We don't want to update every time a TextView's character changes,
+        // so we hold off until the view stops being visible to the user.
+        updateDevice();
```
Nutomic: I just tried to share a device ID. Right after selecting an app to share to, it went back to Syncthing and showed a restart dialog.

You should move this to onPause so it shows earlier, but also check if the device was edited at all.

Also, flipping the Introducer setting in an existing device creates 2 or 3 restart dialogs in a row.

veniosg: Oh wow. I hadn't tried to share but this sounds pretty weird. I'm pretty sure I have tried the introduced switch though and don't remember getting more than one dialogs. Could it be a combination of the two? You are right about the check, completely slipped my mind because the previous logic didn't need it. 

I was under the impression the API would handle the restart dialog display conditions and would not show it if the user has already dismissed another dialog previously. 
I will try to reproduce and fix these issues and try as many weird flows I can think of so we are extra safe. 

Sorry about this, major oversight.

Nutomic: It works correctly if there was a restart dialog before (notification is shown), but if there was none, a new dialog is created (it's probably happening so fast that it doesn't detect the notification yet).

I think if you do the things in my second paragraph, this should be fixed.

And no worries, that's exactly why we do code reviews :)

veniosg:
Seeing the code, it looks like if you quickly toggle the introducer setting you'd get multiple restart dialogs even on the currently released version as the postponedRestart flag is what controls whether RestartActivity will get launched and that is only set after the user interacts with the dialog, not onCreate.
While implementing checking of whether the device has changed and only then restarting will block this behaviour, I feel like it's hiding the issue that the RestartActivity can be launched more than once. 
I'd say that the activity should call getApi().setRestartPostponed() in onCreate tbh.

BTW this happens even before my changes, aka on the released version :)

Nutomic:
Thing is, I haven't seen this problem before (and it hasn't been reported). But if you can fix the underlying issue, that's perfect!

veniosg:
Yeah, I guess no one really smashes the button that way; I tried it on my play store installation though and it happened. Great, I'll handle the general case then.

veniosg:
What I ended up doing is use singleTop on the RestartActivity as getApi() returns null in onCreate() and onServiceConnected might be too late to ensure only one dialog shows. 
With these changes I wasn't able to reproduce using the steps you posted; or any other way I attempted. 
I've also rebased.

Nutomic:
Great!

Now just squash your commits, and I'll merge it. And sorry for the delay, I was busy.

veniosg:
Perfect, done! No worries about the delay :)

Nutomic:
Thanks a lot, merged :)

veniosg:
Happy to help!

PR #466, opened 2015-10-15
Fix inotify

Nutomic:
The parameters set in RestApi#onFolderFileChanged() were not actually
used by PostTask#doInBackground().

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/PostConfigTask.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/PostConfigTask.java
@@ -14,41 +14,37 @@ import org.apache.http.protocol.HTTP;
 import java.io.IOException;
 
 /**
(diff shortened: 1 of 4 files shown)


Zillode:
:+1: 
Syncthing accepts multiple sub arguments, maybe we can accumulate them into a single request? This is what syncthing-inotify does.
```
urlParams.add(new BasicNameValuePair("sub", params[3]));
urlParams.add(new BasicNameValuePair("sub", params[4]));
...
```

Nutomic:
But that needs some kind of caching, right? (eg you have to wait for a few ms for any additional paths)

Unless you have a simple solution for that, I'd say it's not worth the effort for now.

AudriusButkevicius:
You can send an array of paths

Nutomic:
I mean Android gives us a seperate event for each changed path. That means we'd have to collect those paths somewhere, before we make a scan request to Syncthing.

PR #469, opened 2015-10-22
Update Polish language translation.

DEVoytas:
- Translate missing strings
- Make it more diff tools friendly by updating white space formatting
  to match src/main/res/values/strings.xml file. It's much easier
  to track the changes and spot not translated stings when we have
  one to one line equivalent (including line numbers).

### DIFF
--- a/src/main/res/values-pl/strings.xml
+++ b/src/main/res/values-pl/strings.xml
@@ -1,264 +1,454 @@
-<?xml version='1.0' encoding='UTF-8'?>
+<?xml version="1.0" encoding="utf-8"?>
 <resources>
-  <string name="app_name">Syncthing</string>
-  <string name="app_description">Otwarta, godna zaufania oraz zdecentralizowana aplikacja do synchronizacji plików.</string>
-  <!--MainActivity-->
-  <!--Title of the "add folder" menu action-->
-  <string name="add_folder">Dodaj folder</string>
-  <!--Title of the "share device id" menu action-->
-  <string name="share_device_id">Udostępnij ID urządzenia</string>
-  <!--Shown in the chooser dialog when sharing a Device ID-->
-  <string name="send_device_id_to">Wyślij ID urządzenia do</string>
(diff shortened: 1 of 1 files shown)


chucic:
I think you have to use Transifex for syncthing-android translations...

DEVoytas:
Notes for reviewer:
- Due do lot of white space changes noise (see the commit message why) I recommend to ignore white space and blank line changes in diff view. To ignore white space in github viewer just add `?w=1` at the end of URL, like this:
[https://github.com/syncthing/syncthing-android/pull/469/files**?w=1**](https://github.com/syncthing/syncthing-android/pull/469/files?w=1)
Unfortunately I don't know whether github can ignore blank lines.
- Also, for sanity check, comparison with reference file ```src/main/res/values/strings.xml``` may be useful - there should be line to line equivalence including line numbering.
Again I am not aware of github way of doing it, but in command line it goes like this:
```git diff master:src/main/res/values/strings.xml polish-translation:src/main/res/values-pl/strings.xml```

DEVoytas:
@chucic, is usage of Transifex obligatory or supplementary ? I think you need to create account there to use it.

veniosg:
Pretty sure it's obligatory as this helps keep the flow of translations simple since you don't have to remember to reimport translations to transifex and then pull them back in to keep the format consistent. Not even sure it's supported in an automatic way.

chucic:
I don't know, I'm not a contributor, but according to https://github.com/syncthing/syncthing-android/wiki/Translations I think it is obligatory...

DEVoytas:
Well, I thought I was something introduced to help with translations (i.e. additional source not a replacement) and not to make it more difficult. At the moment I am not able to contribute through Transifex as I am not registered there (which I may consider to do later but not now).
For now I will hold on with closing the pull request until @Nutomic confirm that he doesn't want it.

Nutomic:
Yeah sorry, we only do translations through Transifex. Otherwise, things will get overwritten when I pull translations from Transifex again.

Issue #473, opened 2015-10-30
Showing same status info for all devices in device list, 0.7.0-beta2


chucic:
Testing latest beta, my device list shows same status for all devices in device list, so if first connects, it shows syncing (xy%) for all, then "Up to date" for all etc even for devices that are not connected (because they are not even running). I could get some log, if you want (but log feature is also broken, #470 )

Nutomic:
Right, I can reproduce that.

@calmh @AudriusButkevicius I assume this is due to a change in `/rest/system/connections`. [According to the documentation](https://github.com/syncthing/docs/commit/8c76d69c467e4af0ebba646efa5a3e09d11ec63b), there's some additional data added, but nothing that would cause a problem (and I'm not getting parsing errors either). Any idea what causes this?

AudriusButkevicius:
So I wonder if its a bug on our side?
Can you actually get the rest call output as it happens?

Issue #470, opened 2015-10-26
Beta issues


AudriusButkevicius:
Ann servers shows 0/0 though web ui shows 5/8 (with only 6 beind defined?)
Logs for android and syncthing are missing (needs root?)
Opening web ui, going back and opening settings crashes the app

Nutomic:
The number of announce servers seems to be a Syncthing issue. I tried a clean install on Ubuntu, and got the same numbers.

I can't reproduce the crash, can you post a logcat?

AudriusButkevicius:
```
D/AndroidRuntime( 7863): Shutting down VM
E/AndroidRuntime( 7863): FATAL EXCEPTION: main
E/AndroidRuntime( 7863): Process: com.nutomic.syncthingandroid, PID: 7863
E/AndroidRuntime( 7863): java.lang.RuntimeException: Invalid usage report value
E/AndroidRuntime( 7863):        at com.nutomic.syncthingandroid.syncthing.RestApi.getUsageReportAccepted(RestApi.java:986)
E/AndroidRuntime( 7863):        at com.nutomic.syncthingandroid.fragments.SettingsFragment.onApiChange(SettingsFragment.java:79)
E/AndroidRuntime( 7863):        at com.nutomic.syncthingandroid.syncthing.SyncthingService.registerOnApiChangeListener(SyncthingService.java:443)
E/AndroidRuntime( 7863):        at com.nutomic.syncthingandroid.fragments.SettingsFragment.onServiceConnected(SettingsFragment.java:184)
E/AndroidRuntime( 7863):        at com.nutomic.syncthingandroid.activities.SyncthingActivity.onServiceConnected(SyncthingActivity.java:51)
E/AndroidRuntime( 7863):        at android.app.LoadedApk$ServiceDispatcher.doConnected(LoadedApk.java:1208)
E/AndroidRuntime( 7863):        at android.app.LoadedApk$ServiceDispatcher$RunConnection.run(LoadedApk.java:1225)
E/AndroidRuntime( 7863):        at android.os.Handler.handleCallback(Handler.java:739)
E/AndroidRuntime( 7863):        at android.os.Handler.dispatchMessage(Handler.java:95)
E/AndroidRuntime( 7863):        at android.os.Looper.loop(Looper.java:135)
E/AndroidRuntime( 7863):        at android.app.ActivityThread.main(ActivityThread.java:5254)
[...]

PR #478, opened 2015-11-08
Bump 0.12.1

Zillode:


dper:
Would this let the Android client work with other 0.12 clients? Because that would be just amazing.

licaon-kter:
@dper 0.7.0-betas work ok, already.

chucic:
@licaon-kter Not for me: #473

licaon-kter:
Oh, ok, I only saw https://github.com/syncthing/syncthing-gtk/issues/216 on my setup, but I only have 1 host.

chucic:
Yes, I think it's same root cause. Apart from that it's working fine ;-)

Nutomic:
Looks good!

Can you reference #470 in the commit message before merging?

PR #495, opened 2015-11-13
Use CGO to build Syncthing with support for DNS

Zillode:
```
export ANDROID_NDK=/opt/android_ndk
./make-all.bash
./gradlew assembleDebug
```


make-all.bash
```diff
@@ -0,0 +1,14 @@
...
+
+./make-go.bash arm
+./make-syncthing.bash arm
+
+./make-go.bash 386
+./make-syncthing.bash 386
+
+# Will need to bump target API to support this
+#./make-go.bash amd64
+#./make-syncthing.bash amd64
```
Nutomic: What's the reason for this? 64 bit ARM should be supported in Lollipop. Generally, I don't really like having dead code around.

Nutomic:
Yeah, it works :D 

Can you add missing newlines at the end of files, and add a commit to remove `ConfigXml#replaceAnnounceServers` and related code? I'll then make a beta release.

make-all.bash
```diff
@@ -0,0 +1,14 @@
...
+
+# Build the syncthing library
+
+./make-go.bash arm
+./make-syncthing.bash arm
+
+./make-go.bash 386
+./make-syncthing.bash 386
+
+# Will need to bump target API to support this
```
Nutomic: The comment is still here, doesn't seem to make sense? :p

Nutomic:
Why did you remove the gradle build/clean tasks? And as I said, can you link to the original source for the golang patches?

Zillode:
I removed the clean task because the build&copy tasks are also removed. You simply have to run make once and keep that data. This allows much shorter gradle compilation times.

The original source for the golang patches are marked in the files:
```
+From df553e646c597ab7fb26e3ffa2b730c04ab069f8 Mon Sep 17 00:00:00 2001
+From: Hyang-Ah Hana Kim <hyangah@gmail.com>
+Date: Thu, 5 Nov 2015 16:32:27 -0500
+Subject: [PATCH 1/4] cmd,runtime: TLS setup for android/386
+
+Same ugly hack as https://go-review.googlesource.com/15991.
+
+Update golang/go#9327.
+
+Change-Id: I58284e83268a15de95eabc833c3e01bf1e3faa2e
```

Nutomic:
I don't see your point regarding the gradle tasks. In my experience, buildNative is only invoked when the native code changes, or after a clean build. Did you have a different experience, or does this PR change something there?

[Patches are here for reference](https://go-review.googlesource.com/#/c/16678/4).

Nutomic:
As I said on IRC, I'd prefer if you only remove the dependsOn/finalizedBy.

Also, we should change the announce servers back to default (but I can do that myself after merging).

Zillode:
Note: after merging we should update the scripts on jenkins

Nutomic:
Which scripts do you mean?

Zillode:
All targets of http://android.syncthing.net/view/Syncthing-Android/ need to run `./make-all.bash`.

Nutomic:
`gradle buildNative assembleDebug` should be enough ;)

PR #497, opened 2015-11-14
Fix "Wifi only" / "Charging only" not being applied immediately

capi:
When turning on the "Sync only on wifi" option while on a mobile network, this may fail, if no other setting that causes a call to `mSyncthingService.updateState()` is also triggered.

The root cause seems to be that [`SettingsFragment`](https://github.com/syncthing/syncthing-android/blob/master/src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java) line 231 is called in `onPreferenceChange()` which is called _before_ the setting is persisted.

I think, it would be better to remove the call to `mSyncthingService.updateState()` from `SettingsFragment` and enhance `onSharedPreferenceChanged()` directly in the service.


Nutomic:
Thanks :)

PR #501, opened 2015-11-16
Add restriction for allowed WiFi SSIDs when WiFi-only sync

capi:
I encountered several networks that were not really "happy" about the unknown Syncthing traffic while I was using "Always sync" and "Sync only on wifi". This pull-requests adds a setting for creating a white-list of allowed WiFi SSIDs, where syncthing will only consider the Wifi for automatic syncing if the current SSID is in the white-list.

By default, the list is empty, resulting in the current behavior of "any connected (unmetered) WiFi".

The current editor being used is very simple (`EditTextPreference`), where each SSID has to be entered on a new line (WiFi SSIDs may have blanks). Also, it requires addition of `android.permission.ACCESS_WIFI_STATE` for reading the current network's SSID.

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java
+++ b/src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java
@@ -15,16 +15,20 @@ import android.preference.PreferenceManager;
 import android.preference.PreferenceScreen;
 import android.support.v4.app.NavUtils;
 import android.text.InputType;
+import android.text.TextUtils;
 import android.util.Log;
 import android.view.MenuItem;
 import android.widget.Toast;
 
 import com.nutomic.syncthingandroid.R;
 import com.nutomic.syncthingandroid.activities.SyncthingActivity;
+import com.nutomic.syncthingandroid.preferences.WifiSsidPreference;
 import com.nutomic.syncthingandroid.syncthing.RestApi;
 import com.nutomic.syncthingandroid.syncthing.SyncthingService;
 
 import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
 
 import eu.chainfire.libsuperuser.Shell;
 
@@ -52,6 +56,7 @@ public class SettingsFragment extends PreferenceFragment
     private CheckBoxPreference mAlwaysRunInBackground;
     private CheckBoxPreference mSyncOnlyCharging;
     private CheckBoxPreference mSyncOnlyWifi;
+    private WifiSsidPreference mSyncOnlyOnSSIDs;
     private CheckBoxPreference mUseRoot;
     private PreferenceScreen mOptionsScreen;
     private PreferenceScreen mGuiScreen;
@@ -126,6 +131,8 @@ public class SettingsFragment extends PreferenceFragment
         mSyncOnlyCharging = (CheckBoxPreference)
                 findPreference(SyncthingService.PREF_SYNC_ONLY_CHARGING);
         mSyncOnlyWifi = (CheckBoxPreference) findPreference(SyncthingService.PREF_SYNC_ONLY_WIFI);
+        mSyncOnlyOnSSIDs = (WifiSsidPreference) findPreference(SyncthingService.PREF_SYNC_ONLY_WIFI_SSIDS);
+        mSyncOnlyOnSSIDs.setDefaultValue(new TreeSet<String>()); // default to empty list
         mUseRoot = (CheckBoxPreference) findPreference(SyncthingService.PREF_USE_ROOT);
         Preference appVersion = screen.findPreference(APP_VERSION_KEY);
         mOptionsScreen = (PreferenceScreen) screen.findPreference(SYNCTHING_OPTIONS_KEY);
@@ -144,6 +151,7 @@ public class SettingsFragment extends PreferenceFragment
         mAlwaysRunInBackground.setOnPreferenceChangeListener(this);
         mSyncOnlyCharging.setOnPreferenceChangeListener(this);
         mSyncOnlyWifi.setOnPreferenceChangeListener(this);
+        mSyncOnlyOnSSIDs.setOnPreferenceChangeListener(this);
         mUseRoot.setOnPreferenceClickListener(this);
         screen.findPreference(EXPORT_CONFIG).setOnPreferenceClickListener(this);
         screen.findPreference(IMPORT_CONFIG).setOnPreferenceClickListener(this);
@@ -153,8 +161,9 @@ public class SettingsFragment extends PreferenceFragment
         sttrace.setOnPreferenceChangeListener(this);
 
         // Force summary update and wifi/charging preferences enable/disable.
-        onPreferenceChange(mAlwaysRunInBackground, mAlwaysRunInBackground.isChecked());
         SharedPreferences sp = PreferenceManager.getDefaultSharedPreferences(getActivity());
+        onPreferenceChange(mAlwaysRunInBackground, mAlwaysRunInBackground.isChecked());
+        onPreferenceChange(mSyncOnlyOnSSIDs, sp.getStringSet("sync_only_wifi_ssids_set", new TreeSet<String>()));
         user.setSummary(sp.getString("gui_user", ""));
         sttrace.setSummary(sp.getString("sttrace", ""));
     }
@@ -236,11 +245,19 @@ public class SettingsFragment extends PreferenceFragment
                     : R.string.always_run_in_background_disabled);
             mSyncOnlyCharging.setEnabled(value);
             mSyncOnlyWifi.setEnabled(value);
+            mSyncOnlyOnSSIDs.setEnabled(mSyncOnlyWifi.isChecked());
             // Uncheck items when disabled, so it is clear they have no effect.
             if (!value) {
                 mSyncOnlyCharging.setChecked(false);
                 mSyncOnlyWifi.setChecked(false);
             }
+        } else if (preference.equals(mSyncOnlyWifi)) {
+            mSyncOnlyOnSSIDs.setEnabled((Boolean) o);
+        } else if (preference.equals(mSyncOnlyOnSSIDs)) {
+            String ssids = formatWifiNameList((Set<String>) o);
+            mSyncOnlyOnSSIDs.setSummary(ssids.isEmpty()
+                    ? getString(R.string.sync_only_wifi_ssids_all)
+                    : getString(R.string.sync_only_wifi_ssids_values, ssids));
         } else if (preference.getKey().equals(DEVICE_NAME_KEY)) {
             RestApi.Device old = mSyncthingService.getApi().getLocalDevice();
             RestApi.Device updated = new RestApi.Device();
@@ -295,6 +312,14 @@ public class SettingsFragment extends PreferenceFragment
             mSyncthingService.getApi().requireRestart(getActivity());
 
         return true;
+    }
+
+    private String formatWifiNameList(Set<String> ssids) {
+        Set<String> formatted = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
+        for (String ssid : ssids) {
+            formatted.add(ssid.replaceFirst("^\"", "").replaceFirst("\"$", ""));
+        }
+        return TextUtils.join(", ", formatted);
     }
 
     /**
--- a/src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
+++ b/src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
@@ -1,0 +1,129 @@
+package com.nutomic.syncthingandroid.preferences;
+
+import android.content.Context;
+import android.net.wifi.WifiConfiguration;
+import android.net.wifi.WifiManager;
+import android.os.Bundle;
+import android.preference.MultiSelectListPreference;
+import android.util.AttributeSet;
+import android.widget.ListView;
+import android.widget.Toast;
+
+import com.nutomic.syncthingandroid.R;
+
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * MultiSelectListPreference which allows the user to select on which WiFi networks (based on SSID)
+ * syncing should be allowed.
+ *
+ * Setting can be "All networks" (none selected), or selecting individual networks.
+ *
+ * Due to restrictions in Android, it is possible/likely, that the list of saved WiFi networks
+ * cannot be retrieved if the WiFi is turned off. In this case, an explanation is shown.
+ *
+ * The preference is stored as Set&lt;String&gt; where an empty set represents
+ * "all networks allowed".
+ *
+ * SSIDs are formatted according to the naming convention of WifiManager, i.e. they have the
+ * surrounding double-quotes (") for UTF-8 names, or they are hex strings (if not quoted).
+ */
+public class WifiSsidPreference extends MultiSelectListPreference {
+
+    public WifiSsidPreference(Context context, AttributeSet attrs) {
+        super(context, attrs);
+    }
+
+    public WifiSsidPreference(Context context) {
+        super(context);
+    }
+
+    /**
+     * Show the dialog if WiFi is available and configured networks can be loaded.
+     * Otherwise will display a Toast requesting to turn on WiFi.
+     *
+     * On opening of the dialog, will also remove any SSIDs from the set that have been removed
+     * by the user in the WiFi manager. This change will be persisted only if the user changes
+     * any other setting
+     */
+    @Override
+    protected void showDialog(Bundle state) {
+        WifiConfiguration[] networks = loadConfiguredNetworksSorted();
+        if (networks != null) {
+            Set<String> selected = getSharedPreferences().getStringSet(getKey(), new HashSet<String>());
+            // from JavaDoc: Note that you must not modify the set instance returned by this call.
+            // therefore required to make a defensive copy of the elements
+            selected = new HashSet<>(selected);
+            CharSequence[] all = extractSsid(networks, false);
+            filterRemovedNetworks(selected, all);
+            setEntries(extractSsid(networks, true)); // display without surrounding quotes
+            setEntryValues(all); // the value of the entry is the SSID "as is"
+            setValues(selected); // the currently selected values (without meanwhile deleted networks)
+            super.showDialog(state);
+        } else {
+            Toast.makeText(getContext(), R.string.sync_only_wifi_ssids_wifi_turn_on_wifi, Toast.LENGTH_LONG).show();
+        }
+    }
+
+    /**
+     * Removes any network that is no longer saved on the device. Otherwise it will never be
+     * removed from the allowed set by MultiSelectListPreference.
+     */
+    private void filterRemovedNetworks(Set<String> selected, CharSequence[] all) {
+        HashSet<CharSequence> availableNetworks = new HashSet<>(Arrays.asList(all));
+        selected.retainAll(availableNetworks);
+    }
+
+    /**
+     * Converts WiFi configuration to it's string representation, using the SSID.
+     *
+     * It can also remove surrounding quotes which indicate that the SSID is an UTF-8
+     * string and not a Hex-String, if the strings are intended to be displayed to the
+     * user, who will not expect the quotes.
+     *
+     * @param configs the objects to convert
+     * @param stripQuotes if to remove surrounding quotes
+     * @return the formatted SSID of the wifi configurations
+     */
+    private CharSequence[] extractSsid(WifiConfiguration[] configs, boolean stripQuotes) {
+        CharSequence[] result = new CharSequence[configs.length];
+        for (int i = 0; i < configs.length; i++) {
+            // WiFi SSIDs can either be UTF-8 (encapsulated in '"') or hex-strings
+            if (stripQuotes)
+                result[i] = configs[i].SSID.replaceFirst("^\"", "").replaceFirst("\"$", "");
+            else
+                result[i] = configs[i].SSID;
+        }
+        return result;
+    }
+
+    /**
+     * Load the configured WiFi networks, sort them by SSID.
+     *
+     * @return a sorted array of WifiConfiguration, or null, if data cannot be retrieved
+     */
+    private WifiConfiguration[] loadConfiguredNetworksSorted() {
+        WifiManager wifiManager = (WifiManager) getContext().getSystemService(Context.WIFI_SERVICE);
+        if (wifiManager != null) {
+            List<WifiConfiguration> configuredNetworks = wifiManager.getConfiguredNetworks();
+            // if WiFi is turned off, getConfiguredNetworks returns null on many devices
+            if (configuredNetworks != null) {
+                WifiConfiguration[] result = configuredNetworks.toArray(new WifiConfiguration[configuredNetworks.size()]);
+                Arrays.sort(result, new Comparator<WifiConfiguration>() {
+                    @Override
+                    public int compare(WifiConfiguration lhs, WifiConfiguration rhs) {
+                        return lhs.SSID.compareToIgnoreCase(rhs.SSID);
+                    }
+                });
+                return result;
+            }
+        }
+        // WiFi is turned off or device doesn't have WiFi
+        return null;
+    }
+
+}
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/DeviceStateHolder.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/DeviceStateHolder.java
@@ -5,6 +5,8 @@ import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.net.ConnectivityManager;
+import android.net.wifi.WifiInfo;
+import android.net.wifi.WifiManager;
 import android.os.BatteryManager;
 
 /**
@@ -26,17 +28,25 @@ public class DeviceStateHolder extends BroadcastReceiver {
      */
     public static final String EXTRA_IS_CHARGING = "is_charging";
 
+    private Context mContext;
+
     private boolean mIsWifiConnected = false;
+
+    private String mWifiSsid;
 
     private boolean mIsCharging = false;
 
     @TargetApi(16)
     public DeviceStateHolder(Context context) {
+        mContext = context;
         ConnectivityManager cm = (ConnectivityManager)
                 context.getSystemService(Context.CONNECTIVITY_SERVICE);
         mIsWifiConnected = cm.getNetworkInfo(ConnectivityManager.TYPE_WIFI).isConnected();
         if (android.os.Build.VERSION.SDK_INT >= 16 && cm.isActiveNetworkMetered())
             mIsWifiConnected = false;
+        if (mIsWifiConnected) {
+            updateWifiSsid();
+        }
     }
 
     /**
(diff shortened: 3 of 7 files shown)


capi:
Updated the PR to include a UI for selecting the WiFi SSIDs based on saved networks on the device

capi:
@Nutomic I made the UI resemble a `MultiSelectListPreference` as closely as possible as converting it to one shows several problems with dynamic updates (updates to WiFi list are only available the next time the preference is opened).

Please review the current solution.

src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java
```diff
@@ -297,6 +316,22 @@ public boolean onPreferenceChange(Preference preference, Object o) {
...
+            StringBuilder sb = new StringBuilder();
+            for (String ssid : ssids) {
+                if (sb.length() > 0) {
+                    sb.append(", ");
+                }
+                sb.append(ssid.replaceFirst("^\"", "").replaceFirst("\"$", ""));
+            }
+            return sb.toString();
+        }
+        return null;
```
Nutomic: You can probably shorten this with [TextUtils#join()](http://developer.android.com/reference/android/text/TextUtils.html#join%28java.lang.CharSequence,%20java.lang.Iterable%29). Also, I'd prefer to use empty lists instead of null (needs less error checking).

src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java
```diff
@@ -392,6 +427,13 @@ public void onClick(DialogInterface dialogInterface, int i) {
                         })
                         .show();
                 return true;
+            case SyncthingService.PREF_SYNC_ONLY_WIFI_SSIDS:
+                WifiManager wifiManager = (WifiManager) getActivity().getSystemService(Context.WIFI_SERVICE);
+                if (wifiManager == null || wifiManager.getConfiguredNetworks() == null) {
+                    mSyncOnlyOnSSIDs.getDialog().dismiss();
```
Nutomic: This line doesn't make sense, you can only click the preference while the dialog is closed.

capi: It does, the dialog is actually shown before the click handler is called, so this is necessary to close the dialog to only show the Toast.

src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java
capi: Will change accordingly.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
```diff
@@ -0,0 +1,257 @@
...
+ * Due to restrictions in Android, it is possible/likely, that the list of saved WiFi networks
+ * cannot be retrieved if the WiFi is turned off. In this case, an explanation is shown.
+ *
+ * The preference is stored as Set&lt;String&gt; where <code>null</code> represents
+ * "all networks allowed", and a set represents a restriction to the selected SSIDs.
+ *
+ * SSIDs are formatted according to the naming convention of WifiManager, i.e. they have the
+ * surrounding double-quotes (") for UTF-8 names, or they are hex strings (if not quoted).
+ */
+public class WifiSsidPreference extends DialogPreference {
```
Nutomic: What's the reason you're not extending ListPreference? That would give us all the list layout etc for free.

Also, please remove the `// if` etc comments in this file, they are really unnecessary imo.

src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java
Nutomic: Yeah of course. Also see my edit above.

src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java
Nutomic: Ah I see.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
```diff
@@ -0,0 +1,257 @@
...
+                                mSelectedSsids.remove(ssid);
+                            } // if/else
+                            updateSaveButtonState();
+                        }
+                    });
+                }
+                return convertView;
+            } else {
+                // WiFi is most likely turned off
+                LayoutInflater inflater = (LayoutInflater) getContext().getSystemService(Context.LAYOUT_INFLATER_SERVICE);
```
Nutomic: When is this actually used? Wifi disabled is already handled in SettingsFragment.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
capi: That it's not possible to dynamically fill the content of the `MultiSelectListPreference` so that it is shown on _current_ invocation. It will only always show the _next_ time you open the preference, because the dialog's `ListView` and adapter are created before the `onPreferenceClick()` is called.

Regarding the `// if`, sure, they slipped from coding-style we use at work, didn't notice I still had so many left in the code.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
capi: Actually it's a fallback for a very unlikely race-condition where the WiFi service becomes unavailable between the check in `SettingsFragment` and actually building the list here. Rather unlikely, but since I already had the code from the previous iterations, I left it in as a fallback.

Nutomic:
The behaviour is a bit weird when (un)checking the currently connected wifi, or changing the "sync only on wifi" setting while connected to a wifi: The app is just closed without any feedback, which looks like an error. But this only happens sometimes, not sure why.

capi:
Ok, this should not happen, because I tested it a lot, removing the current WiFi was my default test-case. By any chance you got an `adb log` of the exception?

Nutomic:
There wasn't any exception (or maybe I missed it). I'll let you know if I can reproduce it again.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
Nutomic: How about you check for wifi only when the dialog is opened, and if there's no wifi, you close the dialog and show a toast. Would avoid the code duplication.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
Nutomic: I think it should work if you override the right methods (but I've never used that class tbh).

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
capi: Separation of concerns, the dialog should work even though it is not dismissed at once by the caller if WiFi is unavailable. Also there is a very small race-condition that between the check in `SettingsFragment` and building the list in the preference the WiFi service becomes unavailable.

capi:
Updated PR based on your recommendations, where possible.

If you find a way to use a native `MultiSelectListPreference`, please let me know. It seems easy, but it gets ugly really fast, since most data-related methods are only called after the dialog has already been created and the data has been bound already, so you end up with the data of the previous invocation.

From my side this is it. I would not recommend removing more safety-checks and null-checks and fall-backs for the sake of shortness of code. I tried to be as close to your coding style as possible, I hope no more `// if` or the like slipped through.

Regarding the crash regarding turning off WiFi-only sync / currently selected WiFi, I have been tormenting my device for about an hour now, I could not trigger a crash here, so without a stack-trace I have no idea where this should be. If it is related to WiFi-only sync as well, I suspect it to be more in the startup/shutdown area of the native `syncthing.so` because nothing in this area has been changed.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
```diff
@@ -0,0 +1,249 @@
...
+                return result;
+            }
+        }
+        // WiFi is turned of or device doesn't have WiFi
+        return null;
+    }
+
+    @Override
+    protected void showDialog(Bundle state) {
+        super.showDialog(state);
```
Nutomic: You can avoid showing the dialog by not calling the superclass implementation here. So no more checking in SettingsFragment ;)

Nutomic:
Regarding ListPreference, I found [this Stackoverflow answer](http://stackoverflow.com/a/10121132) which seems to work perfectly (using MultiSelectListPreference).

[my test implementation](https://gist.github.com/Nutomic/b40eb047002715f54e57)

src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java
```diff
@@ -21,10 +24,14 @@
...
 import com.nutomic.syncthingandroid.R;
 import com.nutomic.syncthingandroid.activities.SyncthingActivity;
+import com.nutomic.syncthingandroid.preferences.WifiSsidPreference;
 import com.nutomic.syncthingandroid.syncthing.RestApi;
 import com.nutomic.syncthingandroid.syncthing.SyncthingService;
 
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
```
Nutomic: Any reason you're using both HashSet and TreeSet?

src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java
```diff
@@ -236,11 +249,19 @@ public boolean onPreferenceChange(Preference preference, Object o) {
                     : R.string.always_run_in_background_disabled);
             mSyncOnlyCharging.setEnabled(value);
             mSyncOnlyWifi.setEnabled(value);
+            mSyncOnlyOnSSIDs.setEnabled(mSyncOnlyWifi.isEnabled() && mSyncOnlyWifi.isChecked());
```
Nutomic: You can remove the isEnabled() check, because we uncheck the preference when we disable it.

capi: Those are basically defensive checks against future changes in other parts of code, but yeah, can change that :)

src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java
capi: I used the (cheaper) `HashSet` as default value, and the `TreeSet` is required for ordering of the SSID names in the summary. Can use `TreeSet` for default value as well if you mind the import. Since `TreeSet` is special (it requires comparable entries) I tend to avoid it when not explicitly required.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
capi: Not entirely sure what you mean, if I remove the call to super-class the dialog won't ever be shown. If I just conditionally don't show it and show a Toast instead, I'll need to check for side-effects. Will have a look into that.

src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java
```diff
@@ -297,6 +318,22 @@ public boolean onPreferenceChange(Preference preference, Object o) {
         return true;
     }
 
+    @Nullable
+    private String formatWifiNameList(Set<String> ssids) {
+        if (ssids != null) {
```
Nutomic: I don't think this can actually be null. You should remove the check, because it will only hide errors.

capi: It actually was in my case, because I had all WiFi enabled in the old version where `null` was used. Can remove that.

src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java
```diff
@@ -297,6 +318,22 @@ public boolean onPreferenceChange(Preference preference, Object o) {
...
+            ssids = new TreeSet<>(ssids);
+            StringBuilder sb = new StringBuilder();
+            for (String ssid : ssids) {
+                if (sb.length() > 0) {
+                    sb.append(", ");
+                }
+                sb.append(ssid.replaceFirst("^\"", "").replaceFirst("\"$", ""));
+            }
```
Nutomic: To clarify on my earlier comment: I'd handle this by replacing the set entries in place (or into a new set), and then doing `TextUtils.join(", ", ids);` on the result. Bit easier to understand than the length check for commas imo.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
```diff
@@ -0,0 +1,249 @@
...
+     */
+    private WifiConfiguration[] mConfiguredNetworks;
+    /**
+     * The SSID of the networks that have been enabled (in selective mode)
+     */
+    private Set<String> mSelectedSsids;
+    /**
+     * Adapter between the data above and the ListView
+     */
+    private WifiListAdapter listAdapter;
```
Nutomic: No need to keep a reference to this.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
```diff
@@ -0,0 +1,249 @@
...
+    protected void onBindDialogView(View view) {
+        super.onBindDialogView(view);
+
+        if (isPersistent()) {
+            // currently persisted values are pre-selected in the list
+            String key = getKey();
+            Set<String> ssids = getSharedPreferences().getStringSet(key, null);
+            mSelectedSsids = ssids != null ? new HashSet<>(ssids) : new HashSet<String>();
+        } else {
+            mSelectedSsids = new HashSet<>(); // if non-persistent, nothing is pre-selected
```
Nutomic: No need to handle this case, because we only use it as persistent. And I like to avoid dead code :)

src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java
capi: Since you are the owner of this project I'll follow along. But I get the feeling you are bit nitpicking about everything you did not code yourself.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
```diff
@@ -0,0 +1,249 @@
...
+     */
+    @Override
+    protected void onBindDialogView(View view) {
+        super.onBindDialogView(view);
+
+        if (isPersistent()) {
+            // currently persisted values are pre-selected in the list
+            String key = getKey();
+            Set<String> ssids = getSharedPreferences().getStringSet(key, null);
+            mSelectedSsids = ssids != null ? new HashSet<>(ssids) : new HashSet<String>();
```
Nutomic: You should also pass `new Set<String>()` as default value here, to avoid the check.

And what's the wrapper HashSet good for?

capi: `Set` is an interface. You cannot instantiate it.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    /**
+     * Load the configured WiFi networks, sort them by SSID.
+     *
+     * @return a sorted array of WifiConfiguration, or null, if data cannot be retrieved
+     */
+    private WifiConfiguration[] loadConfiguredNetworksSorted() {
+        WifiManager wifiManager = (WifiManager) getContext().getSystemService(Context.WIFI_SERVICE);
+        if (wifiManager != null) {
```
Nutomic: No need to check for null, that's only returned if the name does not exist (which it does).

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
capi: Look up the JavaDoc of shared preferences, sharing the set is not a good idea and cause lot of problems. Defensive copy when setting / retrieving it should be used.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
```diff
@@ -0,0 +1,249 @@
...
+        OnPreferenceChangeListener listener = getOnPreferenceChangeListener();
+        if (listener != null) {
+            listener.onPreferenceChange(this, ssids);
+        }
+    }
+
+    /**
+     * Adapter between the available and selected Wifis in the preference.
+     */
+    class WifiListAdapter extends BaseAdapter {
```
Nutomic: You should use an ArrayAdapter, and pass the array in instead of accessing it directly. Means you only have to implement getView(), nothing else.

src/main/java/com/nutomic/syncthingandroid/syncthing/DeviceStateHolder.java
```diff
@@ -61,5 +71,26 @@ public boolean isWifiConnected() {
...
+            updateWifiSsid();
+        } else {
+            mWifiSsid = null;
+        }
+    }
+
+    public void updateWifiSsid() {
+        mWifiSsid = null;
+        WifiManager wifiManager = (WifiManager) mContext.getSystemService(Context.WIFI_SERVICE);
+        if (wifiManager != null) {
```
Nutomic: Same here about the null check.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
```diff
@@ -0,0 +1,249 @@
...
+
+        @Override
+        public long getItemId(int position) {
+            return position;
+        }
+
+        @Override
+        public View getView(int position, View convertView, ViewGroup parent) {
+            if (mConfiguredNetworks != null) {
+                WifiConfiguration wifi = mConfiguredNetworks[position];
```
Nutomic: You can just use `getItem(position)`.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
capi: According to JavaDoc it may be `null` if:
> The service or null if the class is not a supported system service. 

There are devices without WiFi but 3G out there. We have one of those at work from a Chinese company.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
```diff
@@ -0,0 +1,249 @@
...
+
+                // SSIDs are formatted "name" if they represent an UTF-8 string,
+                // or as 01234567890ABCDEF if hex, but displaying the '"' doesn't look good,
+                // so strip them
+                String displayName = wifi.SSID;
+                if (displayName != null && displayName.length() > 0 && displayName.charAt(0) == '"') {
+                    // If th SSID starts with '"', it is a string name, otherwise a hex string
+                    if (displayName.length() >= 2) {
+                        // actually, this should always be true, otherwise it's a violation of the
+                        // contract of how SSID names are formatted by WifiManager
```
Nutomic: If this is always true, why even check for it?

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
```diff
@@ -0,0 +1,249 @@
...
+                    }
+                }
+
+                CheckBox wifiView = (CheckBox) convertView.findViewById(R.id.wifiName);
+                wifiView.setTag(wifi.SSID); // the SSID this possibly recylced view is currently representing6
+                wifiView.setText(displayName);
+                wifiView.setChecked(mSelectedSsids.contains(wifi.SSID));
+
+                if (viewCreated) {
+                    wifiView.setOnCheckedChangeListener(new CompoundButton.OnCheckedChangeListener() {
```
Nutomic: It's easier if you set a new listener every time, and just access final variables from the surrounding function. Leave something to do for the GC :D

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
capi: Because you and I obviously have a very different opinion about defensive programming and I have been bitten quite often enough by things that "should not be".

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
```diff
@@ -261,6 +263,34 @@ public void updateState() {
         onApiChange();
     }
 
+    private boolean isAllowedWifiConnected() {
+        boolean wifiConnected = mDeviceStateHolder.isWifiConnected();
+        if (wifiConnected) {
+            SharedPreferences sp = PreferenceManager.getDefaultSharedPreferences(this);
+            Set<String> ssids = sp.getStringSet(PREF_SYNC_ONLY_WIFI_SSIDS, new HashSet<String>());
+            if (ssids == null || ssids.isEmpty()) { // should never be null, but better safe than sorry
```
Nutomic: No need to check for null, it's just extra complexity.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
capi: Avoiding to have to use the Inflater service is one of the primary concerns when trying to get a fluent list in Android. But ok.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
Nutomic: Ah I wasn't aware of that (never used StringSet preferences). And I mean an implementation of Set of course.

src/main/java/com/nutomic/syncthingandroid/fragments/SettingsFragment.java
Nutomic: Ah I didn't think it was on purpose. The performance difference should be really minor, so do it however you want.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
Nutomic: It's not about the inflater. I mean you can set a new listener every time you update a view, so you don't need to go through hoops with view tags etc.

capi: Ah, that makes sense. Yes.

Nutomic:
Oh and don't worry about my weird crash. That must have been caused by something else.

capi:
Updated PR according to your suggestions. `MultiSelectListPreference` works now pretty well, I really overlooked the obvious. Thanks for the pointer.

Please review, @Nutomic

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
```diff
@@ -0,0 +1,107 @@
...
+ * "all networks allowed".
+ *
+ * SSIDs are formatted according to the naming convention of WifiManager, i.e. they have the
+ * surrounding double-quotes (") for UTF-8 names, or they are hex strings (if not quoted).
+ */
+public class WifiSsidPreference extends MultiSelectListPreference {
+
+    public WifiSsidPreference(Context context, AttributeSet attrs) {
+        super(context, attrs);
+        setDialogLayoutResource(R.layout.preference_wifi_ssids);
```
Nutomic: Is there a reason you're using a custom layout? On my Nexus 5, it looks a bit weird because there's so much free space at the bottom. I think you should just use the default layout (same goes for list items).

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    @Override
+    protected void showDialog(Bundle state) {
+        WifiConfiguration[] networks = loadConfiguredNetworksSorted();
+        if (networks != null) {
+            Set<String> selected = getSharedPreferences().getStringSet(getKey(), null);
+            // from JavaDoc: Note that you must not modify the set instance returned by this call.
+            // therefore required to make a defensive copy of the elements
+            selected = selected != null ? new HashSet<>(selected) : new HashSet<String>();
```
Nutomic: How about putting `new HashSet<>(getSharedPreferences().getStringSet(getKey(), new HashSet<String>());`?

And no need to quote the docs, I believe you now :D

Nutomic:
Sorry for the late review, I've had problems with my internet at home -.-

The code looks good, except for those few minor things. Also, it would be good if you could add a bit of Javadoc to `extractSsid()` and `showDialog()` in WifiSsidPreference.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
capi: Actually that was an overlooked piece from this being a custom preference. Will be fixed.

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
capi: This saves the unnecessary allocation of an empty HashSet when the key is set, but the price for GC of this instance is cheap, so I'll remove it.

capi:
Updated PR, fixed the points mentioned above. Looking forward to your review again :)

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
```diff
@@ -0,0 +1,129 @@
...
+     */
+    private CharSequence[] extractSsid(WifiConfiguration[] configs, boolean stripQuotes) {
+        CharSequence[] result = new CharSequence[configs.length];
+        for (int i = 0; i < configs.length; i++) {
+            // WiFi SSIDs can either be UTF-8 (encapsulated in '"') or hex-strings
+            if (stripQuotes)
+                result[i] = configs[i].SSID.replaceFirst("^\"", "").replaceFirst("\"$", "");
+            else
+                result[i] = configs[i].SSID;
+        } // for
```
Nutomic: Looks like that slipped in again :p

Nutomic:
Looks good!

I just noticed that I'd personally use lists over arrays, and that Javadoc should start with a single sentence short description, and everything else in a new paragraph.

But these don't really matter, just remove that `// if` and I'll merge ;)

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
capi: It's so natural to my team's style, that it took me (more than) a few seconds to realize what you were referring to :)

capi:
PR updated with the removed comment.

Nutomic:
Great, thank you :)

src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
Nutomic: I find it pretty weird actually. That's what we have syntax highlighting for :o

PR #503, opened 2015-11-17
Add warning that exported private key may be stored in an unsafe location

capi:
The current export functionality stores the unencrypted private Syncthing private key in the backup folder, which is on the shared storage (SD-card or emulated SD card) and can therefore be read by any application on the device and/or when connecting the mobile phone to a PC.

Therefore the user should at least be warned that this might compromise the security of the key.

This pull request adds a warning to the export confirmation dialog in the English translation. Other translations are needed.


Nutomic:
Thanks :)

PR #520, opened 2015-12-17
Update folder_id_required string (de)

xadh00m:
The folder id should not be empty. This is the only issue checked now.


canton7:
Please read https://github.com/syncthing/syncthing-android#translations

AudriusButkevicius:
The project uses transifex for translations.

xadh00m:
Ah sorry. I missed that part. Seems to be fixed already.

PR #525, opened 2015-12-21
update path

mauk81:
update readme shreenshot path


Nutomic:
Oh I forgot about that. Thanks for fixing!

PR #546, opened 2016-01-23
Fix NullPointerException in onNewIntent()

capi:
An implementation for `onNewIntent` has been added in c28065786, but doesn't take into consideration that the notification will also send an intent, but without an action. So need to add a `null` check.


Nutomic:
Whoops, thank you :)

Issue #515, opened 2015-12-12
Syncthing binary crashed with error code 11 after update to 0.7.5


pascalfree:
Hi

Since I updated to Syncthing for Android 0.7.5 the app is stuck on "loading".
This is on an HTC One V with Android 4.0.3

This is the log, when I force close syncthing and it restarts itself:

    2015-12-12 03:32:08,INFORMATION,ActivityManager,Android System,Displayed com.nutomic.syncthingandroid/.activities.MainActivity: +1s357ms
    2015-12-12 03:32:19,INFORMATION,dalvikvm,Syncthing,Could not find method android.net.ConnectivityManager.isActiveNetworkMetered, referenced from method com.nutomic.syncthingandroid.syncthing.DeviceStateHolder.<init>
    2015-12-12 03:32:19,INFORMATION,SyncthingService,Syncthing,Web GUI will be available at https://127.0.0.1:8384
    2015-12-12 03:32:19,INFORMATION,SyncthingService,Syncthing,Starting syncthing according to current state and preferences
    2015-12-12 03:32:20,INFORMATION,DEBUG,/system/bin/debuggerd,pid: 27376, tid: 27376  >>> /data/data/com.nutomic.syncthingandroid/lib/libsyncthing.so <<<

    2015-12-12 03:32:20,INFORMATION,DEBUG,/system/bin/debuggerd,40040000-408d7000 /data/data/com.nutomic.syncthingandroid/lib/libsyncthing.so

    2015-12-12 03:32:20,INFORMATION,System,Syncthing,libcore.io.ErrnoException: kill failed: ESRCH (No such process)
	at libcore.io.Posix.kill(Native Method)
	at libcore.io.ForwardingOs.kill(ForwardingOs.java:77)
	at java.lang.ProcessManager$ProcessImpl.destroy(ProcessManager.java:257)
	at com.nutomic.syncthingandroid.syncthing.SyncthingRunnable.run(SyncthingRunnable.java:136)
	at java.lang.Thread.run(Thread.java:864)

    2015-12-12 03:32:20,ERROR,SyncthingNativeCode,Syncthing,Syncthing binary crashed with error code 11
    2015-12-12 03:32:20,WARNING,SyncthingNativeCode,Syncthing,sh: <stdin>[1]: grep: not found
    2015-12-12 03:32:21,WARNING,SyncthingNativeCode,Syncthing,usage: ionice <pid> [none|rt|be|idle] [prio]
[...]

Issue #498, opened 2015-11-15
Android device disconnects, Web GUI unavailable


janusgrave:
I've been using Syncthing heavily and keep running into an issue where the Android client disconnects from the cluster. This occurs when I have "Always run in background" enabled.

In the Syncthing Android app, all information in the left panel (Device ID, CPU usage, etc) is now blank. When I try to open the Web GUI, a "Webpage not available" screen appears. When I try to view logs through the options menu, the app just displays "Retrieving logs..." and never loads.

So it seems Syncthing android's app GUI is still running but Syncthing itself has crashed. Restarting the Android device corrects the problem, but it eventually happens again when "Always run in background" is enabled.

Running Syncthing 0.12.1 / 0.7.0 on a device running Android 4.4.2

Other devices running 0.12.2 (Linux 64) and 0.12.2 (Linux ARM)

I'm attaching a log taken just after restarting the device and launching syncthing android
[log-syncthing.txt](https://github.com/syncthing/syncthing-android/files/34972/log-syncthing.txt)

Nutomic:
Can you post a log right after Syncthing stops working? I don't think logs are kept during a restart.

lawfulintercept:
I've been unable to reproduce the Web GUI unavailable part of the issue, but I just upgraded syncthing-android to v0.7.4 (Syncthing v0.12.2+18-g56f1c29) and now the Android device isn't connecting at all when it's not on the same network as the other nodes.

I've attached an android log - it should show my android device disconnected after the update, then re-connecting once I switch to to the same Wi-Fi network as the other devices
[syncthing-android.txt](https://github.com/syncthing/syncthing-android/files/35828/syncthing-android.txt)

capi:
The default port-mapping for 0.12 syncthing-client changed, you are sure that your router/firewall is not blocking the 0.12 ports?

lawfulintercept:
It's been working just fine ever since Syncthing 0.12 - so that wouldn't be the case unless it changed with the most recent update that went out today. In either case, my network configuration hasn't changed in that time, and I've been using UPnP and global discovery (which should obviate the need for port-forwarding, no?)

frabcus:
For what it's worth, I get the same symptoms as this. I don't think it is to do with the networking - when I trick Syncthing into restarting (e.g. by changing a setting back and forth) it then works fine.

It is like the background daemon can die, but without giving any clear error message.

NickPyz:
+1 similar weirdness here

[...]

PR #547, opened 2016-01-25
Experimental setting to keep a partial wakelock

capi:
Issues like #498 and #515 might be caused by the CPU going to sleep if not connected to a power-supply. In Android, for the CPU to remain active after screen has turned off, it is necessary to keep a [partial wake-lock][1]. 

Current discussion in the issues mentioned above indicates, that the native binary is not aware of this and does not hold any wake-locks and might crash if there is a time-jump forward (which will be the case if CPU goes to sleep for some undefined time). Also for other devices this will be seen as a disconnect, even if the binary does not crash.

This PR adds an "Experimental" section to the settings that allows to enable holding a partial wake-lock while the binary is running.

This PR does not yet address Android 6 usability for handling [App Standby and Doze][2] mode.

[1]: http://developer.android.com/reference/android/os/PowerManager.html#PARTIAL_WAKE_LOCK
[2]: http://developer.android.com/training/monitoring-device-state/doze-standby.html#support_for_other_use_cases

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
@@ -4,6 +4,7 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.os.Environment;
+import android.os.PowerManager;
 import android.preference.PreferenceManager;
 import android.text.TextUtils;
 import android.util.Log;
@@ -95,7 +96,14 @@ public class SyncthingRunnable implements Runnable {
         }
         // Loop Syncthing
         Process process = null;
+        // Potential fix for #498, keep the CPU running while native binary is running
+        PowerManager pm = (PowerManager) mContext.getSystemService(Context.POWER_SERVICE);
+        PowerManager.WakeLock wakeLock = useWakeLock()
+                ? pm.newWakeLock(PowerManager.PARTIAL_WAKE_LOCK, TAG)
+                : null;
         try {
+            if (wakeLock != null)
+                wakeLock.acquire();
             // Loop to handle Syncthing restarts (these always have an error code of 3).
             do {
                 ProcessBuilder pb = (useRoot())
@@ -132,6 +140,8 @@ public class SyncthingRunnable implements Runnable {
         } catch (IOException | InterruptedException e) {
             Log.e(TAG, "Failed to execute syncthing binary or read output", e);
         } finally {
+            if (wakeLock != null)
+                wakeLock.release();
             if (process != null)
                 process.destroy();
             if (ret != 0) {
@@ -147,6 +157,14 @@ public class SyncthingRunnable implements Runnable {
     private boolean useRoot() {
         SharedPreferences sp = PreferenceManager.getDefaultSharedPreferences(mContext);
         return sp.getBoolean(SyncthingService.PREF_USE_ROOT, false) && Shell.SU.available();
+    }
+
+    /**
+     * Returns true if the experimental setting for using wake locks has been enabled in settings.
+     */
+    private boolean useWakeLock() {
+        SharedPreferences sp = PreferenceManager.getDefaultSharedPreferences(mContext);
+        return sp.getBoolean(SyncthingService.PREF_USE_WAKE_LOCK, false);
     }
 
     /**
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -296,6 +296,12 @@ Please report any problems you encounter via Github.</string>
 
     <string name="export_config">Export Configuration</string>
 
+    <string name="experimental_settings">Experimental</string>
+
+    <string name="keep_wakelock_while_binary_running">Keep the CPU awake while Syncthing is running</string>
+
+    <string name="keep_wakelock_while_binary_running_summary">Use this setting if you experience unexpected disconnects while operating on battery. This will result in increased battery consumption.</string>
+
     <!-- Toast shown after config was successfully exported -->
     <string name="config_export_successful">Config was exported to %1$s</string>
 
--- a/src/main/res/xml/app_settings.xml
+++ b/src/main/res/xml/app_settings.xml
@@ -147,6 +147,18 @@
             android:title="@string/streset_title"
             android:singleLine="true" />
 
+        <PreferenceScreen
+            android:key="syncthing_experimental"
+            android:title="@string/experimental_settings">
+
+            <CheckBoxPreference
+                android:key="wakelock_while_binary_running"
+                android:title="@string/keep_wakelock_while_binary_running"
+                android:summary="@string/keep_wakelock_while_binary_running_summary"
+                android:defaultValue="false" />
+
+        </PreferenceScreen>
+
     </PreferenceCategory>
 
     <PreferenceCategory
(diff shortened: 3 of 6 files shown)


Nutomic:
@lawfulintercept, @frabcus, @NickPyz, @C0R10L4NU2, @pascalfree and anyone else affected by this issue, can you test the following apk please? It will install over your existing installation. And please report how it affects CPU usage for you.
https://dl.dropboxusercontent.com/u/22457329/syncthing-pr547-armeabi.apk
https://dl.dropboxusercontent.com/u/22457329/syncthing-pr547-x86.apk

@capi As I wrote on IRC, I'd prefer to have this enabled for everyone, rather than tell people to enable some hidden setting. So let's test this way first and see how it goes :)

@calmh We didn't change anything in this regard in the app, and these issues only came up recently. Is that something on your side? I'd really prefer to fix that in Syncthing, rather than introduce additional battery drain.

capi:
Especially users on Android 6.0 will only benefit if they also enable the exception in *Battery Optimization*, otherwise the app won't be able to hold on to the wake-lock.

At the moment, I do think holding a wake-lock while the service is running is the only way for people who want sync while the device is on battery and screen not turned on.

While I never saw any binary crashes during my testing, I did notice that on battery the connection to other devices is often lost after a few minutes, which also correlates to the "awake" column in the "Battery history details" of my device.

The idea for hiding it behind the "Experimental" page was that we could roll that out and ask people who report issues like the ones referenced above to enable it, until we know how it affects overall usage. But I'm very fine to test this out and roll it as a main release once we know if it really makes things better.

quantum1423:
Random guy barging in here:

This behavior is actually quite strange. Android is supposed to wake up apps that are blocking on network operations when the radio gets traffic; generally a properly-written app shouldn't need to hold wakelocks for network access at all.

capi:
@quantum1423 true, the issue here is, that Syncthing is a native ARM binary written in Go that's actually not aware of the requirements of Android. The binary does not know anything about Alarms, Wakeloks, WifiLocks, etc. So the problem is, that it's timer's are actually sleeps on a thread, which won't wake up a sleeping CPU once they expire, etc. The hosting Android app itself can only work around this issue.

The only "real" in my opinion, solution in the long term is a "real" Android application that speaks BEP (Block Exchange Protocol) with "desktop-syncthing". But until then, the current solution provides a lot of use to a lot of people (including me), but it requires some non-Android-ish workarounds like this one.

quantum1423:
I see. And the binary doesn't like being randomly put to sleep? Does this mean that suspending syncthing even on non Android can cause problems? That seems like it should be a bug, since laptops can also be suspended. My comment was that IIRC you don't need to do anything android specific for network activity to wake up the whole system.

wweich:
I have been using this test build for 12 days now, after noticing that Syncthing on my phone would disconnect from my server after lying on the table for a short amount of time. I already had Syncthing on the exception list for battery optimization, but that alone did not make a difference.

With this test build, I haven't noticed any disconnects related to standby.
According to the battery usage screen, Syncthing is now the most battery hungry app on my phone. Sometimes even more than the screen ;).
But the overall battery performance did not really suffer a noticeable amount. After 10 hours without power cord, I only noticed a maximum difference of a few percent more battery drain.

I am using a Moto X Style with stock Marshmallow and data on an 128GB SD Card.

Apart from Syncthing, I also have my e-mail app (K9 Mail) and automation app (Automagic) in the "Not optimized" list.

Nutomic:
Okay so I'm pretty indifferent about this. I can't really judge how many people are affected by these issues, and if this PR is actually a proper workaround. And I'm not sure if affected people will even find this option. Anyway, this would only be a temporary workaround until these bugs are properly fixed upstream.

So what do you guys think, should we merge this?

capi:
**CON:**
- Its a workaround hack, that will drain battery of people who use syncthing while on battery. It does *not* fix the crash issues of the binary. It *won't* affect people that use it with a "on AC only" setting.
- People will most likely not find it by themselves unless directed by someone here in an issue or on the forum. That could also be listed as a PRO, though.
- Manual steps are required on Android 6.0 to actually make this work (it needs to be manually added to the Doze exception list). This could be improved in a later version by asking to be added to the list via API when the setting is enabled (I don't have an Android 6 device at the moment, that's why it's not in yet).

**PRO:**
- It fixes power-saving-enabled disconnect issues of people that use it while on battery. Disconnects are bad, especially on mobile data connections, since index can be quite large and use up data allowance pretty fast if on most screen-on/off the index needs to be exchanged. The index on my phone is >4MB which is exchanged on every reconnect.
- Since it's an experimental setting, disabled by default, it won't ruin anything for existing users. So it's "cheap" to merge this (in my opinion).
- I think the "experimental" section is a good start, I'd also move the "root" setting there in a next pull-request :-)

Nutomic:
I agree with your reasoning. However, if people can't find this PR to comment, I doubt they would find the setting, and know what to do with it (as you mentioned). That's why I was asking for comments by affected people.

And I definitely agree that the root setting should go to an experimental setting.

delexi:
FWIW, I would test this out, if it came as a hidden option with the next release.

Nutomic:
Okay I'm gonna merge this. But first, can you carify the description of the setting a bit? It should describe when exactly this should be used, and that it will increase battery usage. I wouldn't mention the Syncthing binary, because no one cares about that.

Also, you can require a restart in SettingsFragment, so no need to put it in the description (you can check how the root setting does it).

capi:
@Nutomic Sure, will do. I expect sometime next week I'll find time to work on this again.

capi:
@Nutomic I have updated the description and use the API to request a restart. Please re-evaluate if you can merge this now.

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
```diff
@@ -95,7 +96,12 @@ public void run() {
         }
         // Loop Syncthing
         Process process = null;
+        // Potential fix for #498, keep the CPU running while native binary is running
+        PowerManager pm = (PowerManager) mContext.getSystemService(Context.POWER_SERVICE);
+        PowerManager.WakeLock wakeLock = useWakeLock() ? pm.newWakeLock(PowerManager.PARTIAL_WAKE_LOCK, TAG) : null;
```
Nutomic: Can you put the `?` and the `:` from the ternary operator on seperate lines? Makes it a bit easier to read.

src/main/res/values/strings.xml
```diff
@@ -296,6 +296,12 @@ Please report any problems you encounter via Github.</string>
 
     <string name="export_config">Export Configuration</string>
 
+    <string name="experimental_settings">Experimental</string>
+
+    <string name="keep_wakelock_while_binary_running">Keep Wakelock</string>
+
+    <string name="keep_wakelock_while_binary_running_summary">Keep the CPU awake while Syncthing is running. Use this setting if you experience unexpected disconnects while operating on battery. This will result in increased battery consumption.</string>
```
Nutomic: I would just use "Keep the CPU awake while Syncthing is running" as the title. Means the description isn't so long, and most people don't know what a wakelock is anyway.

src/main/res/xml/app_settings.xml
```diff
@@ -147,6 +147,19 @@
...
+        <PreferenceScreen
+            android:key="syncthing_experimental"
+            android:title="@string/experimental_settings">
+
+            <CheckBoxPreference
+                android:key="wakelock_while_binary_running"
+                android:title="@string/keep_wakelock_while_binary_running"
+                android:summary="@string/keep_wakelock_while_binary_running_summary"
+                android:defaultValue="false"
+                />
```
Nutomic: I know this is really nitpicking, but the `/>` belongs on the same line as the last line of xml :p

Nutomic:
Just some really minor nitpicks, looks good to merge otherwise!

capi:
Adjusted as per your suggestions.

Nutomic:
Great, thanks :)

PR #560, opened 2016-02-09
Don't restart before config has been updated (fixes #398).

### LINKED ISSUE #398: Use asynchrounous sequences when updating config
On line https://github.com/syncthing/syncthing-android/blob/2ed6dac071fbb5757653054724906370e557edfc/src/main/java/com/nutomic/syncthingandroid/syncthing/RestApi.java#L359 there is a race condition because the PostTask might not be completed before restarting Syncthing.
To solve this, an asynchronous sequence (PostTask.Then(Restart)) is required, but that might be an issue for callers which expect blocking behavior to make sure Synchting is halted before continuing.

Nutomic:
@Zillode Is this what you had in mind with #398? I don't really understand what you mean with your last sentence ("blocking behaviour"), so I didn't merge it myself.

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/RestApi.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/RestApi.java
@@ -352,13 +352,20 @@ public class RestApi implements SyncthingService.OnWebGuiAvailableListener,
      * @param activity The calling activity.
      * @param updateConfig If true, {@link #mConfig} will be sent to `/rest/system/config`.
      */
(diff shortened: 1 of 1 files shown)


Zillode:
Partially, the blocking(!) restart call needs to [update the GUI](https://github.com/syncthing/syncthing-android/blob/fix-398/src/main/java/com/nutomic/syncthingandroid/syncthing/RestApi.java#L372).

Nutomic:
Do you mean when `updateConfig` is true? `restart()` is also called in that case.

Zillode:
Yes, but there is a (potentially large) delay before `restart()`

PR #584, opened 2016-03-14
General code quality fix-1

faisal-hameed:
This pull request is focused on resolving occurrences of Sonar rules 
squid:S2864 - "entrySet()" should be iterated when both the key and value are needed.
You can find more information about the issue here: 
https://dev.eclipse.org/sonar/rules/show/squid:S2864
 
Please let me know if you have any questions.

Faisal Hameed

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/RestApi.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/RestApi.java
@@ -568,9 +568,10 @@ public class RestApi implements SyncthingService.OnWebGuiAvailableListener,
                         }
                     }
                     Map<String, Connection> connections = new HashMap<>();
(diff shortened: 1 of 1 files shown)


Nutomic:
The changes to map iteration definitely make sense. But what's the point of the private constructors? No one's gonna instantiate those classes anyway, so it just adds clutter. And "some style guide says it" isn't a good enough reason for me.

faisal-hameed:
@Nutomic Excluded private constructors.

Nutomic:
You didn't update the commit message though :p 

And imo you should put an actual text in the summary, because no one just knows what these squid IDs represent.

faisal-hameed:
@Nutomic Updated commit message.

Nutomic:
Thanks :)

PR #590, opened 2016-03-20
Nocgo

Zillode:


Nutomic:
Thanks!

- Did you also check that DNS works, just to be sure?
- I don't think we should rename the build files to *-nocgo. 
- Do you have a link to the GOOS=android build issue? That should go into a comment there. 
- And we already have Syncthing v0.12.20, no need to change that :p

Zillode:
- I don't have a link to the upstream issue. It just doesn't build.
- Removed Syncthing version commit.
- I tested it but we should test the version built by jenkins as well.

Nutomic:
@st-jenkins build this please

Let's see if this works.

Issue #460, opened 2015-09-22
[Feature Request] Periodic Sync


tinloaf:
Hi,

I think it would be great if Syncthing on Android would not just have the two "start me manually" and "always run" modes, but could also be configured to start every couple of hours, look for changes, sync if there are any, and then shut down again. This would provide a nice trade-off between data/battery consumption and "oh damn, my files are out of sync since I forgot to start Syncthing again".

Obviously, some sane assumptions must be made: If a device cannot be connected to within timeframe X, assume it to have no changes and shut down, etc pp

Nutomic:
This can be done with broadcast intents now, see the [instructions here](https://github.com/syncthing/syncthing-android/wiki/Remote-Control-by-Broadcast-Intents). I'm closing this as out of scope.

PR #594, opened 2016-03-25
Add Broadcast-Receiver to remotely control the app (Addresses also #460)

sqrt-1764:
In order to not have to reinvent the wheel, SyncThing should be controllable by some intents.
I also need an option to save battery and need SyncThing running only at certain places / times.

We have tools like Tasker, Llama or Automate to customize our device. So the user-logic can be implemented by the users if SyncThing is remotely configurable.

A implemented a quick and simple solution:
Added a IntentService to receive Broadcast-Intents to remotely control / configure the app.

As a first step only the "quit"-command is implemented in this service.
It simply stops the SyncThingService.

TODO (?): Also kill a bound activity if any. At the moment, if the activity is active (bound to the service), the service is stopped only after the activity is left.
Before putting work into this I want to make shure that my solution is a step into the right direction. ;-)

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -91,6 +91,11 @@
                 <action android:name="android.intent.action.BOOT_COMPLETED" />
             </intent-filter>
         </receiver>
+        <receiver android:name=".syncthing.AppConfigReceiver">
+            <intent-filter>
+                <action android:name="com.nutomic.syncthingandroid.APP_CONFIGURE" />
+            </intent-filter>
+        </receiver>
     </application>
 
 </manifest>
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
@@ -1,0 +1,29 @@
+package com.nutomic.syncthingandroid.syncthing;
+
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.os.Bundle;
+
+/**
+ *
+ * Created by sqrt-1764 on 25.03.16.
+ */
+public class AppConfigReceiver extends BroadcastReceiver {
+    public static final String ACTION_QUIT_SYNCTHING = "quit";
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        final Bundle extras = intent.getExtras();
+
+        if (extras == null) return;
+
+        for (String key : extras.keySet()) {
+            switch (key) {
+                case ACTION_QUIT_SYNCTHING:
+                    context.stopService(new Intent(context, SyncthingService.class));
+                    break;
+            }
+        }
+    }
+}


Nutomic:
This feature is definitely good to have, but you picked a rather complicated call to start with. Shutting down Syncthing has some corner cases:

- Shutdown is not supported if "always run in background" is enabled (in that case, Syncthing would just be restarted whenever BatteryReceiver or NetworkReceiver is triggered). If this setting is enabled, your receiver should probably log a warning and do nothing.

- What happens if the user has Syncthing opened when this intent is received? Closing Syncthing and keeping the activity open is bad, because the UI won't work any more. Closing the activity seems to be really bad UX. Maybe we could delay the shutdown until the user switches to a different app?

Maybe there are also other cases that I can't think of right now. To be clear, this isn't about the code yet, as we need to figure out the expected behaviour first.

sqrt-1764:
This is the reason why I stepped forward so early whithout much code. There are clearly some points that should be discussed before implementing this in the productive version.

The option to shutdown SyncThing remotely clearly collides with the "always run in background" option. So it has to be decided who should win in this case.

- Either the intent wins - you wanted it to shut down, then you get it. Leaving the behaviour that the service could be restartet automatically when the events mentioned are triggerd. In this case I also consider the possible shutdown of the UI acceptable (call the option "SHUTDOWN_SYNCTHING_FORCEFULLY") - this was my initial strategy.

- Or SyncThing wins - the intent is to be ingored when "always run in background" is active. This would be similar to disabling the exit-option in SynThing in this case. Then maybe a notification that the intent was discarded would be nice (which may cointain an intent to navigate to the Settings-Activity). Maybe the more user-friendly version.
("always run in background" may be made configurable by this intent also ;-) )


Regarding the delay:
Currently the service is only stopped if the intent was received (and there the stopService in the AppConfigReceuver was called). But if an activty was active, the service runs until the last object that bound to the service is closed. In other words, the service keeps running as long as the activity is running.

By the way: Wouldn't it be better to move the bindService command (in SyncThingActivity) from onCreate() to onResume()? (And the unbindService accordingly to onPause()) Then the connection to the service is only held active when the acitivy is in foreground and not as long as it is in Memory.
The startService in onCreate would retain the current startup-behaviour ...

Nutomic:
I think a shutdown with the current implementation would be pretty pointless, as you don't have any guarantees, Syncthing might just be started a second later and you won't even know it. I think it's better to ignore the intent in that case. As another option, we could add support for shutting down Syncthing while "run in background" is enabled. But that might be too confusing for users.

Ah right, I didn't remember that the activity keeps the service alive. That sounds good to me then.

Moving the bind/unbind to onResume/onPause sounds fine to me. We'll just have to test so it doesn't break anything.

sqrt-1764:
I now also think that ignoring the intent if background is enabled is the better option. I give this a try - doesn't seem to be too complicated.

After that an option to control running in background could be added to the BroadCast-Receiver.

Nutomic:
I don't see how exiting with "always run in background" could be done as an option on the intent. Instead, this could be done by adding an extra variable in syncthing, something like `isStopped`, and ignoring wifi/charging changes until that variable is reset. But like I said, that might be confusing for users.

sqrt-1764:
Just finished the first part ;-)
I will test this a little bit before i push it to github but it looks promising. (Or if you can't wait ... ;-) )

Regarding a unit-test:
I am not shure how to write a unit-test that verifies the correct behaviour. What has to be tested?
I can see the service starting and stopping. But I am not aware of any side effects that might be important. At the moment I see no other problems ...

For the option of configuring "always run in background" :
I haven't taken a detailled look, but all that should be done here should be the same as if the user changes that option on the settings-page.
But maybe this option is not as interesting as I first thought. Maybe the manual setting is sufficient.

Nutomic:
You could probably write a tests that checks that `Context#stopService` is called with the correct parameters after the intent is received.

I don't mean changing that option, i mean we should add `isStopped` as an additional, internal option (probably in DeviceStateHolder). For this PR, it should be fine to give some kind of warning/error if "always run in background" is enabled.

Nutomic:
Why did you close this? :o

sqrt-1764:
Ah - I only renamed the branch. Didn't know that this closes the PR. Will reopen it on the new branch

Nutomic:
Ah renaming the branch isn't supported i guess. And there's usually no reason to do that.

sqrt-1764:
I just committed my "final" version, if you want to take a look. Now I have to create the tests ...

The AppConfigReceiver-Service can be easily extended to support more commands.
Currently I implemented the action "sycthingservice" with the commands "start" and "stop".


In short words what I did:

"start" just (re)starts the Syncthingservice via context.startService().
"stop" stops the Syncthingservice by a call to startService with the new action "shutdown". This because a stopService command when the Syncthingservice wasn't active resulted in a runtime-error of the syncthing-binary.
Like this, the service can handle the shutdown internally and delay the stopService call if needed.

For the delay, I found the mStopScheduled member-variable which seems to be there for exact this job.
If mCurrentStatus is not INIT or STARTING, I stop the service immediately.
Otherwise I set mStopScheduled to true - which then does the rest.

PR #598, opened 2016-03-26
Update README.md

lkwg82:
fixed broken screenshots

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,11 +5,11 @@
 
 A wrapper of [Syncthing](https://github.com/syncthing/syncthing) for Android.
 
-<img src="src/fat/play/en-UK/listing/phoneScreenshots/screenshot_phone_1.png" alt="screenshot 1" width="200" /> 
-<img src="src/fat/play/en-UK/listing/phoneScreenshots/screenshot_phone_2.png" alt="screenshot 2" width="200" /> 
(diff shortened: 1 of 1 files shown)


wweich:
Why use a Wikipedia image for play store, instead of th official one?
https://play.google.com/intl/en_us/badges/images/apps/en-play-badge.png

Nutomic:
Damn, I never notice when I break the readme images. Thanks for fixing it. And I agree, using the official Google Play image seems better.

lkwg82:
1. I did not know the official path. 
2. Svg is smooth with resizing.
Am 27.03.2016 10:05 vorm. schrieb Wulf Weich <notifications@github.com>:Why use a Wikipedia image for play store, instead of th official one?
https://play.google.com/intl/en_us/badges/images/apps/en-play-badge.png

Nutomic:
Yeah please use the official image. Also, I don't see why you changed the link to HTML, markdown is totally fine.

lkwg82:
html was used to explicitly set the size

Nutomic:
Ah I didn't see that. Can you squash the commits though?

lkwg82:
Of course. 

Btw. can we integrate st-bot to merge PRs easily?

Nutomic:
Thanks. I haven't looked into the not yet, because we don't have that many PRs here. But maybe I'll get around to it.

PR #599, opened 2016-03-27
Appconfig receiver

sqrt-1764:
Reopened Pull-Request after renaming my branch.

### DIFF
--- a/src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/AppConfigReceiverTest.java
+++ b/src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/AppConfigReceiverTest.java
@@ -1,0 +1,108 @@
+package com.nutomic.syncthingandroid.test.syncthing;
+
+import android.content.Intent;
+import android.preference.PreferenceManager;
+import android.test.AndroidTestCase;
+
+import com.nutomic.syncthingandroid.syncthing.AppConfigReceiver;
+import com.nutomic.syncthingandroid.syncthing.SyncthingService;
+import com.nutomic.syncthingandroid.test.MockContext;
+
+/**
+ * Test the correct behaviour of the AppConfigReceiver
+ *
+ * Created by sqrt-1674 on 27.03.16.
+ */
+public class AppConfigReceiverTest extends AndroidTestCase {
+    private AppConfigReceiver mReceiver;
+    private MockContext mContext;
+
+    @Override
+    protected void setUp() throws Exception {
+        super.setUp();
+        mReceiver = new AppConfigReceiver();
+        mContext = new MockContext(getContext());
+    }
+
+    @Override
+    protected void tearDown() throws Exception {
+        PreferenceManager.getDefaultSharedPreferences(mContext).edit().clear().commit();
+        super.tearDown();
+    }
+
+
+    /**
+     * Test starting the Syncthing-Service if "always run in background" is enabled
+     * In this case starting the service is allowed
+     */
+    public void testStartSyncthingServiceBackground() {
+        PreferenceManager.getDefaultSharedPreferences(mContext)
+                .edit()
+                .putBoolean(SyncthingService.PREF_ALWAYS_RUN_IN_BACKGROUND, true)
+                .commit();
+
+        Intent intent = new Intent(new Intent(mContext, AppConfigReceiver.class));
+        intent.setAction(AppConfigReceiver.ACTION_START);
+
+        mReceiver.onReceive(mContext, intent);
+        assertEquals("Start SyncthingService Background", 1, mContext.getReceivedIntents().size());
+        assertEquals("Start SyncthingService Background", SyncthingService.class.getName(),
+                        mContext.getReceivedIntents().get(0).getComponent().getClassName());
+    }
+
+    /**
+     * Test stopping the service if "alway run in background" is enabled.
+     * Stopping the service in this mode is not allowed, so no stopService-intent may be issued.
+     */
+    public void testStopSyncthingServiceBackground() {
+        PreferenceManager.getDefaultSharedPreferences(mContext)
+                .edit()
+                .putBoolean(SyncthingService.PREF_ALWAYS_RUN_IN_BACKGROUND, true)
+                .commit();
+
+        Intent intent = new Intent(new Intent(mContext, AppConfigReceiver.class));
+        intent.setAction(AppConfigReceiver.ACTION_STOP);
+
+        mReceiver.onReceive(mContext, intent);
+        assertEquals("Stop SyncthingService Background", 0, mContext.getStopServiceIntents().size());
+    }
+
+    /**
+     * Test starting the Syncthing-Service if "always run in background" is disabled
+     * In this case starting the service is allowed
+     */
+    public void testStartSyncthingServiceForeground() {
+        PreferenceManager.getDefaultSharedPreferences(mContext)
+                .edit()
+                .putBoolean(SyncthingService.PREF_ALWAYS_RUN_IN_BACKGROUND, false)
+                .commit();
+
+        Intent intent = new Intent(new Intent(mContext, AppConfigReceiver.class));
+        intent.setAction(AppConfigReceiver.ACTION_START);
+
+        mReceiver.onReceive(mContext, intent);
+        assertEquals("Start SyncthingService Foreround", 1, mContext.getReceivedIntents().size());
+        assertEquals("Start SyncthingService Foreround", SyncthingService.class.getName(),
+                            mContext.getReceivedIntents().get(0).getComponent().getClassName());
+    }
+
+    /**
+     * Test stopping the Syncthing-Service if "always run in background" is disabled
+     * In this case stopping the service is allowed
+     */
+    public void testStopSyncthingServiceForeground() {
+        PreferenceManager.getDefaultSharedPreferences(mContext)
+                .edit()
+                .putBoolean(SyncthingService.PREF_ALWAYS_RUN_IN_BACKGROUND, false)
+                .commit();
+
+        Intent intent = new Intent(new Intent(mContext, AppConfigReceiver.class));
+        intent.setAction(AppConfigReceiver.ACTION_STOP);
+
+        mReceiver.onReceive(mContext, intent);
+        assertEquals("Stop SyncthingService Foreround", 1, mContext.getStopServiceIntents().size());
+        Intent receivedIntent = mContext.getStopServiceIntents().get(0);
+        assertEquals("Stop SyncthingService Foreround", SyncthingService.class.getName(),
+                                                receivedIntent.getComponent().getClassName());
+    }
+}
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
@@ -1,0 +1,75 @@
+package com.nutomic.syncthingandroid.syncthing;
+
+import android.app.Notification;
+import android.app.NotificationManager;
+import android.app.PendingIntent;
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.os.Build;
+import android.support.v4.app.NotificationCompat;
+
+import com.nutomic.syncthingandroid.R;
+import com.nutomic.syncthingandroid.activities.MainActivity;
+
+/**
+ * Broadcast-receiver to control and configure SyncThing remotely
+ *
+ * Created by sqrt-1764 on 25.03.16.
+ */
+public class AppConfigReceiver extends BroadcastReceiver {
+    private static final int ID_NOTIFICATION_BACKGROUND_ACTIVE = 3;
+
+    /**
+     * Start the Syncthing-Service
+     */
+    public static final String ACTION_START = "com.nutomic.syncthingandroid.action.START";
+
+    /**
+     * Stop the Syncthing-Service
+     * If alwaysRunInBackground is enabled the service must not be stopped. Instead a
+     * notification is presented to the user.
+     */
+    public static final String ACTION_STOP  = "com.nutomic.syncthingandroid.action.STOP";
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        switch (intent.getAction()) {
+            case ACTION_START:
+                context.startService(new Intent(context, SyncthingService.class));
+                break;
+
+            case ACTION_STOP:
+                if (SyncthingService.alwaysRunInBackground(context)) {
+                    final String msg = context.getString(R.string.appconfig_receiver_background_enabled);
+
+                    Context appContext = context.getApplicationContext();
+
+                    NotificationCompat.Builder nb = new NotificationCompat.Builder(context)
+                            .setContentText(msg)
+                            .setTicker(msg)
+                            .setStyle(new NotificationCompat.BigTextStyle().bigText(msg))
+                            .setContentTitle(context.getText(context.getApplicationInfo().labelRes))
+                            .setSmallIcon(R.drawable.ic_stat_notify)
+                            .setAutoCancel(true)
+                            .setContentIntent(PendingIntent.getActivity(appContext,
+                                                        0,
+                                                        new Intent(appContext, MainActivity.class),
+                                                        PendingIntent.FLAG_UPDATE_CURRENT));
+
+                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+                        nb.setCategory(Notification.CATEGORY_ERROR);        // Only supported in API 21 or better
+                    }
+
+                    NotificationManager nm =
+                            (NotificationManager)context.getSystemService(Context.NOTIFICATION_SERVICE);
+
+                    nm.notify(ID_NOTIFICATION_BACKGROUND_ACTIVE, nb.build());
+
+                } else {
+                    context.stopService(new Intent(context, SyncthingService.class));
+                }
+                break;
+        }
+    }
+}
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
@@ -59,7 +59,6 @@ public class SyncthingService extends Service implements
     public static final String ACTION_RESET =
             "com.nutomic.syncthingandroid.service.SyncthingService.RESET";
 
-
     /**
      * Interval in ms at which the GUI is updated (eg {@link com.nutomic.syncthingandroid.fragments.DrawerFragment}).
      */
@@ -145,6 +144,12 @@ public class SyncthingService extends Service implements
     private State mCurrentState = State.INIT;
 
     /**
+     * Object that can be locked upon when accessing mCurrentState
+     * Currently used to male onDestroy() and PollWebGuiAvailableTaskImpl.onPostExcecute() tread-safe
+     */
+    private final Object stateLock = new Object();
+
+    /**
      * True if a stop was requested while syncthing is starting, in that case, perform stop in
      * {@link PollWebGuiAvailableTaskImpl}.
      */
@@ -227,10 +232,8 @@ public class SyncthingService extends Service implements
             if (mConfig != null) {
                 mCurrentState = State.STARTING;
 
-                if (mApi != null) {
+                if (mApi != null)
                     registerOnWebGuiAvailableListener(mApi);
-                    mApi.setWebGuiUrl(mConfig.getWebGuiUrl());
-                }
                 if (mEventProcessor != null)
                     registerOnWebGuiAvailableListener(mEventProcessor);
                 new PollWebGuiAvailableTaskImpl(getFilesDir() + "/" + HTTPS_CERT_FILE)
@@ -426,12 +429,25 @@ public class SyncthingService extends Service implements
 
     /**
      * Stops the native binary.
+     *
+     * The native binary crashes if stopped before it is fully active. In that case signal the
+     * stop request to PollWebGuiAvailableTaskImpl that is active in that situation and terminate
+     * the service there.
      */
     @Override
     public void onDestroy() {
-        super.onDestroy();
-        Log.i(TAG, "Shutting down service");
-        shutdown();
+
+        synchronized (stateLock) {
+            if (mCurrentState == State.INIT || mCurrentState == State.STARTING) {
+                Log.i(TAG, "Delay shutting down service until initialisation of Syncthing finished");
+                mStopScheduled = true;
+
+            } else {
+                Log.i(TAG, "Shutting down service immediately");
+                shutdown();
+            }
+        }
+
         SharedPreferences sp = PreferenceManager.getDefaultSharedPreferences(this);
         sp.unregisterOnSharedPreferenceChangeListener(this);
     }
@@ -512,14 +528,24 @@ public class SyncthingService extends Service implements
             super(httpsCertPath);
         }
 
+        /**
+         * Wait for the web-gui of the native syncthing binary to come online.
+         *
+         * In case the binary is to be stopped, also be aware that another thread could request
+         * to stop the binary in the time while waiting for the GUI to become active. See the comment
+         * for SyncthingService.onDestroy for details.
+         */
         @Override
         protected void onPostExecute(Void aVoid) {
-            if (mStopScheduled) {
-                mCurrentState = State.DISABLED;
-                onApiChange();
-                shutdown();
-                mStopScheduled = false;
-                return;
+            synchronized (stateLock) {
+                if (mStopScheduled) {
+                    mCurrentState = State.DISABLED;
+                    onApiChange();
+                    shutdown();
+                    mStopScheduled = false;
+                    stopSelf();
+                    return;
+                }
             }
             Log.i(TAG, "Web GUI has come online at " + mConfig.getWebGuiUrl());
             mCurrentState = State.STARTING;
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -470,4 +470,5 @@ Please report any problems you encounter via Github.</string>
     <!-- Format string for folder size, eg "500 MiB / 1 GiB" -->
     <string name="folder_size_format">%1$s / %2$s</string>
 
+    <string name="appconfig_receiver_background_enabled">Stopping Syncthing is not supported when running in background is enabled.</string>
 </resources>
--- a/src/androidTest/java/com/nutomic/syncthingandroid/test/MockContext.java
+++ b/src/androidTest/java/com/nutomic/syncthingandroid/test/MockContext.java
@@ -16,6 +16,7 @@ import java.util.List;
 public class MockContext extends ContextWrapper {
 
     private ArrayList<Intent> mReceivedIntents = new ArrayList<>();
+    private ArrayList<Intent> mStopServiceIntents = new ArrayList<>();
 
     /**
      * Use the actual context for calls that aren't easily mocked. May be null if those
@@ -36,8 +37,18 @@ public class MockContext extends ContextWrapper {
         return null;
     }
 
+    @Override
+    public boolean stopService(Intent intent) {
+        mStopServiceIntents.add(intent);
+        return true;
+    }
+
     public List<Intent> getReceivedIntents() {
         return mReceivedIntents;
+    }
+
+    public List<Intent> getStopServiceIntents() {
+        return mStopServiceIntents;
     }
 
     public void clearReceivedIntents() {
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -91,6 +91,12 @@
                 <action android:name="android.intent.action.BOOT_COMPLETED" />
             </intent-filter>
         </receiver>
+        <receiver android:name=".syncthing.AppConfigReceiver">
+            <intent-filter>
+                <action android:name="com.nutomic.syncthingandroid.action.START" />
+                <action android:name="com.nutomic.syncthingandroid.action.STOP" />
+            </intent-filter>
+        </receiver>
     </application>
 
 </manifest>
--- a/src/main/java/com/nutomic/syncthingandroid/activities/SyncthingActivity.java
+++ b/src/main/java/com/nutomic/syncthingandroid/activities/SyncthingActivity.java
@@ -33,14 +33,20 @@ public abstract class SyncthingActivity extends ToolbarBindingActivity implement
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         startService(new Intent(this, SyncthingService.class));
-        bindService(new Intent(this, SyncthingService.class),
-                this, Context.BIND_AUTO_CREATE);
     }
 
     @Override
-    protected void onDestroy() {
-        super.onDestroy();
+    protected void onPause() {
         unbindService(this);
+
+        super.onPause();
+    }
+
+    @Override
+    protected void onResume() {
+        super.onResume();
+
+        bindService(new Intent(this, SyncthingService.class), this, Context.BIND_AUTO_CREATE);
     }
 
     @Override


Nutomic:
So this is good to test now? What app did you use to test the intents?

sqrt-1764:
Grrr - commented on the wrong thread ;-)

I just committed my "final" version, if you want to take a look. Now I have to create the tests ...

The AppConfigReceiver-Service can be easily extended to support more commands.
Currently I implemented the action "sycthingservice" with the commands "start" and "stop".

In short words what I did:

"start" just (re)starts the Syncthingservice via context.startService().
"stop" stops the Syncthingservice by a call to startService with the new action "shutdown". This because a stopService command when the Syncthingservice wasn't active resulted in a runtime-error of the syncthing-binary.
Like this, the service can handle the shutdown internally and delay the stopService call if needed.

For the delay, I found the mStopScheduled member-variable which seems to be there for exact this job.
If mCurrentStatus is not INIT or STARTING, I stop the service immediately.
Otherwise I set mStopScheduled to true - which then does the rest.

sqrt-1764:
I tested this with automate
Created a flow for these two actions and hooked them to the connect and disconnect of my WIFI.

I then toggled my WIFI and watched Syncthing starting and stopping.

[SyncThing.flo.zip](https://github.com/syncthing/syncthing-android/files/190958/SyncThing.flo.zip)

sqrt-1764:
I think it is good for a first test now.

sqrt-1764:
Added tests for the AppConfigReceiver class.
Did not test the changed SyncThingService class.

Nutomic:
What error did you get with stopService? The [documentation](http://developer.android.com/reference/android/content/Context.html#stopService%28android.content.Intent%29) says it does nothing if the service is not running.

I tried to test it with Automate (imported your flow from the zip), but I think I'm doing something wrong. After pressing the start button, checking both items and pressing ok, it either starts or stops the service, but doesn't follow the wifi state. Also, Automate says "Running Fibers: Not Running".

When Syncthing is shut down, the "Syncthing is disabled" dialog is coming up (with a button leading to the settings). There should probably be a different dialog in this case, saying that Syncthing was disabled by an external application.

Also, it's "Syncthing", without capital T ;)

sqrt-1764:
Regarding the error:
You mean the error I got when I tried to stop the service while it was indexing?
That was a runtime-error from the syncthing binary I could see in the Android-Log. 
I could reproduce that error when I tried to stop the service with a call to stopService() before all my three folders were indexed.
When I waited long enough, I could stop the service without an error.
So that method was not failsafe and I looked for an alternative.

Regarding Automate:
Sorry - documentation was/is missing.
My flow only wraps the creation of the intents. This could be a flow that could be presented to the Automate-Users to simplify the interface. They can call it using one of the 2 entry-points and don't have to know about the exact values of the parameters of the intent. Then the rest is up to the phantasy of the user (so i omitted the part that checks the WIFI. Could as well be the battery state or time of day to control the service ...).
The flow has 2 entry-points - one for activation one for deactivation of the service. Checking both items  at the same time makes no sense. I could have created 2 separate flows, but like this the list of flows becomes not so messy. When you select a flow as a subroutine, you can only chose one entry-point.
(I consider the option to chose 2 or more starting-points as a bug - will report that there ;-) )

I don't get what you mean with "Syncthing is disabled" dialog.
When I stop the Syncthing-Service, the ongoing notification in the statusbar disappears and the device gets inactive on the device-lists of the other devices. No dialog is popping up for my devices.

Regarding the translations:
I am new to Transifex - will take a closer look now. ;-)

sqrt-1764:
I have merged the latest commits into my version and am testing it on my devices (Galaxy Nexus - Android 4.4, Nexus 7 (2012) - Android 5.1.1 (Cyanogen), Nexus 7 (2013) - Android 6, Nexus 5x - Android N Developer Preview.
The test of the first three started this noon, on the 5x I developed. So far it looks good. Only on the 5x I had had sporadic crashes. (One per day or so) but they did not seem to be related to starting or stopping the service. Syncthing crashed while the device was pausing or doing other things.
How to investigate this? Is there an option to switch on permanent logging? At the moment it seems that I have to manually collect the logs after a crrash? (Maybe because it is a debug-version?)

Nutomic:
If you get an error on stopService(), that seems to be a bug. Do you still have the log for that by any chance? If not, I'll look into it later. I'd rather fix the root issue than use this ugly hack of using startService to stop the service ;)

For Automate, that means I can just fire the start/stop intent by checking the correct item and pressing ok? That's totally fine for my testing then.

I'm not sure if devices keep a history of their logs. Best leave them connected to a PC with `adb logcat` running, or use a logcat app on the phone.

sqrt-1764:
For your convenience I added my first attempt to stop via stopService. Now there are two methods to stop the Syncthing-Service.
The "stop" value ot the "sycnthingservice" parameter still points to the method to shutdown via startService()
The "stop2" value points directly to stopService()

I also updated (and renamed ;-) ) the Syncthing.flo for automate. Now you have one option for start and 2 options to stop the service. Simply chose one and press OK to issue that command.

Currently strange things seem to happen if I call stopService too early. I had the case that I could not start the service after that. Apparently the service wasn't completely shut down and I got messages like
"Failed to call Rest API at https://127.0.0.1:8384/rest/events?since=0&limit=1"
in the log. Could only be cured by starting the Syncthing-App.
The other method to stop the service via startService() seems to work OK.

[Syncthing.flo.zip](https://github.com/syncthing/syncthing-android/files/195630/Syncthing.flo.zip)


Regarding the logs:
Difficult if the crash is rare. I saw them when I switched on my phone after a while or suddenly popping up when I was doing something else. But at the moment no further ones since yesterday.

Nutomic:
OK I just tried it and it seems to work perfectly. I like how Syncthing is started when you open it, and then automatically stopped when you close it, if the stop intent was last sent :)

Using stopService worked fine, I'd say we go with that and fix the crash if we get a log of that.

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * Broadcast-receiver to control and configure SyncThing remotely
+ *
+ * Created by sqrt-1764 on 25.03.16.
+ */
+public class AppConfigReceiver extends BroadcastReceiver {
+    private static final int ID_NOTIFICATION_BACKGROUND_ACTIVE = 2;
+
+    public static final String ACTION_SYNCTHINGSERVIVE = "syncthingservice";
```
Nutomic: You're using the action/extras kinda weird imo. Why not just use ACTION_START and ACTION_STOP? There's no need for extras here.

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
```diff
@@ -0,0 +1,101 @@
...
+
+                            NotificationCompat.Builder nb = new NotificationCompat.Builder(context)
+                                    .setWhen(System.currentTimeMillis())
+                                    .setContentText(msg)
+                                    .setTicker(msg)
+                                    .setStyle(new NotificationCompat.BigTextStyle().bigText(msg))
+                                    .setContentTitle(context.getText(context.getApplicationInfo().labelRes))
+                                    .setSmallIcon(context.getApplicationInfo().icon)
+                                    .setAutoCancel(true)
+                                    .setContentIntent(PendingIntent.getActivity(context, 0, new Intent(context, MainActivity.class), PendingIntent.FLAG_UPDATE_CURRENT));
```
Nutomic: The icon should be `R.drawable.ic_stat_notify`. setWhen() shouldn't be necessary (I think it's automatically set to the current time). setAutoCancel() is true by default. Do you see any effect on your device for setTicker()/setStyle()?

Basically, just set the things that are necessary, to keep it simple ;)

src/main/res/values/strings.xml
```diff
@@ -467,4 +467,5 @@ Please report any problems you encounter via Github.</string>
     <!-- Format string for folder size, eg "500 MiB / 1 GiB" -->
     <string name="folder_size_format">%1$s / %2$s</string>
 
+    <string name="appconfig_receiver_background_enabled">Stopping SyncThingService is not supported when running in background enabled.</string>
```
Nutomic: As I said, it's Syncthing without capital T :p 

And I would just write "Stopping Syncthing is not supported", no need to confuse people what a "service" is.

src/main/res/values-de/strings.xml
```diff
@@ -286,4 +286,5 @@ Bitte auftretenden Probleme via Github melden.</string>
   <string name="state_unknown">Unbekannt</string>
   <!--Format string for folder size, eg "500 MiB / 1 GiB"-->
   <string name="folder_size_format">%1$s / %2$s</string>
+  <string name="appconfig_receiver_background_enabled">Das Anhalten des SyncThingServices ist bei aktiviertem Hintergrunddienst nicht möglich.</string>
```
Nutomic: Please remove this and submit the translation via Transifex later ;)

Nutomic:
I got this exception from testStartStopSyncthingServiceForeground. Is that the one you meant?
```
java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String java.lang.String.intern()' on a null object reference
at android.os.Parcel.readException(Parcel.java:1552)
at android.os.Parcel.readException(Parcel.java:1499)
at android.app.ActivityManagerProxy.stopService(ActivityManagerNative.java:3462)
at android.app.ContextImpl.stopServiceCommon(ContextImpl.java:1740)
at android.app.ContextImpl.stopService(ContextImpl.java:1699)
at android.content.ContextWrapper.stopService(ContextWrapper.java:521)
at com.nutomic.syncthingandroid.syncthing.AppConfigReceiver.onReceive(AppConfigReceiver.java:63)
at com.nutomic.syncthingandroid.test.syncthing.AppConfigReceiverTest.testStartStopSyncthingServiceForeground(AppConfigReceiverTest.java:98)
at android.test.AndroidTestRunner.runTest(AndroidTestRunner.java:191)
at android.test.AndroidTestRunner.runTest(AndroidTestRunner.java:176)
at android.test.InstrumentationTestRunner.onStart(InstrumentationTestRunner.java:555)
at android.app.Instrumentation$InstrumentationThread.run(Instrumentation.java:1853)
```

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
sqrt-1764: My Idea behind this is to have a single entry-point to configure the app. If sometimes later other options are added, you can configure more than one property with one intent.

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/AppConfigReceiverTest.java
```diff
@@ -0,0 +1,107 @@
...
+                .putBoolean(SyncthingService.PREF_ALWAYS_RUN_IN_BACKGROUND, true)
+                .commit();
+
+
+        Intent intent = new Intent(new Intent(mContext, AppConfigReceiver.class));
+        intent.putExtra(AppConfigReceiver.ACTION_SYNCTHINGSERVIVE, "start");
+
+        mReceiver.onReceive(mContext, intent);
+        assertEquals("Start SyncthingService Background", 1, mContext.getReceivedIntents().size());
+        assertEquals("Start SyncthingService Background", "com.nutomic.syncthingandroid.syncthing.SyncthingService", mContext.getReceivedIntents().get(0).getComponent().getClassName());
```
Nutomic: You can just use `SyncthingService.class.getName()` instead of writing down the ful class name.

Also, I didn't even know you could add a message to the assert like that. Prett neat!

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/AppConfigReceiverTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+        intent = new Intent(new Intent(mContext, AppConfigReceiver.class));
+        intent.putExtra(AppConfigReceiver.ACTION_SYNCTHINGSERVIVE, "stop");
+
+            // onReceive now creates a notification.
+            // This is expected to throw an exception as creating a notification is not supported in this Test-Context
+            // so an exception is an indicator of success in this case.
+        try {
+            mReceiver.onReceive(mContext, intent);
+            throw new Exception("Stop SyncthingService Background - didn't fail to create the notification");
```
Nutomic: `Assert.fail()` woud be clearer here.

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/AppConfigReceiverTest.java
```diff
@@ -0,0 +1,107 @@
...
+        } catch (Exception e) {
```
Nutomic: Can you catch the exact exception that is thrown? This would actually catch the exception you were throwing above, or also any assert.

Nutomic:
Your tests seem kind of long. Can you split them up so each one has around 10 lines?

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
sqrt-1764: I just copied and pasted a piece of code I used for my other notifications.

SetTicker used to show a text when the notification first was displayed. This behaviour was changed later. But it still seems to be useful:
http://stackoverflow.com/questions/29725102/is-notification-builder-setticker-still-useful-in-android-5-and-above

Regarding setWhen and setAutocancel:
The documentation does not mention the default behaviour, so I explicitly did set those values to be on the safe side.
But I have no problem to remove those calls.

Nutomic: If never used setWhen/setAutoCancel, and I'd rather keep it simple. There's probably no harm in setTicker though.

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
Nutomic: I generally don't like the idea of writing code for potential later changes (in the end, it never gets used that way and is just needlessly complicated). And I don't see any need to start/stop Syncthing and change settings in one intent.

sqrt-1764: A thing that is in the back of my head would be the following:
I use Syncthing also for backup of large folders. So it would be nice to be able to disable synchronisation of some shares if I am "on the road" only enabling them when I am at home and charging - or something like this.
Those settings for the folders would fit perfectly into my current pattern. You could then pass a collection of folders to be enabled/disabled in the extras as well as a command to start or stop the service. Or maybe a restartService command might be handy in some cases?

IMHO this would be a perfect place to interface the app ... all external configuration-changes would then come through this receiver ...

If we limit the function to only starting and stopping the service then the Receiver also should be renamed to reflect this "specialized" behaviour. Something like "conrolSyncthingService" ...

src/main/res/values-de/strings.xml
sqrt-1764: Removed

src/main/res/values/strings.xml
sqrt-1764: OK, you are the boss ;-)

Changed to: "Stopping Syncthing is not supported when running in background enabled."

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
Nutomic: But the current intents start/stop Syncthing, so you can't pause/unpause folders. 

I'd rather use a seperate intent action for that, maybe PAUSE_FOLDER, with parameters folder IDs and pause state.

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/AppConfigReceiverTest.java
sqrt-1764: Agreed

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/AppConfigReceiverTest.java
sqrt-1764: Changed that.

We always learn new things. ;-)

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/AppConfigReceiverTest.java
sqrt-1764: You are perfectly right! My fault!

sqrt-1764:
As the tests do test only a mocked service, I could split testing start and stop to separate tests. (Otherwise I would have to start a service in order to stop it - which was the pattern I used before)

sqrt-1764:
Regarding the crashes:
On my current test device I have 3 shared with 3.x GB of data.
So in the log I can see 3 times a message like "Ready to synchronize <Share-Name>"

I I issue the stopService before I got those three ready-messages I get the following crash:

`04-02 01:03:53.542 8637-8637/com.nutomic.syncthingandroid.debug D/EventProcessor: Shutdown event processor.
04-02 01:03:53.544 8637-8637/com.nutomic.syncthingandroid.debug I/SyncthingService: Starting syncthing according to current state and preferences
04-02 01:03:53.549 8637-8637/com.nutomic.syncthingandroid.debug I/ConfigXml: Checking for needed config updates
04-02 01:03:53.740 8637-10455/com.nutomic.syncthingandroid.debug I/SyncthingNativeCode: [OKYFF] 23:03:53 INFO: syncthing v0.12.21 "Beryllium Bedbug" (go1.6 linux-arm) unknown-user@Colossus 2016-03-23 08:26:04 UTC
04-02 01:03:53.742 8637-10455/com.nutomic.syncthingandroid.debug I/SyncthingNativeCode: [OKYFF] 23:03:53 INFO: My ID: OKYFFXC-EPZ6WT6-MMWMTKJ-5EDUF47-2C7SE2I-UO22YQW-5DK36GM-45LIFQY
04-02 01:03:54.113 8637-10455/com.nutomic.syncthingandroid.debug I/SyncthingNativeCode: [OKYFF] 23:03:54 INFO: Single thread hash performance is ~13 MB/s
04-02 01:03:54.778 8637-10481/com.nutomic.syncthingandroid.debug W/SyncthingNativeCode: ionice: exec 10454: Permission denied
04-02 01:03:54.780 8637-10457/com.nutomic.syncthingandroid.debug I/SyncthingRunnableIoNice: ionice performed on libsyncthing.so
04-02 01:03:54.781 8637-10457/com.nutomic.syncthingandroid.debug E/SyncthingRunnableIoNice: Failed to set ionice 127
04-02 01:03:54.784 8637-10481/com.nutomic.syncthingandroid.debug W/SyncthingRunnable: Failed to read Syncthing's command line output
                                                                                      java.io.InterruptedIOException: read interrupted
                                                                                          at libcore.io.Posix.readBytes(Native Method)
                                                                                          at libcore.io.Posix.read(Posix.java:169)
                                                                                          at libcore.io.BlockGuardOs.read(BlockGuardOs.java:231)
                                                                                          at libcore.io.IoBridge.read(IoBridge.java:471)
                                                                                          at java.io.FileInputStream.read(FileInputStream.java:252)
                                                                                          at java.io.BufferedInputStream.read1(BufferedInputStream.java:273)
                                                                                          at java.io.BufferedInputStream.read(BufferedInputStream.java:334)
                                                                                          at sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:287)
                                                                                          at sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:350)
                                                                                          at sun.nio.cs.StreamDecoder.read(StreamDecoder.java:179)
                                                                                          at java.io.InputStreamReader.read(InputStreamReader.java:184)
                                                                                          at java.io.BufferedReader.fill(BufferedReader.java:165)
                                                                                          at java.io.BufferedReader.readLine(BufferedReader.java:328)
                                                                                          at java.io.BufferedReader.readLine(BufferedReader.java:393)
                                                                                          at com.nutomic.syncthingandroid.syncthing.SyncthingRunnable$2.run(SyncthingRunnable.java:307)
                                                                                          at java.lang.Thread.run(Thread.java:803)
04-02 01:03:57.688 8637-8637/com.nutomic.syncthingandroid.debug I/SyncthingService: Shutting down service
04-02 01:03:57.690 8637-8637/com.nutomic.syncthingandroid.debug D/EventProcessor: Shutdown event processor.
04-02 01:03:57.784 8637-10455/com.nutomic.syncthingandroid.debug I/SyncthingNativeCode: [OKYFF] 23:03:57 OK: Ready to synchronize Leo-DCIM (read-write)
04-02 01:03:58.873 8637-8637/com.nutomic.syncthingandroid.debug I/SyncthingRunnableKill: Killed Syncthing process 10454
04-02 01:04:02.053 8637-8637/com.nutomic.syncthingandroid.debug I/SyncthingRunnableKill: Killed Syncthing process 10454
04-02 01:04:02.088 8637-10450/com.nutomic.syncthingandroid.debug E/AndroidRuntime: FATAL EXCEPTION: Thread-9
                                                                                   Process: com.nutomic.syncthingandroid.debug, PID: 8637
                                                                                   java.lang.RuntimeException: Syncthing binary crashed with error code 137, output:
                                                                                   [OKYFF] 23:03:53 INFO: syncthing v0.12.21 "Beryllium Bedbug" (go1.6 linux-arm) unknown-user@Colossus 2016-03-23 08:26:04 UTC
                                                                                   [OKYFF] 23:03:53 INFO: My ID: OKYFFXC-EPZ6WT6-MMWMTKJ-5EDUF47-2C7SE2I-UO22YQW-5DK36GM-45LIFQY
                                                                                   [OKYFF] 23:03:54 INFO: Single thread hash performance is ~13 MB/s
                                                                                   [OKYFF] 23:03:57 OK: Ready to synchronize Leo-DCIM (read-write)`

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
sqrt-1764: Yes, it could be done that way.
But if the actions are too small, more than one intent might be needed to perform a certain task. Perform a configuration and then restart the service would need 2 intents - resulting in starting and stopping the service 2 times. 
For performance-reasons one intent would be better - but then more than one action would have to be configurable ... which is why i came to that solution to deliver my commands in the extras of the intent.

Nutomic: But you don't really want to start or stop Syncthing after a settings change, you want to restart it. So we could just always restart after receiving a settings change, or use a boolean extra DO_RESTART.

veniosg: I'd argue that "SyncthingService" is not really an action. It should at the very least contain a verb, especially if it's a public API.
It also has a typo in the name. Furthermore, it'd be wise to follow common conventions such as keeping the value namespaced, e.g. "android.intent.action.VIEW", so syncthing's actions should be "com.nutomic.syncthingandroid.action.<ACTION_NAME>"

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
```diff
@@ -0,0 +1,101 @@
...
+
+                            if (Build.VERSION.SDK_INT >= 21) {
```
veniosg: This should be using a constant from Build.VERSION_CODES.

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
```diff
@@ -57,6 +57,11 @@
      */
     public static final String ACTION_RESET = "reset";
 
+    /**
+     * Intent action to shut down the SyncthingService
+     */
+    public static final String ACTION_SHUTDOWN = "shutdown";
```
veniosg: Same point about the values of actions as the one above. Naming seems fine for these ones though.

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
sqrt-1764: Well, even Android Studio itself only gives the numerical value. But agreed, it looks nicer.
Changed

sqrt-1764:
Just pushed the new version
After I made friends with the action-parameter (we always can introduce a "bulk"-action later ;-) ) I now changed the AppConfigReceiver to use the action-field.

The changes:
- Changed the AppConfigReceiver-Intent from actions in intent-extras to actions in the action-property. 
- Renamed the actions according to the naming conventions of Android. 
- Moved the code that handles the service to a private inner static class to isolate that functionality. 
- Adjusted the Instrumentation-Tests

sqrt-1764:
... and the updated version of the Syncthing-Flow for those who test with Automate:

[Syncthing.flo.zip](https://github.com/syncthing/syncthing-android/files/201314/Syncthing.flo.zip)

Nutomic:
The error you had during shutdown was caused by SIGKILL, which we use to stop Syncthing. I've pushed a commit to ignore this exit code, so this should work fine now. That means you can really get rid of the "shutdown via start" code :D 

I haven't tried this version again, because I assume it still works. The code looks also fine at a glance. I just noticed you have a test commented out, why is that?

sqrt-1764:
There is a TODO entry in the comment ;-)
The current mocked context that I reused for my tests doesn't support the stopService command and crashes. So I delayed the test until it is really needed.

Will test if the crash went away and then switch to the stopService-Variant.

Nutomic:
Sorry I was kinda lost in all the comments there :p 

You can probably just forward the call to the actual context in MockContext (depending on what you need, haven't looked closer yet).

sqrt-1764:
I merged the current master from you and tested. I still get this erroror am I doing something wrong?

`04-04 22:59:19.488 16423-16423/com.nutomic.syncthingandroid.debug D/EventProcessor: Shutdown event processor.
04-04 22:59:20.674 16423-16423/com.nutomic.syncthingandroid.debug I/SyncthingRunnableKill: Killed Syncthing process 17235
04-04 22:59:23.834 16423-16423/com.nutomic.syncthingandroid.debug I/SyncthingRunnableKill: Killed Syncthing process 17235
04-04 22:59:23.872 16423-17233/com.nutomic.syncthingandroid.debug E/AndroidRuntime: FATAL EXCEPTION: Thread-9
                                                                                    Process: com.nutomic.syncthingandroid.debug, PID: 16423
                                                                                    java.lang.RuntimeException: Syncthing binary crashed with error code 137, output:
                                                                                    [OKYFF] 20:59:11 INFO: syncthing v0.12.21 "Beryllium Bedbug" (go1.6 linux-arm) unknown-user@Colossus 2016-03-23 08:26:04 UTC
                                                                                    [OKYFF] 20:59:11 INFO: My ID: OKYFFXC-EPZ6WT6-MMWMTKJ-5EDUF47-2C7SE2I-UO22YQW-5DK36GM-45LIFQY
                                                                                    [OKYFF] 20:59:12 INFO: Single thread hash performance is ~13 MB/s
                                                                                    [OKYFF] 20:59:16 OK: Ready to synchronize Leo-DCIM (read-write)`

AudriusButkevicius:
This is not your bug, as far as I understood from convos on IRC.

Nutomic:
Did you rebase on master?

sqrt-1764:
No, I simply fetched your master branch, merged that into my copy of the master branch and then merged that master branch (that now should be identical to your master branch) into my appconfig-receiver branch.
According to the docs rebase only makes a cleaner history(?)

Nutomic:
I'm not really sure what you're doing there. You can just run `git rebase master`. And yeah, it will safe you some cleaning up of the commits ;)

And actually, I didn't push my changes before, but I did now.

sqrt-1764:
Just to make shure (I'm not a git-master yet) - what id did on my local repository was:
git fetch upstream/master
git checkout master
git merge upstream/master
git checkout appconfig-receiver
git merge master

Should get my all your changes into my appconfig-receiver branch

AudriusButkevicius:
You should probably squash all your commits, checkout master, update master from upstream, check out your branch, and do git rebase master.

What you are doing now is merging master into your branch, where as the prefered way is to move your branch on top of master.

sqrt-1764:
OK, I'll look into that

AudriusButkevicius:
Just make a new branch on current head incase you mess something up

sqrt-1764:
But I think, my current configuration is also correct, so I post the result of my current tests:
I started Syncthing from Automate via the startService command
I then immediately issued the stopService command from there.
Now I can see that the service is stopped and no runtime-error any more.

But:
I now can't start Syncthing via startService command any more.
If I try to start the Syncthing-App I get the start-dialog running forever.
After I killed that process (swiping it off the tasklist) the startService-Command issued before apparently is executed - I can see Syncthing starting.

AudriusButkevicius:
actually I think nutomic can get github to squash and it will be fine.

sqrt-1764:
No problem - I also want to master git. So it is a good exercise for me. (And always backup before making big changes ;-) )

Nutomic:
I think the new Github squashing only works when you actually merge the PR (at least I can't find a button otherwise).

Here's a guide on squashing commits:
http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

sqrt-1764:
If I understood it right, my merging-chain was also correct but produced a result that was hard to read.
I now issued a simple
     git rebase upstream/master
and now all of my commits are on top of the current master - what was the preferred solution.
I don't see the the need to squash all my commits - personally I like to be able to also see what I maybe did wrong in the past.
Or is there an option to push a squashed version while keeping an unsquashed version locally?

AudriusButkevicius:
So you can create a new branch at the same location, and then squash this branch.
The history will be lost in this repo, but you will still have your branch.

The reason for squashing is to have clean history and a commit per feature, so that when you annotate, you don't need to assemble it from many small pieces.

sqrt-1764:
Did a rebase on the current upstream master, created my private dev-branch. Then squashed the commits to one and force-pushed that branch to my github-fork.

Currently I have left the 2 stop-Methods in the code because I have the reported issues with the stopservice-Variant.

sqrt-1764:
In case of later changes: Shall I sqash only to my last public push or back to that rebased master then?

Nutomic:
You're still getting the same crash? Do you have 7f090e3653485da364369b4d79f9b80181716a26 listed in `git log`?

For later changes, you can either use `git commit --amend` to add changes to your existing commit, or you can squash. Ideally you rebase on master before we merge it, in case there are any conflicts.

sqrt-1764:
Yes, I have this commit in my git branch.
As I reported yesterday: That runtime-crash is now gone.
But under some circumstances (I stopped the service immediately after starting it while it was indexing my 3 GB data) the sercie stopped, but I wasn't able to start the service again.
I then noticed that the syncthing GUI didn't start any more. The starthing-dialog was shown forever.
When I put that into background and then swept it away in the tasklist (killing the process) the Syncthingservice started. Apparently from the startcommand issued later before.

sqrt-1764:
Regarding git: I plan to rebase on upstream/master on every public push I make.
As I want to try to privately keep my fine granulated commits, I will look into a proper way to sqash before my commits.
So you prefer to merge only one commit if the time comes - OK.

Nutomic:
Ah right. Like I said, I've already experienced this before, and I'm pretty sure it has nothing to do with how you stop the service. But if you can get any logs of this, I can maybe figure out what's going wrong.

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/AppConfigReceiverTest.java
```diff
@@ -0,0 +1,169 @@
...
+                .edit()
+                .putBoolean(SyncthingService.PREF_ALWAYS_RUN_IN_BACKGROUND, true)
+                .commit();
+
+        Intent intent = new Intent(new Intent(mContext, AppConfigReceiver.class));
+        intent.setAction(AppConfigReceiver.ACTION_START);
+
+        mReceiver.onReceive(mContext, intent);
+        assertEquals("Start SyncthingService Background", 1, mContext.getReceivedIntents().size());
+        assertEquals("Start SyncthingService Background", SyncthingService.class.getName(), mContext.getReceivedIntents().get(0).getComponent().getClassName());
```
Nutomic: Please limit your lines to 100 characters.

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/AppConfigReceiverTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+        Intent intent = new Intent(new Intent(mContext, AppConfigReceiver.class));
+        intent.setAction(AppConfigReceiver.ACTION_STOP);
+
+        // onReceive now tries to create a notification.
+        // This is expected to throw an exception as creating a notification is not supported in this Test-Context
+        // so an exception is an indicator of success in this case.
+        try {
+            mReceiver.onReceive(mContext, intent);
+            Assert.fail("Stop SyncthingService Background (1) - didn't fail to create the notification");
```
Nutomic: This message seems wrong, it would fail because it failed to create the notification, right?

Also, I'm not sure we need these messages. If a test fails, we get the test case and line logged, so there's no problem finding it. But I don't have a problem with it either.

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/AppConfigReceiverTest.java
sqrt-1764: Well, as long as it fits on my screen I usually prefer fewer lines.
But I now added some linefeeds.

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/AppConfigReceiverTest.java
sqrt-1764: Excactly. The mocked environment crashes, if a notification is to be shown. So I expect the test to crash in that case(s).
So what's wrong with the message? It states exacly this.

I like this messages when I am in a debug session. From the message I can see where and why the test failed without jumping into the code ... so I created those rather verbose messages.

sqrt-1764:
Pretty easy to reproduce the problem. Exactly the steps I mentioned.


04-06 00:18:32.756 2654-3870/com.nutomic.syncthingandroid.debug I/ConfigXml: Checking for needed config updates
04-06 00:18:32.759 2654-2654/com.nutomic.syncthingandroid.debug I/SyncthingService: Web GUI will be available at https://127.0.0.1:8384
04-06 00:18:32.760 2654-2654/com.nutomic.syncthingandroid.debug D/EventProcessor: Shutdown event processor.
04-06 00:18:32.762 2654-2654/com.nutomic.syncthingandroid.debug I/SyncthingService: Starting syncthing according to current state and preferences
04-06 00:18:32.768 2654-2654/com.nutomic.syncthingandroid.debug I/ConfigXml: Checking for needed config updates
04-06 00:18:32.966 2654-3874/com.nutomic.syncthingandroid.debug I/SyncthingNativeCode: [OKYFF] 22:18:32 INFO: syncthing v0.12.21 "Beryllium Bedbug" (go1.6 linux-arm) unknown-user@Colossus 2016-03-23 08:26:04 UTC
04-06 00:18:32.971 2654-3874/com.nutomic.syncthingandroid.debug I/SyncthingNativeCode: [OKYFF] 22:18:32 INFO: My ID: OKYFFXC-EPZ6WT6-MMWMTKJ-5EDUF47-2C7SE2I-UO22YQW-5DK36GM-45LIFQY
04-06 00:18:33.351 2654-3874/com.nutomic.syncthingandroid.debug I/SyncthingNativeCode: [OKYFF] 22:18:33 INFO: Single thread hash performance is ~13 MB/s
04-06 00:18:34.001 2654-3912/com.nutomic.syncthingandroid.debug W/SyncthingNativeCode: ionice: exec 3873: Permission denied
04-06 00:18:34.006 2654-3876/com.nutomic.syncthingandroid.debug I/SyncthingRunnableIoNice: ionice performed on libsyncthing.so
04-06 00:18:34.006 2654-3876/com.nutomic.syncthingandroid.debug E/SyncthingRunnableIoNice: Failed to set ionice 127
04-06 00:18:34.026 2654-3912/com.nutomic.syncthingandroid.debug W/SyncthingRunnable: Failed to read Syncthing's command line output
                                                                                     java.io.InterruptedIOException: read interrupted
                                                                                         at libcore.io.Posix.readBytes(Native Method)
                                                                                         at libcore.io.Posix.read(Posix.java:169)
                                                                                         at libcore.io.BlockGuardOs.read(BlockGuardOs.java:231)
                                                                                         at libcore.io.IoBridge.read(IoBridge.java:471)
                                                                                         at java.io.FileInputStream.read(FileInputStream.java:252)
                                                                                         at java.io.BufferedInputStream.read1(BufferedInputStream.java:273)
                                                                                         at java.io.BufferedInputStream.read(BufferedInputStream.java:334)
                                                                                         at sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:287)
                                                                                         at sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:350)
                                                                                         at sun.nio.cs.StreamDecoder.read(StreamDecoder.java:179)
                                                                                         at java.io.InputStreamReader.read(InputStreamReader.java:184)
                                                                                         at java.io.BufferedReader.fill(BufferedReader.java:165)
                                                                                         at java.io.BufferedReader.readLine(BufferedReader.java:328)
                                                                                         at java.io.BufferedReader.readLine(BufferedReader.java:393)
                                                                                         at com.nutomic.syncthingandroid.syncthing.SyncthingRunnable$2.run(SyncthingRunnable.java:315)
                                                                                         at java.lang.Thread.run(Thread.java:803)
04-06 00:18:35.939 2654-2654/com.nutomic.syncthingandroid.debug I/SyncthingService: Shutting down service
04-06 00:18:35.939 2654-2654/com.nutomic.syncthingandroid.debug D/EventProcessor: Shutdown event processor.
04-06 00:18:37.105 2654-2654/com.nutomic.syncthingandroid.debug I/SyncthingRunnableKill: Killed Syncthing process 3873
04-06 00:18:37.763 2654-3874/com.nutomic.syncthingandroid.debug I/SyncthingNativeCode: [OKYFF] 22:18:37 OK: Ready to synchronize Leo-DCIM (read-write)
04-06 00:18:40.283 2654-2654/com.nutomic.syncthingandroid.debug I/SyncthingRunnableKill: Killed Syncthing process 3873



Whereas when I wait until all three folders are up and ready I get:
...
04-06 00:22:22.338 7082-7082/com.nutomic.syncthingandroid.debug I/EventProcessor: Unhandled event StateChanged
04-06 00:22:22.339 7082-7082/com.nutomic.syncthingandroid.debug I/EventProcessor: Unhandled event StateChanged
04-06 00:22:22.339 7082-7082/com.nutomic.syncthingandroid.debug I/EventProcessor: Unhandled event FolderCompletion
04-06 00:22:22.340 7082-7082/com.nutomic.syncthingandroid.debug I/EventProcessor: Unhandled event RemoteIndexUpdated
04-06 00:22:22.345 7082-7082/com.nutomic.syncthingandroid.debug I/EventProcessor: Unhandled event StateChanged
04-06 00:22:24.485 7082-7113/com.nutomic.syncthingandroid.debug I/SyncthingNativeCode: [OKYFF] 22:22:24 WARNING: Failed to set UPnP port mapping: external port 0 on device 'FB-Leo' (192.168.84.1).
04-06 00:22:32.798 7082-7082/com.nutomic.syncthingandroid.debug I/SyncthingService: Shutting down service
04-06 00:22:32.798 7082-7082/com.nutomic.syncthingandroid.debug D/EventProcessor: Shutdown event processor.
04-06 00:22:33.558 7082-7113/com.nutomic.syncthingandroid.debug I/SyncthingNativeCode: [OKYFF] 22:22:33 OK: Exiting
04-06 00:22:33.959 7082-7082/com.nutomic.syncthingandroid.debug I/SyncthingRunnableKill: Killed Syncthing process 7112

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/AppConfigReceiverTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+        } catch (NullPointerException e) {
```
Nutomic: I really don't like to catch NPE like this, because it can easily hide actual errors. 

I'm not sure why this crash happens. The error is somewhere in the call to `PendingIntent.getActivity()`, but I don't see any reason for that. But instead of catching the NPE, you can override `getContentResolver()` in MockContext, and throw a custom exception (or set a flag).

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
```diff
@@ -0,0 +1,117 @@
...
+ * Broadcast-receiver to control and configure SyncThing remotely
+ *
+ * Created by sqrt-1764 on 25.03.16.
+ */
+public class AppConfigReceiver extends BroadcastReceiver {
+    private static final int ID_NOTIFICATION_BACKGROUND_ACTIVE = 2;
+
+    public static final String ACTION_START = "com.nutomic.syncthingandroid.action.ACTION_SYNCTHING_START";
+    public static final String ACTION_STOP  = "com.nutomic.syncthingandroid.action.ACTION_SYNCTHING_STOP";
+    public static final String ACTION_STOP2 = "com.nutomic.syncthingandroid.action.ACTION_SYNCTHING_STOP2";
```
Nutomic: I'm not sure if there's a standard for actions, but I've used the following pattern for other actions in the app now:
`com.nutomic.syncthingandroid.service.SyncthingService.RESTART`

So "full package" + "class name" + "ACTION_NAME". I'd prefer to use the same pattern for these actions, too. Unless you found some docs that say otherwise.

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
```diff
@@ -0,0 +1,117 @@
...
+                SyncthingHelper.startSynchthingService(context);
+                break;
+
+            case ACTION_STOP:
+                SyncthingHelper.stopSynchtingService(context);
+                break;
+
+            case ACTION_STOP2:
+                SyncthingHelper.stopSynchtingService2(context);
+                break;
```
Nutomic: I don't see any reason to use a static helper class here. It makes sense to put `stopSynctingService()` into a method because it's pretty long, but that should be in the same class. `startSyncthingService()` doesn't need an extra method, because it's only one line. And you have some typos (Sync**h**thing).

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
```diff
@@ -0,0 +1,117 @@
...
+
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+                    nb.setCategory(Notification.CATEGORY_ERROR);        // Only supported in API 21 or better
+                }
+
+                NotificationManager nm = (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
+                nm.notify(ID_NOTIFICATION_BACKGROUND_ACTIVE, nb.build());
+
+            } else {
+                Intent intent1 = new Intent(context, SyncthingService.class);
```
Nutomic: No need to number your variables, just call it `intent`.

src/main/res/values/strings.xml
```diff
@@ -470,4 +470,5 @@ Please report any problems you encounter via Github.</string>
     <!-- Format string for folder size, eg "500 MiB / 1 GiB" -->
     <string name="folder_size_format">%1$s / %2$s</string>
 
+    <string name="appconfig_receiver_background_enabled">Stopping Syncthing is not supported when running in background enabled.</string>
```
Nutomic: "when running in background **is** enabled"

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
```diff
@@ -0,0 +1,117 @@
...
+            if (SyncthingService.alwaysRunInBackground(context)) {
+                final String msg = context.getString(R.string.appconfig_receiver_background_enabled);
+
+                NotificationCompat.Builder nb = new NotificationCompat.Builder(context)
+                        .setContentText(msg)
+                        .setTicker(msg)
+                        .setStyle(new NotificationCompat.BigTextStyle().bigText(msg))
+                        .setContentTitle(context.getText(context.getApplicationInfo().labelRes))
+                        .setSmallIcon(R.drawable.ic_stat_notify)
+                        .setContentIntent(PendingIntent.getActivity(context, 0, new Intent(context, MainActivity.class), PendingIntent.FLAG_UPDATE_CURRENT));
```
Nutomic: Sorry I was wrong here, you actually need to call `.setAutoCancel(true)`.

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/AppConfigReceiverTest.java
Nutomic: Not everyone has the same screen :p 

And if you're doing side-by-side diffs, space runs out really fast. Terminals may also have limited width.

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/AppConfigReceiverTest.java
Nutomic: The problem is in a case like this:
```
Object x = null;
x.doSomething();
```
And that may be much more hidden in actual code. Your test will just assume everything worked fine, and won't be able to find that error.

Like I said, I don't have a problem with the messages, so keep them if you want (but the message seems to be wrong in this one).

Nutomic:
I suppose you're using a slow device and/or lots of data? I'll try to reproduce it, but I'm pretty sure you'll get the same if you close the app via the menu button (which also just sends `stopService()`).

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
sqrt-1764: Well in this case I would like to point to an argument from you from another thead:
This will be a public interface and therefore the user isn't aware of the internal structure of the app. The only thing that the user is interested in is the action. IMHO _where_ the action is executed should not be exposed in the public. Maybe it changes later?

Look at the values that usually are used in the intent filters in the android-manifest. I followed that pattern.

Whereas the last part of the String (the command) can be discussed. Maybe "START" and "STOP" are sufficient?

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
sqrt-1764: Typos fixed ...

Well, when I created this class I had the idea that it might me nice to have the function isolated from the use so it can be reused more easily in later cases.
If you ever need to stop Syncthing for other reasons there then is a well tested way to do so. And then the isolation would be handy and would make refactoring (moving the functions to other places) more easy.
Otherwise it isn't reallly needed in the current use-case yet.

I'm not shure, but maybe these functions could be moved to the Application-object then? That's the place I would look for such a function in the first place.

sqrt-1764:
Well, it's a Nexus 5x 32 GB, running the Developer Preview for Android N. So it is not the slowest device on the market. ;-)
I sync 3 GB data in 3 differenf folders.

But anyway - if it produces an error, I won't be the only one that will get it.

And no, when you close the app via the close menu button, the service seems to be fully started. Before the menu is disabled. In that case the stopService-command works. Only if you stop the service before it is fully started, you get the problem. (That's what I observed)

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/AppConfigReceiverTest.java
sqrt-1764: Yes, I am also not very happy with my current solution. Catching Null-Pointer-Exceptions is very bad practise, But I still had no better idea yet.

Yes, the call to PendingIntent.getActivity crashes and that apparently comes from the mocked content that is used there. The code works in "production" mode. The only thing that differs in test is the context.

But I don't see the getContentResolver here. So why do you think I can catch the error there?

So in this case - the code is short and the test focus is limited - I could live with it (although I'm not happy with it)

src/main/res/values/strings.xml
sqrt-1764: Fixed

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
Nutomic: That's a good point, especially seeing as this is a public interface (the other intents/actions aren't). But both "action" and syncthing appear twice in the string. I would just go with `com.nutomic.syncthingandroid.action.START` then.

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
Nutomic: Starting the service is a one-liner, so having an extra function is more (mental) overhead than it's worth.

And the stopping is only so long because of the notification, which we wouldn't need when stopping the service from any other place. But it seems like a good idea to put it in a seperate function (just not in a new class).

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/AppConfigReceiverTest.java
Nutomic: I don't understand either why it crashes, seeing as any calls are passed through to the normal context.

I don't really know what we should do here. But getContentResolver() is called directly before the NPE occurs. That means we can check that call for the same result, and without the side effect of hiding all other NPEs.

Nutomic:
Ah that makes sense. But there's no need to change the external API for this. Instead, it should be enough to set `mStopScheduled = true` in `onDestroy()`, if Syncthing is not started yet. I will look into that (if you're not faster).

Nutomic:
Actually, `mStopScheduled` is not enough, because that only stops Syncthing, but does not stop the service. So you should add a new variable `mDestroyScheduled`, that forces a call to `stopSelf()` after Syncthing is started.

sqrt-1764:
Just pushed my new version.

I now fixed the stopService-Method so that the other method via startService() became obsolete. So I deleted that part from the code.

The current method works as follows:
In onDestroy I check if Syncthing is currently starting. If that is the case PollWebGuiAvailableTaskImpl is active and waits for the Gui to become available. In its onPostExcecute() there is the only place where the value of mStopScheduled is read - so here is where the action is. ;-)
In order to prevent thread-concurrency problems I introduced a lock-object on which I synchronize so one thread has to wait for the other here.
In onPostExecute I also added a stopSelf() that I thing was missing there. (Could not add it to shutdown() as it is used also elsewhere in different context.)

There could be the rare possibility that shutdown is called twice. (E.g. when Syncthing was cancelled by the GUI and at the same time by a stopservice-Command via the broadcast-Receiver while the shutdown was not complete (otherwise the stopService() would not have called onDestroy()). But it seems to me that that case is handled OK by the current implementation of shutdown().

I also checked the full set of instrumentation tests and found that there were three other tests that are currently failing.
I couldn't withstand to fix testGetReadableTransferRate as it looks that it only was a matter of changed return-values.

But testImportExportConfig() fails with exact the cause that my stopService-Test is failing - a notification can not be created.So my first question before further investigation is - did this ever run?

sqrt-1764:
Regarding the SyncthingHelper class in AppConfigReceiver:
To me it is a matter of consistency in the level of abstaction. If I use a method to stop something, I would expect a method to start that thing at the same "level" - even if it is a one-liner. This would then become the one and only official point to start the service.
And I found hints, that such an implementation makes sense. In the code there sometimes is mentioned that the user could _disable_ Syncthing. In that case a startService should be rejected. And this startSyncthingService-Method would be the perfect place to implement this ...

sqrt-1764:
Just pushed my "final" version:
I now also fixed the problem with the stopservice-test.
I had to use mContext.getApplicationContext() - then the intents and pendingIntents didn't crash any more and I could rewrite the test to check that no stopService-Command was issued when background-mode was enabled.

I tried the same in SyncthingServiceTest.testImportExportConfig() which crashes with the same reason. After changing the context it still crashes, but now in the codebase. So I left it as it is as it is another case.

sqrt-1764:
Sorry for the inconvenience:
While fiddeling with the git rebase and squash, somehow a version of AppConfigReceiver with the wrong intent-action values sneaked back in. I just uploaded a fixed version.

Nutomic:
Sorry, I know I'm really slow with reviewing this. I didn't forget about it, it's just hard finding the time to review it. I'll try and look at your new changes during the weekend.

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
```diff
@@ -0,0 +1,87 @@
...
+import com.nutomic.syncthingandroid.R;
+import com.nutomic.syncthingandroid.activities.MainActivity;
+
+/**
+ * Broadcast-receiver to control and configure SyncThing remotely
+ *
+ * Created by sqrt-1764 on 25.03.16.
+ */
+public class AppConfigReceiver extends BroadcastReceiver {
+    private static final int ID_NOTIFICATION_BACKGROUND_ACTIVE = 2;
```
Nutomic: We're already using that notification ID in RestartActivity. You have to choose a different one, or they will replace each other.

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/RestApiTest.java
```diff
@@ -82,8 +80,8 @@ public void testReadableFileSize() {
     }
     
     public void testGetReadableTransferRate() {
-        assertEquals("1 MiB/s", RestApi.readableTransferRate(getContext(), 8388608L));
-        assertEquals("1 GiB/s", RestApi.readableTransferRate(getContext(), 8589934592L));
+        assertEquals("1 Mb/s", RestApi.readableTransferRate(getContext(), 8388608L));
+        assertEquals("1 Gb/s", RestApi.readableTransferRate(getContext(), 8589934592L));
```
Nutomic: These changes don't seem to be related to this PR (and actually break the test).

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/AppConfigReceiverTest.java
Nutomic: You still have some very long lines (that are even too long for Github).

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
```diff
@@ -0,0 +1,87 @@
...
+                break;
+
+            case ACTION_STOP:
+                SyncthingHelper.stopSynchtingService(context);
+                break;
+        }
+    }
+
+
+    private static class SyncthingHelper {
```
Nutomic: I still don't think we need this helper class. Just put the code from `startSyncthingService()` directly into the switch, and put `stopSynchtingService()` directly in AppConfigReceiver. We can always change this later if needed.

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
```diff
@@ -429,9 +432,19 @@ public IBinder onBind(Intent intent) {
      */
     @Override
     public void onDestroy() {
-        super.onDestroy();
-        Log.i(TAG, "Shutting down service");
-        shutdown();
+//        super.onDestroy();    Call of super not required -> https://developer.android.com/guide/components/services.html
```
Nutomic: I'd prefer to use a Javadoc that describes what this method does, instead of this comment (which isn't really helpful).

Nutomic:
Regarding the helper functions in AppConfigReceiver, and disabling the app: Disabling is actually already handled by DeviceStateHolder and SyncthingService. Adding new methods for this only makes things more complicated. 

However, we could think about moving all the enable/disable logic into DeviceStateHolder, with a method `isSyncthingEnabled()`. Then we could also handle the start/stop intents in that place. And we could have a variable to remember which intent was sent last, and have Syncthing disabled if the user starts it. But that doesn't have to be part of this PR.

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
sqrt-1764: Well, I made this comment only to inform you why I deleted that line and put a reference for the reason there. The call to super isn't required according to Google. So to follow your style to remove unecessary code, I deleted that line.

Regarding the javadoc:
Shall I describe the change in a javadoc assigned to onDestroy?
Then also a comment to the onPostExecute would be appropriate as the other half of the change lives there.

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
sqrt-1764: Currently, after removing the other stop-method this class isn't very necessary, I agree.
So I will now  move the code out of the class and directly to the switch statements.

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/RestApiTest.java
sqrt-1764: So what shall I do then?
Change my changes back? Also remove the test? Or leave it like it is?

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
sqrt-1764: You mean the value of the notification id. ;-)
OK, I only had found one reference and icremented my value from there.
So is the value 3 OK? Or is there maybe another reference?

Or would it be a good idea to bundle these constants in a single place - e.g. the application-object?

Nutomic: Grep didn't show anything else, so this should be fine. I'm not sure about moving them somewhere else, we can think about this later.

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/RestApiTest.java
Nutomic: Just rebase on master, I've already removed those tests.

sqrt-1764:
That is why I made my comments describing my ideas why i implemed it like that.
I also argree that that should be another change.
If you wish, I could take a look into that matter then. ;-)

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
Nutomic: Okay thats cool. Yeah, onDestroy() got somewhat complicated with these changes, so you should add a comment saying what it does in which case (and probably mention that onDestroy is not called and not required). onPostExecute() doesn't look so complicated, but more documentation surely won't hurt.

Nutomic:
The main problem/inconsistency I see now is this: If the app is disabled due to in-app settings (only sync on wifi etc), you can't start it, and a dialog will come. If the app is stopped via these intents, it can still be started by the user. That could cause some serious data usage if users don't pay attention. So it would be good to have this consistent.

I actually didn't think about this case when you started working about this, sorry about that.

sqrt-1764:
Well I see it like this:
If the user sends an intent to start the service then the user knows what he does. He wants it - he gets it.
But as I mentioned in my previous comments, I also have no problems ignoring the intents, if the app is disabled by the settings.

So what about implementing a (fake) boolean method isSyncthingEnabled in the devicestateHolder now, making the start-Intent aware of that function and then fill the isSyncthingEnabled with life maybe in a second change?

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
sqrt-1764: Well, AndroidStudio is quite good at refactoring ;-)
I could move my constant to a place you propose, and you could move your constants there too at a later time ... ;-)

sqrt-1764:
When I make my changes - shall I sqash back to my first change and force push that version?
Or shall I append my changes to the current state?

Nutomic:
Yeah, best squash and force push. 

I'm not quite sure how you want to make this "fake" method. We can just handle that in a new PR later.

sqrt-1764:
Well, just a method that returns a constant value with a javadoc describing the intended purpose.

So the hook is implemented an can be used, but may be implemented later.
But we also can postpone that.

Am 19. April 2016 17:27:41 MESZ, schrieb Felix Ableitner <notifications@github.com>:

sqrt-1764:
Back at keyboard and implemented your requests.
Only 2 items left:
- Move ID_NOTIFICATION_BACKGROUND_ACTIVE to a place where all application-wide constants can be collected? Would make it easier to introduce new constants if needed in the future.
Personally I would move such constants to the application-object which has the longest liifetime. Every other class would require to load unnecessary code in some cases. (If I understand the lifetime of the java-objects right.) But an application-object doesn't exist yet.

- Implement a dummy isSyncthingEnabled() function and query that for the start-service logic?
If yes - where to put it? I think it should be a static method that can be called from everywhere to any time. But DeviceStateHolder is a BroadcastService. So if it is put there, that class would always have to stay in memory.
Again the application-object would be a candidate. Or maybe an extra singleton?
... but then that might be a candidate for an extra change.

sqrt-1764:
Ah, and before you asked: I reintroduced setAutoCancel at my notification in AppConfigReceiver. Without it, the notification is not cancelled when the user clicks it - which is the way I would like to see it.

Nutomic:
Moving the notification IDs to a common place should be part of a seperate PR. Where that is doesn't matter, because the IDs are static final, so the containing object doesn't need to be instantiated.

And I don't see the point of having a dummy method in the code. Having such a method definitely makes sense, but we should also do it in a seperate PR.

ext/syncthing/src/github.com/syncthing/syncthing
```diff
@@ -1 +1 @@
-Subproject commit 668eb7c398c2e8d0b9f7beb319b7789681d7d89f
+Subproject commit 6d280e7b6461d407c90f8706e348393c657e995f
```
Nutomic: Ah I missed this, you shouldn't change anything in the submodule.

Nutomic:
Did you forget to push your changes?

ext/syncthing/src/github.com/syncthing/syncthing
sqrt-1764: I did not intend to change something there. Besides my changes in the java-classes, I only rebased on your master - could the change come from there?

sqrt-1764: Regarding the static finals: I am not 100% shure if the corresponding class is loaded if you use a static final field of that class. Technically it would not be necessary (if the initial value is a constant), but I did not find a clear reference for that in the web. So I thought defensive and would put the variable to the object that is living the longest fime ;-)

Regarding the dummy method: My thought was, that the later implementation of that dummy method would not change the other code. So a later change would be more focused.
But looking at the whole change it would make no difference - you have to make a change at both places.

So I forget about the dummy method for now.

sqrt-1764:
Just pushed my changed.
Not shure what to to about the change in the submodule as I don't know what has caused that change.

ext/syncthing/src/github.com/syncthing/syncthing
Nutomic: This seems to be because of [this accidental change](https://github.com/syncthing/syncthing-android/commit/6a686cdb9571980d0b72205c37f785aa510d1c08), your rebase didn't take that change for same reason. I reverted that already, so you won't have to do anything (6d280e is correct).

Static fields can be accessed without an object, just with a class. So this isn't a problem. But as I said, that's something for a seperate PR.

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/RestApiTest.java
```diff
@@ -75,7 +73,7 @@ public void onReceiveSystemInfo(RestApi.SystemInfo info) {
     public void testGetFolders() {
         assertNotNull(mApi.getFolders());
     }
-    
+
     public void testConvertNotCrashing() {
```
Nutomic: Please revert your changes to this file. Of course the whitespace and extra imports shouldn't be here, but removing that should go into a seperate cleanup commit/PR.

src/main/java/com/nutomic/syncthingandroid/syncthing/AppConfigReceiver.java
```diff
@@ -0,0 +1,73 @@
...
+            case ACTION_START:
+                // Start the Syncthing-Service
+
+                context.startService(new Intent(context, SyncthingService.class));
+                break;
+
+            case ACTION_STOP:
+                // Stop the Syncthing-Service
+                // If alwaysRunInBackground is enabled the service must not be stopped. Instead a
+                // notification is presented to the user.
```
Nutomic: I'd put this documentation on the ACTION_START/STOP definitions (as Javadoc).

sqrt-1764: Better, agreed

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/RestApiTest.java
sqrt-1764: Hmm, as expected, I had had conflicts upon rebase and I then accepted the version that came from you without any further editing.

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
```diff
@@ -426,12 +429,40 @@ public IBinder onBind(Intent intent) {
...
+     * happens the native binary can be shut down safely. PollWebGuiAvailableTaskImpl.onPostExecute
+     * already did this if mStopScheduled is true.
+     * A problem could arise if onDestroy is called when PollWebGuiAvailableTaskImpl.onPostExecute
+     * is just about to start the service. Then onDestroy sets mStopScheduled too late. and the
+     * binary is left running.
+     * So a lock is introduced:
+     * Either onDestroy is first. Then PollWebGuiAvailableTaskImpl.oonPostExecute waits for onDestroy
+     * to finish -> mStopScheduled is set and the binary is shut down then.
+     * Or PollWebGuiAvailableTaskImpl.onPostExecute is first. Then onDestroy waits and then kills
+     * the now active sycnthing binary on the "old" execution path.
```
Nutomic: Uh this is really long. Can you try to limit yourself to 2 - 3 sentences? Just describe what it does in which case, without describing so many concrete scenarios. 

And there's no need to write when this was added, that's what `git blame` is for. Just describe what the code does now. Also, there are some typos, maybe paste this into LibreOffice first?

sqrt-1764: Too short?

The native binary crashes if stopped before it is fully active. In that case signal the stop
request to PollWebGuiAvailableTaskImpl that is active in that situation and terminate the
service there.

sqrt-1764:
Pushed the updated version right now

src/androidTest/java/com/nutomic/syncthingandroid/test/syncthing/RestApiTest.java
sqrt-1764: Edit 3:
I just confirmed: Android Studio (at least in Version 2.0) automatically made those changes without notifying me. Simply viewing the file in the editor is enough

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
Nutomic: Nope that sounds good to me.

Nutomic:
Thanks, I finally merged it!

There was still the conflit with the submodule which I had to fix. That's why the commit might look weird on Github, and why Github doesn't recognize this PR as merged :/

sqrt-1764:
OK - as long as my code is merged ;-)

I just joined Transifex to add the translation for my string. But the server has errors and displas no data for me. But apparently someone already added the translation as the untranslated count went from 2 to 0 this afternoon ...

Nutomic:
Do you want to write some documentation on this? A wiki article would probably be good.

Also, `./gradlew lint` is complaining because the "Receiver does not require permission". [Documentation for receiver and permissions](https://developer.android.com/guide/topics/manifest/receiver-element.html). Can we add that, or will it break Tasker etc?

capi:
I suppose it'd break Tasker, since Tasker won't be holding the permission in question.

sqrt-1764:
Well, I made the code, so I also should provide the documentation. ;-)
I think a few sentences and maybe an example would be enough. People who want to use this feature should already know what an intent is ...
I also will publish my flow for automate in the automatie community once the feature is published ... so there will be a working example for the rest of the users.

Regarding the permission:
If we define a permission for the receiver, every app that wants to send those intents the receiver has to hold that permission. AFAIK (correct me if I'm wrong) that means that this permission has to be listed in the manifest of that app under "uses-permission" which will make the feature unusable for Tasker etc.
It is intended as a public service - and therefore must not be protected by a permission, so I think that lint-message has to be suppressed.

Nutomic:
Alright, I'll disable the warning then. 

Oh and it would be cool if you could do the integration of AppConfigReceiver and DeviceStateHolder when you get to it :)

sqrt-1764:
You mean writing this wiki-article?

Nutomic:
No I mean in the code, with the `isSyncthingEnabled()` method we talked about.

sqrt-1764:
Ah - will look into this soon.

sqrt-1764:
I just created an article in the wiki for this.
I also tested the behavior of Syncthing in case "always run in background" is enabled. In this case for me the intents are ignored. Will say the configured behavior took precedence.
This seems to be the behavior you wanted. So is this isSynctingEnabled() method currently needed? If yes - where shall we discuss the details? I think a new thread would be appropriate then ...

Nutomic:
IRC would be good, are you on there? #syncthing on Freenode.

PR #600, opened 2016-03-27
Patch Go's DNS

AudriusButkevicius:
Please test this before merging.

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
@@ -126,7 +126,7 @@ public class SyncthingRunnable implements Runnable {
                 mSyncthing.set(process);
 
                 mErrorLog = "";
-                Thread lInfo = log(process.getInputStream(), Log.INFO, false);
+                Thread lInfo = log(process.getInputStream(), Log.INFO, true);
                 Thread lWarn = log(process.getErrorStream(), Log.WARN, true);
 
                 niceSyncthing();


make-go.bash
```diff
@@ -44,6 +25,29 @@ case "$1" in
...
+            pushd ext/golang/go1.4/src
+            ./make.bash --no-clean
+            popd
+            # Add Go 1.4 to the environment
+            export GOROOT="$(pwd)/ext/golang/go1.4"
+    fi
+    # Add Go 1.4 compiler to PATH
+    export GOROOT_BOOTSTRAP=$GOROOT
+fi
+
```
Nutomic: These changes here break my build. And I suppose it might also break the build for F-Droid. What were you trying to do here?

AudriusButkevicius: I don't think thats what broke it. I was just validating the arguments before bootstraping, though now thinking about it that might not be right.

Nutomic: It's probably because I haven't set up Go correctly on my laptop. But I'm too lazy for that right now, and as I said, it might break the F-Droid built. If you wanna change this, put it in a seperate PR please ;)

make-go.bash
```diff
@@ -60,6 +64,11 @@ mkdir -p "$GOROOT_FINAL"
 
 pushd ext/golang/go/src
 
+for PATCH in ../../../../patches/golang/all/*.patch; do
+    echo "Applying $PATCH"
+    git apply $PATCH
+done
```
Nutomic: The patch isn't applied, not quite sure why. Maybe it's because the file is in a different git repo. It might be better to use `patch`.

patches/golang/all/dns.patch
```diff
@@ -0,0 +1,82 @@
+diff --git a/net/dnsconfig_unix.go b/net/dnsconfig_unix.go
```
Nutomic: I would just store this file as `patches/golang_dns.patch`, no need for the long path.

AudriusButkevicius: I'd argue that we'd leave it as it is, as it allows us patches per arch.

Nutomic:
Why did you close this?

AudriusButkevicius:
Because I am smashed and I don't know what I am doing.

Nutomic:
Hahaha 😂

Issue #593, opened 2016-03-24
v. 0.7.12 update - disconnected


tguser:
After latest update on Android, v0.7.12, my 2 android clients (stock Marshmallow) cannot find other nodes.
Global discovery is 0/6.
Syncthing log is empty, there is only:
--------- beginning of main

kvj:
Same for me:

Android: 0.7.12
Binary: 0.12.21

Stopped connecting to server both 0.12.20 and 0.12.21 (upgraded, for testing)

There is one possibly relevant message in log output:
W SyncthingNativeCode: ionice: exec 12573: Permission denied

kvj:
Well, in my case there is a name resolving issue. Host names are not resolved anymore. Switching to ip address resolved it. 

Probably because of fce35f15016ae315fa6a68bb0dc078efaea147e5 ?

AudriusButkevicius:
Yeah, but CGO was crashing for a lot of other people.

kvj:
"What's new" section on Play Store has no mention about this degradation... :(

Global Discovery also stopped working, right?

tguser:
As for now my client is useless, I can't sync data because of global discovery with 0 server available. 
Will we receive a new update? (Maybe a rollback of some code?)
TY

Nutomic:
What devices are you using? I tested DNS on my LG G4running Marshmallow and it worked (I think @Zillode also tested it).

tguser:
I have a Nexus 5x, stock android with march update.
It worked until latest syncthing update, after that It stopped to find global discovery servers.
I tried to reinstall but it didn't worked.
Same behaviour in wifi and mobile data.
How do I test DNS?

Nutomic:
Okay to explain this issue a bit further:

Until the last version (0.7.10), we used cgo to compile Syncthing, because it supports DNS on Android. However, many people were getting crashes with that version (#518 and other issues). To help them, we switched back to the standard Go compiler now. As I said, DNS also works with that compiler on my device, but apparently it fails on some devices. To see if DNS works, check the number of announce servers (should be at least 5/8), or try to connect to a device by address.

Of course we could switch back to the cgo compiler, but I think it's better to have failing DNS for some people, than to have instant crashes for others. However I don't know how many people are affected by each of these issues.

[...]

PR #601, opened 2016-03-27
Patch GO DNS

AudriusButkevicius:


AudriusButkevicius:
I hate git especially when I am drunk and I cannot explain it.

make-go.bash
```diff
@@ -60,6 +60,11 @@ mkdir -p "$GOROOT_FINAL"
 
 pushd ext/golang/go/src
 
+for PATCH in ../../../../patches/golang/all/*.patch; do
+    echo "Applying $PATCH"
+    patch -p1 <$PATCH
```
Nutomic: Running `./gradlew cleanNative buildNative` gave me [this error](https://gist.github.com/Nutomic/b92f0b3c29fa87c619bdfa63ba5c4f1d), because `make-go.bash` is called twice (for each arch). You'll have to use `patch ... || true` to ignore additional calls (and please add a short comment explaining this).

AudriusButkevicius: Patch should be a noop if its already applied, atleast on my version of patch.

AudriusButkevicius:
There, added a check.

Nutomic:
Ah that's even better. I think it would be better to use `$MYDIR` instead of the `../` stuff though. And please add "fixes #593" to the commit message.

@Zillode Any comments?

AudriusButkevicius:
Using $MYDIR

Zillode:
:+1:

AudriusButkevicius:
This actually seems to return the default google servers all the time.

PR #610, opened 2016-04-02
Really fix DNS

AudriusButkevicius:
The previous patch was merged against a different version of Go, plus had bugs.
Now given I am able to build stuff, this seems to do the right job.

You should look at the raw patch rather than the diff to the diff.

### DIFF
--- a/make-go.bash
+++ b/make-go.bash
@@ -60,14 +60,12 @@ mkdir -p "$GOROOT_FINAL"
 
 pushd ext/golang/go/src
 
+git reset --hard HEAD
+
 # Apply patches to Golang
 for PATCH in $MYDIR/patches/golang/all/*.patch; do
-    if ! patch -R -p1 --dry-run <$PATCH &>/dev/null; then
-        echo "Applying $PATCH"
-        patch -p1 <$PATCH
-    else
-        echo "Patch $PATCH already applied"
-    fi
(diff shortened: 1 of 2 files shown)


Nutomic:
I suppose this won't work if the previous version of the patch was applied?

@Zillode Does Jenkins discard changes in the submodule after each build? Or do we have to do something to get rid of the old patch?

AudriusButkevicius:
atleast I had to do that

Nutomic:
Okay I just wiped the workspace from Jenkins (didn't know you can do that). I hope that's enough.

AudriusButkevicius:
Your "Oh this branch builds fine" builds master anyway, so that doesn't matter much :D

Nutomic:
I'm not sure what you mean. We have to make sure the patch goes into new release builds, right?

AudriusButkevicius:
What I mean is if you look at:

http://android.syncthing.net/job/Syncthing-Android-PR/197/console

In every build you will see:

`Checking out Revision 09bba7140aa3d811edf1c6322e24731a1978b06b (origin/master)`

Nutomic:
Right, I guess it's fine. I just wanted to be sure, and I really don't understand Jenkins.

Zillode:
`Delete workspace before build starts` is now enabled

PR #613, opened 2016-04-10
fix broken bug report link

lkwg82:
the link to open an issue points to old github.com/nutomic/syncthing-android/issues


Nutomic:
Thanks!

Issue #620, opened 2016-04-18
Crash opening WiFi whitelist


scienmind:
After upgrading to Android 6, selecting "Restrict to certain WiFi networks" crashes the app:

1. Open the SyncThing Android application
2. Open application's settings
3. Enable "Background syncing"
4. Enable "WiFi Only" syncing
5. Click "Restrict to certain WiFi networks"
-Application Crushes-

I tried deleting ST's data, and re-importing the config. Still same crush...

ST v.0.12.20
Android wrapper v.0.7.11 (latest f-droid release...)
OS: CM13 latest stable snapshot

Log:
```
--------- beginning of main
04-18 16:39:50.645 21384 21404 I chatty : uid=10139(com.nutomic.syncthingandroid) expire 14 lines
04-18 16:40:06.121 21384 21444 I chatty : uid=10139(com.nutomic.syncthingandroid) expire 11 lines
04-18 16:40:22.095 21384 21384 I chatty : uid=10139(com.nutomic.syncthingandroid) expire 10 lines
04-18 16:43:25.296 21384 21444 I chatty : uid=10139(com.nutomic.syncthingandroid) expire 10 lines
04-18 16:44:22.105 21384 21404 I chatty : uid=10139(com.nutomic.syncthingandroid) expire 6 lines
04-18 16:44:56.138 21384 21409 I chatty : uid=10139(com.nutomic.syncthingandroid) expire 3 lines
04-18 16:45:34.165 21384 21394 I chatty : uid=10139(com.nutomic.syncthingandroid) expire 1 line
04-18 16:45:34.327 21384 21403 I chatty : uid=10139(com.nutomic.syncthingandroid) expire 1 line
04-18 16:46:27.026 22656 22656 I chatty : uid=10139(com.nutomic.syncthingandroid) expire 10 lines
04-18 16:46:27.378 22656 22672 I chatty : uid=10139(com.nutomic.syncthingandroid) expire 2 lines
04-18 16:46:27.531 22656 22673 I chatty : uid=10139(com.nutomic.syncthingandroid) expire 1 line
04-18 16:46:27.548 22676 22676 I chatty : uid=10139(com.nutomic.syncthingandroid) chmod expire 2 lines
04-18 16:46:27.957 22656 22679 I chatty : uid=10139(com.nutomic.syncthingandroid) expire 10 lines
04-18 16:46:28.709 22691 22691 I chatty : uid=10139(com.nutomic.syncthingandroid) ps expire 14 lines
[...]

PR #621, opened 2016-04-19
Add checks for null-SSID (fix for #620)

capi:
This is a simple safety check for null when accessing and displaying SSIDs in `WifiSsidPreference`. I still think that there is something going on inside CM regarding maybe missing permissions, because I cannot think of a valid case where a stored WiFi network has no SSID.

The check in `SyncthingService` is already null-safe.

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
+++ b/src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
@@ -91,11 +91,13 @@ public class WifiSsidPreference extends MultiSelectListPreference {
     private CharSequence[] extractSsid(WifiConfiguration[] configs, boolean stripQuotes) {
         CharSequence[] result = new CharSequence[configs.length];
         for (int i = 0; i < configs.length; i++) {
+            // See #620: there may be null-SSIDs
+            String ssid = configs[i].SSID != null ? configs[i].SSID : "";
             // WiFi SSIDs can either be UTF-8 (encapsulated in '"') or hex-strings
             if (stripQuotes)
-                result[i] = configs[i].SSID.replaceFirst("^\"", "").replaceFirst("\"$", "");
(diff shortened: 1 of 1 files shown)


src/main/java/com/nutomic/syncthingandroid/preferences/WifiSsidPreference.java
```diff
@@ -115,7 +117,10 @@ private void filterRemovedNetworks(Set<String> selected, CharSequence[] all) {
                 Arrays.sort(result, new Comparator<WifiConfiguration>() {
                     @Override
                     public int compare(WifiConfiguration lhs, WifiConfiguration rhs) {
-                        return lhs.SSID.compareToIgnoreCase(rhs.SSID);
+                        // See #620: There may be null-SSIDs
+                        String l = lhs != null && lhs.SSID != null ? lhs.SSID : "";
+                        String r = rhs != null && rhs.SSID != null ? rhs.SSID : "";
+                        return l.compareToIgnoreCase(r);
```
Nutomic: Is there any reason to check if the config is null? The issue (and your other fix) is only about the SSID being null, so I would only handle that. 

It's pretty annoying that Java doesn't have a standard method for this, but you could use [this code](http://stackoverflow.com/a/11271611) to shorten it (which also looks rather complicated).

capi: Just didn't want to get caught by the next thing being null which never should :) But yeah, can do that.

capi:
Updated check as per your request. I leave it in a separate assignment, since I find it easier to read than the other proposed version (but that's a matter of taste). Any decent JVM will optimize both to the same code anyway.

Nutomic:
Yeah agreed about the readability, this is probably the best way to write it. 

And it looks like the code would break in other places too if the config was null, so let's hope that won't happen :D

PR #630, opened 2016-04-26
Track WiFi changes again in NetworkReceiver (fixes #625)

### LINKED ISSUE #625: Sync on Wifi not working correctly
0.12.22 and when the option Always Run In Background and WiFi is chosen, it does will not start.

CM 13
Nexus 5

capi:


Nutomic:
Thanks for finding and fixing this!

I'll make a new release in a bit.

PR #631, opened 2016-04-26
[WIP] Show Syncthing activity in notification

capi:
This is work-in-progress, please **DO NOT MERGE**.

This attempts to show some useful activity information in the notification. It is based on tracking the events from the native binary to avoid having to poll for changes all the time.

It shows (in order of preferences):
* Folders that are downloading
* Devices which are uploading
* Folders which are scanning
* Number of devices that are currently connected

The following things I'd like to implement before this is "done":

- [x] Track basic events from API model and update notification
- [ ] Handle "out-of-sync" case (if Syncthing's event buffer overflows)
- [ ] Better progress reporting
- [ ] I18N
- [ ] Squash the individual commits


src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
```diff
@@ -415,6 +492,13 @@ public void onConfigChanged() {
             });
 
             mEventProcessor = new EventProcessor(SyncthingService.this, mApi);
+            mEventProcessor.addOnEventBasedModelChangedListener(new EventBasedModel.OnEventBasedModelChangedListener() {
+                @Override
+                public void onEventBasedModelChanged(EventBasedModel model) {
+                    // TODO: needs to run on main thread!
```
Nutomic: You can create a `Handler` on the main thread, then do `handler.post()`. Or maybe there's already a handler in this class.

src/main/java/com/nutomic/syncthingandroid/syncthing/EventBasedModel.java
```diff
@@ -0,0 +1,272 @@
...
+            ByteArrayOutputStream bos = new ByteArrayOutputStream();
+            ObjectOutputStream oos = new ObjectOutputStream(bos);
+            oos.writeObject(this);
+            oos.close();
+
+            ByteArrayInputStream bin = new ByteArrayInputStream(bos.toByteArray());
+            ObjectInputStream oin = new ObjectInputStream(bin);
+            return new Snapshot((EventBasedModel) oin.readObject());
+        } catch (Exception e) {
+            throw new IllegalStateException("Failed to deep-copy EventBasedModel: " + e.getMessage(), e);
```
Nutomic: Why do you put the exception twice? :p

Nutomic:
The code is so verbose, would be great to have Scala, or Java 8 D:

Any idea how the notification should look? Seems like it might be a lot of info. You should probably go for an extended notification to hold the names of folders/devices. The normal notification could show number of devices and status (syncing/idle/whatever).

capi:
The notification is currently most of the time truncated on my device, but it serves it's purpose. You clearly see if something's going on or if the device is idle. For me this works and I'm not sure we could get it better with an extended notification. It even works fine on the lock-screen when just simply turning on the device shortly to see if everything is already synced (it is, if you see `n devices connected`).

But I'm open to that :)

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
capi: I think this is still something that I need to clean up, because it used to run on a separate thread and I'm quite sure it is already called in a `Handler` in the meantime.

src/main/java/com/nutomic/syncthingandroid/syncthing/EventBasedModel.java
capi: Habit from work, where we tend to have a shorter log for tailing on a console (`info.log`) with just having the `Log Message: TheExceptionMessage` and an additional `verbose.log` including the full stack-trace. Gonna change :)

Nutomic:
Yeah it was just an idea, because it's not in your code yet. Mind sharing a screenshot? ;)

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
```diff
@@ -299,6 +305,7 @@ private void updateNotification() {
                     .setContentTitle(getString(R.string.syncthing_active))
                     .setSmallIcon(R.drawable.ic_stat_notify)
                     .setOngoing(true)
+                    .setContentText(formatOngoingSyncInfo())
```
capi: @Nutomic: Here it is.

capi:
It is in the first of the 4 commits (a0607c41ec082ff58a8d7bb8e13f4ce13bd9f7ae). Screenshots attached below.

screenshot1
screenshot2

Nutomic:
Ah I was looking for a bigger change, not just one line :P 

Yeah it looks pretty cramped with the folders like that. Connected devices seems fine, and/or overall status. The other stuff could then be expanded.

capi:
It's not currently working with current master, since I had no time to keep it up-to-date. I'll close the PR for now and will open a new one once I updated the feature to current master.

PR #638, opened 2016-04-30
Update Google Play and F-Droid badges

relan:
Use the latest versions of badges:

Get it on Google Play Get it on F-Droid


Nutomic:
Thanks!

PR #656, opened 2016-05-31
German translation updated.

twckr:


wweich:
As stated in the readme (front page), translations are done in [transifex](https://www.transifex.com/projects/p/syncthing-android/).

lkwg82:
So this can be closed, as 'will not be merged, plz use transifex' ?

PR #662, opened 2016-06-07
Syncthing version

lkwg82:
This adds the current syncthing version to the slider. No more open the web gui to check the version.

screenshot1

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,6 +24,7 @@ dependencies {
     compile 'eu.chainfire:libsuperuser:1.0.0.201602271131'
     compile 'com.android.support:design:23.3.0'
     compile 'com.google.zxing:android-integration:3.2.1'
+    compile 'com.google.code.gson:gson:2.6.2'
     androidTestCompile 'com.squareup.okhttp:mockwebserver:2.4.0'
 }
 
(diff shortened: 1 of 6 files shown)


capi:
First this gets a :+1: from me.

Second, I'd squash the two commits into one.

lkwg82:
@capi thx and what was this github feature to squash commits?

capi:
@lkwg82 It's a git feature. Use `git rebase -i` and mark them all as "fixup". Then make a force-push to this branch and it will update the Pull Request.

lkwg82:
@capi I meant the github ui feature to squash commits in UI, anyway I did it

lkwg82:
So any hints or may I merge this?

AudriusButkevicius:
Wait for @Nutomic

Nutomic:
Using a JSON library is definitely a good idea, seeing how much boilerplate code we have there.

I'm not sure if it's a good idea to put the version into the drawer. On larger devices that's fine, but on smaller devices it means that the share and restart buttons can be pushed out (so you have to scroll to them). Any opinions on that?

Other than that, it looks good to me.

lkwg82:
mhm, I just checked the web gui, what are the most important informations. But you are right. I dont know.

Nutomic:
Alternatively, we could change the scrollview so that it only contains the info view on top, and not the buttons. That way, people with a smaller screen wouldn't miss so much.

lkwg82:
What do you mean?

Nutomic:
You can test this by rotating your device into landscape: When opening the drawer, and it doesn't all fit on the screen, you can scroll the info on top, together with the web gui/share/restart buttons. We could change it so that only the info view scroll, and the buttons are static.

PR #669, opened 2016-06-09
fixes RestApi options for 'listenAddress'

lkwg82:
this fixes display of syncthing listenAddress(es) in options menu.


Nutomic:
Thanks! Looks like no one even noticed that...

lkwg82:
There is "upnpEnabled" left, but I dont know how the api changed. ;)

AudriusButkevicius:
It didn't at all. It just does more than before.

Zillode:
I think it was renamed to `natEnabled`?

PR #674, opened 2016-06-12
fix NPE, when pressing back button on LogActivity (fixes 604)

### LINKED ISSUE #604: Crash in Log press back arrow to go back to Settings
This was in the Log->Android Log, not sure why or when:
```
E/AndroidRuntime(17900): FATAL EXCEPTION: main
E/AndroidRuntime(17900): Process: com.nutomic.syncthingandroid, PID: 17900
E/AndroidRuntime(17900): java.lang.RuntimeException: Unable to start activity ComponentInfo{com.nutomic.syncthingandroid/com.nutomic.syncthingandroid.activities.SettingsActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.String.hashCode()' on a null object reference
E/AndroidRuntime(17900):        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2379)
E/AndroidRuntime(17900):        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2442)
E/AndroidRuntime(17900):        at android.app.ActivityThread.access$800(ActivityThread.java:156)
E/AndroidRuntime(17900):        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1351)
E/AndroidRuntime(17900):        at android.os.Handler.dispatchMessage(Handler.java:102)
E/AndroidRuntime(17900):        at android.os.Looper.loop(Looper.java:211)
E/AndroidRuntime(17900):        at android.app.ActivityThread.main(ActivityThread.java:5373)
E/AndroidRuntime(17900):        at java.lang.reflect.Method.invoke(Native Method)
E/AndroidRuntime(17900):        at java.lang.reflect.Method.invoke(Method.java:372)
E/AndroidRuntime(17900):        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1020)
E/AndroidRuntime(17900):        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:815)
E/AndroidRuntime(17900): Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'int java.lang.String.hashCode()' on a null object reference
E/AndroidRuntime(17900):        at com.nutomic.syncthingandroid.activities.SettingsActivity.onCreate(SettingsActivity.java:41)
E/AndroidRuntime(17900):        at android.app.Activity.performCreate(Activity.java:5990)
E/AndroidRuntime(17900):        at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1106)
E/AndroidRuntime(17900):        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2332)
E/AndroidRuntime(17900):        ... 10 more
```

lkwg82:
This should fix #604.

I think this is valid, as the back button gives no action just wants back.
But I'm no android expert.

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
+++ b/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
@@ -44,6 +44,7 @@ public class SettingsActivity extends SyncthingActivity {
                     savedInstanceState.getString("fragment_name"));
         } else {
             String action = getIntent().getAction();
+
             if (action == null) {
                 action = ACTION_APP_SETTINGS;
             }
--- a/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
+++ b/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
@@ -2,6 +2,7 @@ package com.nutomic.syncthingandroid.activities;
 
 import android.app.Fragment;
 import android.app.FragmentManager;
+import android.content.Intent;
 import android.os.Bundle;
 
 import com.nutomic.syncthingandroid.R;
@@ -42,7 +43,12 @@ public class SettingsActivity extends SyncthingActivity {
             mFragment = fm.getFragment(savedInstanceState,
                     savedInstanceState.getString("fragment_name"));
         } else {
-            switch (getIntent().getAction()) {
+            String action = getIntent().getAction();
+            if (action == null) {
+                action = ACTION_APP_SETTINGS;
+            }
+
+            switch (action) {
                 case ACTION_APP_SETTINGS:
                     setTitle(R.string.settings_title);
                     mFragment = new SettingsFragment();


Nutomic:
If you wanna go this route, you can just use the default to open settings (and have a fallthrough in `ACTION_APP_SETTINGS`).

The thing is that this crash isn't actually supposed to happen, because we always pass that parameter. But I guess Android can kill the activity when it runs out of memory, and then it doesn't remember the intent correctly.

lkwg82:
> The thing is that this crash isn't actually supposed to happen, because we always pass that parameter.

Where do we pass it on in that specifica case?

AudriusButkevicius:
So I am sure it's nothing todo with memory because it crashes for me 100% of the time immdiately after starting the app.

lkwg82:
@Nutomic How do I do that?

Nutomic:
@lkwg82 We pass it whenever we send a start intent for SettingsActivity (you can find that with right click and "find usages" in Android Studio). This extra should also be there when you're coming back from a different activity I think. But maybe that isn't true on all Android versions.

What do you mean with "How do I do that"?

lkwg82:
What do you mean by fallthrough?

Nutomic:
Ah, I thought we could just handle null in the default case, then we could do a fallthrough from ACTION_APP_SETTINGS to default. But unfortunately Java doesn't allow that.

Can you add "(fixes #604)" to the commit message so Github closes it? Otherwise it looks good to me.

lkwg82:
I did.

Nutomic:
Did you push that? It doesn't show in Github :p

lkwg82:
*scratch ... moment ;)

Nutomic:
Github shows the commit message "fix NPE, when pressing back button on LogActivity". I mean you should append "(fixes #604)" to that, so Github closes that issue automatically. Like [this issue](https://github.com/syncthing/syncthing-android/commit/b4e2c44ab20d6085d9fa42d94fbf5af342348872) for example.

lkwg82:
ah, not in the title of that issue .. you mean as commit message. Will do later at home.

Nutomic:
Now you have 3 commits with different commit messages, and you're missing the `#` in front of the issue number (should be `#604`). I'm not sure if the Github merger handles this correctly, so can you please squash the commits? Let me know if you need help with that.

Nutomic:
Now you had 3 commits with different commit messages. You could have just used `git commit --amend` to change the previous commit. Anyway I squashed it via Github so that's good. Thanks for the fix!

PR #675, opened 2016-06-12
Use folderLabel from FolderRejected events

Unrud:
Show folder label in notification and preset value for new folder.


Nutomic:
Thanks!

PR #680, opened 2016-06-21
update syncthing version to v0.13.7

lkwg82:


Nutomic:
I don't really see the point of doing PRs for Syncthing updates. Updating the submodule only takes a few seconds, and until we automate this, I have to check it manually anyway. 

So I'm gonna close this for now.

PR #681, opened 2016-06-21
fix api access to activate/deactive natEnabled

lkwg82:
this fixes the changed api access to internal syncthing

How can the translation changed?

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -274,7 +274,7 @@ Please report any problems you encounter via Github.</string>
 
     <string name="local_announce_enabled">Local Discovery</string>
 
-    <string name="upnp_enabled">Enable UPnP</string>
+    <string name="nat_enabled">Enable NAT</string>
 
     <string name="global_announce_server">Global Discovery Server</string>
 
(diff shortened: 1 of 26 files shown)


Nutomic:
Change the english string in `res/values/strings.xml`, I will then push it to Transifex. You should probably change the string's key for consistency, and delete the string from all other language folders.

lkwg82:
Plz. delete branches on merge, thx!

src/main/res/values/strings.xml
```diff
@@ -274,7 +274,7 @@ Please report any problems you encounter via Github.</string>
 
     <string name="local_announce_enabled">Local Discovery</string>
 
-    <string name="upnp_enabled">Enable UPnP</string>
+    <string name="nat_enabled">Enable NAT</string>
```
AudriusButkevicius: What does Enable NAT mean? 
NAT traversal maybe?

Nutomic: The web gui says "Enable NAT traversal", so I think we should go with that.

lkwg82: Of course, I missed it.

Nutomic:
Thanks!

PR #683, opened 2016-06-21
upgrades latest gradle and build tools

lkwg82:
I double checked the DeviceFragment Crash in the Emulator and on my cm13 android device. It is gone!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,7 +45,10 @@ android {
         testHandleProfiling true
         testFunctionalTest true
     }
-
+    // Without this parameter, all builds fail with error:
+    //     Execution failed for task ':mergeArmeabiDebugResources'.
+    //     > Some file crunching failed, see logs for details
+    // https://code.google.com/p/android/issues/detail?id=163378
     aaptOptions.setProperty("cruncherEnabled", false)
 
     // We haven't upgraded to the new HTTP library yet.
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -46,15 +46,7 @@ android {
...
-    // Without this parameter, all builds fail with error:
-    //     Execution failed for task ':mergeArmeabiDebugResources'.
-    //     > Some file crunching failed, see logs for details
-    // https://code.google.com/p/android/issues/detail?id=163378
-    //
-    // NOTE: Setting this causes a crash when opening the DeviceFragment:
-    //       Resources$NotFoundException: File res/drawable/ic_tag_text_outline_black_24dp.xml
-    //       This means we can't use build tools 2.0 until this is fixed.
-    //aaptOptions.setProperty("cruncherEnabled", false)
+    aaptOptions.setProperty("cruncherEnabled", false)
```
Nutomic: Did you check if this is still needed? If yes, you should leave that part of the comment.

gradle.properties
```diff
@@ -1 +1,2 @@
 android.useDeprecatedNdk=true
+org.gradle.jvmargs=-Xmx2048M
```
Nutomic: Why are you setting this?

lkwg82: this follows a warning by the gradle run ... see the console output without it.

build.gradle
lkwg82: Still needed and yes comments should be left in place.

PR #689, opened 2016-06-27
test fix jenkins

Nutomic:


AudriusButkevicius:
Why?

Nutomic:
[Because of this](http://android.syncthing.net/view/Syncthing-Android/job/Syncthing-Android-Release/101/console). 

For some reason, it fails to apply your patch. @calmh suggested it's because the script uses master instead of the submodule commit (golang 1.6.2). The weird thing is that it works fine for me locally.

AudriusButkevicius:
Also, dunno if its related but jenkins seems to check out master rather than any branch. I told you about this before I think. Not sure if that is the case.

The reset will break subsequent builds tho

Nutomic:
After this change, [Jenkins fails to clone the submodules](http://android.syncthing.net/view/Syncthing-Android/job/Syncthing-Android-PR/235/console). I really don't know what's going on here, or how to fix it...

You mean Jenkins builds the master branch when it's supposed to build PRs? That would be pretty bad. As I said before, I really don't understand Jenkins, so any help to fix it would be great :)

lkwg82:
This become kind of urgent with new 0.14 release.

lkwg82:
I'll try to dig into it today. Later.

Nutomic:
I can do releases just fine without jenkins, so no hurry. But a working build server would definitely be great :)

lkwg82:
ah, ok.

PR #692, opened 2016-07-03
Sort folders by label

xduugu:
If the label is empty or unset, use the folder ID in comparision.

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/fragments/FolderListFragment.java
+++ b/src/main/java/com/nutomic/syncthingandroid/fragments/FolderListFragment.java
@@ -29,10 +29,16 @@ import java.util.TimerTask;
 public class FolderListFragment extends ListFragment implements SyncthingService.OnApiChangeListener,
         AdapterView.OnItemClickListener, AdapterView.OnItemLongClickListener {
 
+    /**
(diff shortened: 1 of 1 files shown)


lkwg82:
Plz add test(s).

Nutomic:
I would prefer if you put the comment as Javadoc before the class definition. Tests would be nice, but I wouldn't consider them mandatory (seeing as we don't have any tests there yet).

xduugu:
Moved comment to Javadoc.
About the test, I'm not sure how to write one. So I might not be able to provide one in the next days.

lkwg82:
> Tests would be nice, but I wouldn't consider them mandatory (seeing as we don't have any tests there yet).

Lets start with some trivial unit tests. If we dont start at some point in time it will never happens. They should be really trivial ones.

Nutomic:
There are already some tests which you can check out. To write a test for this class, you'll have to make it public (which I don't really like tbh). Then you can add a new class `FolderListFragmentTest` and add the test there. Basically, just pass in an unsorted list of folders and assert that it comes out sorted.

Nutomic:
Merged, because I don't think tests are essential here (and we didn't have a test for this anyway).

Thanks for your contribution :)

Issue #634, opened 2016-04-28
 Sync on Wifi not working correctly #625  not fixed really


gerroon:
#625  fix still does not work for me in 0.7.16.

It works the first time but if I start it again (because it crashed) when the wifi off the next time it still brings that "do you want to change your preferences' window, which makes no sense whatsoever 

This is an Android 4.2 device.

wweich:
>when the wifi off

If you set to only sync on Wifi and you are not connected to Wifi, then this is by design.

gerroon:
@wweich 

I actually do not think so. If this is by design then I as a user find it to be a weird design choice.

Here is why it is weird.

I enable Wifi
I open Syncthing
I enable wifi only
Let it run
I restart my device for some reason
I enable wifi (I have to because ST will refuse to run as we know)
I start ST
Sometime later my battery is getting low so I disable WIfi.
I open ST to see what is going ton
ST refuses to let me look around because there is no WIFI

AudriusButkevicius:
It's still by design. It kills the syncthing process which does expensive things even if not connected to anyone.

gerroon:
@AudriusButkevicius 

I do not mind if it is killing the process. What does this have to do with me not having wifi enabled at the time of clicking on the interface? 

Can't it still open the interface instead of blocking?

AudriusButkevicius:
No, because the interface is driven by the process that is killed.

gerroon:
So I have to have Wifi enabled at all times if I want to open the interface properly? 

How do I know things are working with ST when the wifi is off? The thing is that, how would I know if ST binary process is running or not when the wifi is off? ;)  I cant test out the wifi off case because as we know I am not allowed to.

The issue is this, I am using an Ereader and things like Wifi is abit expensive on these devices, that is why I am trying to find the best use case.

AudriusButkevicius:
Things are not working when wifi is off, so there is nothing to check. The binary ia not runnig when Wifi is off, thats what the wrapper ensures. 

I don't know/understand why you are not allowed to try wifi only case, all you need to do is connect to wifi.

gerroon:
@AudriusButkevicius 

Ok you are saying thatt nothing is working. So I tested this one my CM12

[...]

PR #694, opened 2016-07-11
Experimental "Run as foreground service" setting

capi:
This PR adds a setting to have the service running with `startForeground()` as a setting in the "Experimental" section.

On my OnePlus 2, the background service gets killed after 5-10 minutes of high activity (i.e. when transferring many/large files) due to a custom addition to `ActivityManager` called `BgDetect`.

    07-06 20:52:26.562  1181  1363 I ActivityManager: [BgDetect]chkExcessCpu doKills: true uptime: 300309
    07-06 20:52:26.960  1181  1363 I ActivityManager: [BgDetect]detect excessive cpu on forked process libsyncthing.so(pid : 2758) plan to stop it 371060 during 300309
    07-06 20:52:26.960  1181  1363 I ActivityManager: [BgDetect]force stop com.nutomic.syncthingandroid.debug (uid 10132)
    07-06 20:52:26.965  1181  1363 I ActivityManager: Force stopping com.nutomic.syncthingandroid.debug appid=10132 user=0: from pid 1181
    07-06 20:52:26.966  1181  1363 I ActivityManager: Killing 1360:com.nutomic.syncthingandroid.debug/u0a132 (adj 15): stop com.nutomic.syncthingandroid.debug
    07-06 20:52:26.968  1181  1363 W ActivityManager: Scheduling restart of crashed service com.nutomic.syncthingandroid.debug/com.nutomic.syncthingandroid.syncthing.SyncthingService in 1000ms
    07-06 20:52:26.968  1181  1354 V ActivityManager: killProcessGroupAsync took 1 ms for PID 1360 on thread 14
    07-06 20:52:26.971  1181  6410 I libprocessgroup: Killing pid 2758 in uid 10132 as part of process group 1360
    07-06 20:52:26.973  1181  6410 I libprocessgroup: Killing pid 2758 in uid 10132 as part of process group 1360
    07-06 20:52:26.974  1181  1363 I ActivityManager:   Force stopping service ServiceRecord{47304c8 u0 com.nutomic.syncthingandroid.debug/com.nutomic.syncthingandroid.syncthing.SyncthingService}

The only workaround for now is to have it running with foreground priority. This proved to be stable in my experiments and solved the issue on my ROM. Similar issues might affect other ROMs and could be part of the reason for e.g. #634 and similar issues caused by not-running / crashed services.

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
@@ -97,6 +97,7 @@ public class SyncthingService extends Service implements
     public static final String PREF_USE_ROOT                 = "use_root";
     private static final String PREF_NOTIFICATION_TYPE       = "notification_type";
     public static final String PREF_USE_WAKE_LOCK            = "wakelock_while_binary_running";
+    public static final String PREF_FOREGROUND_SERVICE       = "run_as_foreground_service";
 
     private static final int NOTIFICATION_ACTIVE = 1;
 
@@ -291,6 +292,13 @@ public class SyncthingService extends Service implements
     private void updateNotification() {
         SharedPreferences sp = PreferenceManager.getDefaultSharedPreferences(this);
         String type = sp.getString(PREF_NOTIFICATION_TYPE, "low_priority");
+        boolean foreground = sp.getBoolean(PREF_FOREGROUND_SERVICE, false);
+        if ("none".equals(type) && foreground) {
+            // foreground priority requires any notification
+            // so this ensures that we either have a "default" or "low_priority" notification,
+            // but not "none".
+            type = "low_priority";
+        }
         NotificationManager nm = (NotificationManager) getSystemService(NOTIFICATION_SERVICE);
         if ((mCurrentState == State.ACTIVE || mCurrentState == State.STARTING) &&
                 !type.equals("none")) {
@@ -304,15 +312,23 @@ public class SyncthingService extends Service implements
             if (type.equals("low_priority"))
                 builder.setPriority(NotificationCompat.PRIORITY_MIN);
 
-            nm.notify(NOTIFICATION_ACTIVE, builder.build());
+            if (foreground) {
+                builder.setContentText(getString(R.string.syncthing_active_foreground));
+                startForeground(NOTIFICATION_ACTIVE, builder.build());
+            } else {
+                stopForeground(false); // ensure no longer running with foreground priority
+                nm.notify(NOTIFICATION_ACTIVE, builder.build());
+            }
         } else {
+            // ensure no longer running with foreground priority
+            stopForeground(false);
             nm.cancel(NOTIFICATION_ACTIVE);
         }
     }
 
     @Override
     public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, String key) {
-        if (key.equals(PREF_NOTIFICATION_TYPE))
+        if (key.equals(PREF_NOTIFICATION_TYPE) || key.equals(PREF_FOREGROUND_SERVICE))
             updateNotification();
         else if (key.equals(PREF_SYNC_ONLY_CHARGING) || key.equals(PREF_SYNC_ONLY_WIFI)
                 || key.equals(PREF_SYNC_ONLY_WIFI_SSIDS))
@@ -464,6 +480,7 @@ public class SyncthingService extends Service implements
             mApi.shutdown();
 
         NotificationManager nm = (NotificationManager) getSystemService(NOTIFICATION_SERVICE);
+        stopForeground(false);
         nm.cancel(NOTIFICATION_ACTIVE);
 
         for (FolderObserver ro : mObservers) {
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -298,6 +298,10 @@ Please report any problems you encounter via Github.</string>
 
     <string name="keep_wakelock_while_binary_running_summary">Use this setting if you experience unexpected disconnects while operating on battery. This will result in increased battery consumption.</string>
 
+    <string name="run_as_foreground_service">Run service with foreground priority</string>
+
+    <string name="run_as_foreground_service_summary">If enabled, Syncthing will run with foreground priority and is less likely to be stopped by Android. This might cause other services to be stopped if available memory is low.</string>
+
     <!-- Toast shown after config was successfully exported -->
     <string name="config_export_successful">Config was exported to %1$s</string>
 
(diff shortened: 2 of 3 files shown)


licaon-kter:
Wasn't the point of the notification exactly this? I though this was a feature already.

Nutomic:
You should definitely add something to the summary that this setting could result in other apps getting killed if the phone is low on resources (eg music player). Pretty sure that's the worst case if you enable it. Also, I'm not sure if we should close the related issues or keep them open. What's your opinion on that?

@licaon-kter Notification is required for an app to run in foreground.

lkwg82:
I vote for closing the related issues as this PR can be seen as a "solution". Or we let the reporter chose to close them, if they finally approve the solution.

Or last suggestion ;), offfer a custom build which the reporters can install manually to test the solution.

Nutomic:
Regarding the issues, we have to think about how users will find this option. We should probably have a FAQ somewhere for issues like this, and tell people to use the experimental options. 

A build should be available [here on Jenkins](http://android.syncthing.net/view/Syncthing-Android/job/Syncthing-Android-PR/239/) when it's finished.

licaon-kter:
I vote for it being silently enabled when the notification is on.
As per https://github.com/M66B/NetGuard/blob/master/FAQ.md#FAQ24

Nutomic:
That sounds like a bad idea, because there's a real possibility that other apps might get killed if Syncthing uses too many resources. It shouldn't be on if the user doesn't know about it. Enabling it automatically with another (unrelated) setting also makes it very confusing for users.

capi:
I'm with @Nutomic on this, I wouldn't enable it as default. Let's first see if it actually fixes problems for more than a few people. I think having it in the experimental section and pointing people to it in issues that might be affected by it, is an appropriate first step.

licaon-kter:
It's not unrelated since you actually need that notification anyway, right?

capi:
But people won't expect a foreground service just by enabling a notification. Foreground services are not really meant for file synchronization but for something the user interacts with directly (i.e. playing music, playing videos). So it is a slight abuse and it shouldn't be enabled by default, in my opinion. And as I said, let's just see if it actually fixes issues to consider moving it from "experimental".

capi:
Pushed a new version with extended explanation of the option. Also fixed a typo in the setting's name, so you'll have to re-enable it if you have been using the previous build.

Nutomic:
So you already tested this and everything works fine? I don't really feel like testing stuff myself right now, so I'll just trust you. The code looks good to me.

capi:
I've been running a simplified version without settings for a few days, I'll keep testing it a day more or two, then I'll merge. Unfortunately a thunderstorm disconnected me from the Internet for tonight, and mobile reception down to Edge, so no more to do for me today, see you tomorrow :)

Nutomic:
Oh another question, how does this play with the "keep cpu awake" experimental setting? Do we really need both? Or maybe we can merge them into one setting?

wweich:
I had problems with Syncthing getting "killed" when in standby for a while. After disabling battery saving for Syncthing and enabling the CPU wake lock setting, I don't have any problems anymore.
So if enabling the foreground service setting would eventually result in other foreground services being killed, I would like to keep the settings separate, as I (and probably others) don't need such an "aggressive" setting.

capi:
@Nutomic The docs don't say anything about foreground services and wake-locks. My interpretation would have been that the foreground priority only affects CPU scheduling priority and clean-up priority, but it won't prevent CPU from going to sleep if screen is turned off. We will need to test that.

scienmind:
@capi , can you elaborate why you expect it to also fix https://github.com/syncthing/syncthing-android/issues/634 ? 

My knowledge on android specifics is very poor, so excuse my ignorance. 
My understanding was that ST registers for invocation upon event (i.e. whitelisted network connected), while not being active in its absence. And there's some bug in "not being invoked properly".

Or does ST's wrapper keeps a background service alive for watching for this event? And then fails to launch ST due to being closed itself by OS? If so, fixing #634 by preventing this service termination would make sense.

capi:
@scienmind I am not sure if it fixes, it _might_ fix. For me several of the reports look like that the [`BroadcastReceiver`](https://developer.android.com/reference/android/content/BroadcastReceiver.html)s that are invoked by the system for WiFi / AC changes are not called (some logs indicate otherwise, but at least some look like that to me). One reason for that is, that if the system forcefully terminates a process (like the custom `BgDetect` on OxygenOS, or a "terminate all background processes"), it is also unregistered from those BroadcastReceivers, which are registered only again once the application is launched. This is a safe-guard in Android that an often-crashing service / often-terminating service enters a loop of "starting / crashing / starting / crashing ...". This may also happen if a system is shut down for power-saving reasons. Some Samsung ROMs for instance put background-services on a blacklist if the corresponding app has not been actively launched for 3 days (my Galaxy Tab S2 tells me that regularely, but I never tested Syncthing on it). In this case it also might help.

It _will not_ help for cases where we actually handle the events wrong. But those cases I still could not reproduce.

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
```diff
@@ -291,6 +292,9 @@ private boolean isAllowedWifiConnected() {
     private void updateNotification() {
         SharedPreferences sp = PreferenceManager.getDefaultSharedPreferences(this);
         String type = sp.getString(PREF_NOTIFICATION_TYPE, "low_priority");
+        boolean foreground = sp.getBoolean(PREF_FOREGROUND_SERVICE, false);
+        if ("none".equals(type) && foreground)
+            type = "low_priority"; // foreground priority requires a notification
```
Nutomic: This whole thing looks really complicated with changing the value of foreground. I think you can just get rid of this and move `builder.setPriority(NotificationCompat.PRIORITY_MIN);` to line 312.

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
```diff
@@ -304,15 +308,22 @@ private void updateNotification() {
...
-            nm.notify(NOTIFICATION_ACTIVE, builder.build());
+            if (foreground) {
+                builder.setContentText(getString(R.string.syncthing_active_foreground));
+                startForeground(NOTIFICATION_ACTIVE, builder.build());
+            } else {
+                stopForeground(false); // ensure no longer running with foreground priority
+                nm.notify(NOTIFICATION_ACTIVE, builder.build());
+            }
         } else {
+            stopForeground(false); // ensure no longer running with foreground priority
```
Nutomic: I put comments on their own line before the code, I think that's easier to see. But I don't have a strong strong preference either way ...

src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingService.java
capi: No, because if the user selected "none", we would not even enter the "if" condition in line 299.

Above code just guarantees that _at least_ "min priority" is selected. I also wouldn't want my "default priority" notification to be downgraded to "min priority" just because I enable foreground service.

capi:
Pushed updated PR, including clarifying comment for the adjustment from `none` to `low_priority` if foreground service is enabled. Also adjusted comment location as per your request.

Nutomic:
Okay I get it now. I don't think it's the perfect way to do this, but I don't really see a better option either. Merged :)

PR #697, opened 2016-07-20
Switch to wget instead of heavyweight submodules

Zillode:


xduugu:
How about adding variables for the bootstrap and the build go source somewhere at the top of the script and using these instead of the hardcoded `go-1.4.tar.gz` and `go-latest.tar.gz` archive names? Something like

```
GO_SOURCE_URL="https://storage.googleapis.com/golang/"

GO_BOOTSTRAP_SOURCE="go1.4.3.src.tar.gz"
GO_BOOTSTRAP_SHA1="486db10dc571a55c8d795365070f66d343458c48"

GO_SOURCE="go1.6.3.src.tar.gz"
GO_SHA1="b487b9127afba37e6c62305165bf840758d6adaf"
```

lkwg82:
what is the purpose you are aiming for?

Nutomic:
I like the idea, especially because it makes Syncthing updates easier. I will try to review the PR in the next days.

lkwg82:
It seems to just like another flavor. What does it bring us?

xduugu:
I guess it's faster to download the source tarballs than checking out the golang repository.

make-go.bash
```diff
@@ -2,20 +2,36 @@
...
     if [ -z $GOROOT ] || [[ $(go version) != go\ version\ go1.4* ]] ; then
-            git submodule update --init ext/golang/go1.4
-            # Build Go 1.4 for host
-            pushd ext/golang/go1.4/src
-            ./make.bash --no-clean
-            popd
-            # Add Go 1.4 to the environment
-            export GOROOT="$(pwd)/ext/golang/go1.4"
+      if [ ! -f go-1.4.tar.gz ]; then
+        wget -O go-1.4.tar.gz "$GO_SOURCE_URL$GO_BOOTSTRAP_SOURCE"
```
xduugu: `go-1.4.tar.gz` could also be replaced by `$GO_BOOTSTRAP_SOURCE`.

make-go.bash
```diff
@@ -50,10 +66,20 @@ if [ -d "$GOROOT_FINAL" ]; then
 fi
 mkdir -p "$GOROOT_FINAL"
 
+# Fetch latest Go
+if [ ! -f go-latest.tar.gz ]; then
+  wget -O go-latest.tar.gz $GO_SOURCE_URL$GO_SOURCE
```
xduugu: You could use `"$GO_BOOTSTRAP_SOURCE"` instead of `go-latest.tar.gz.

.gitignore
```diff
@@ -39,7 +39,9 @@ gradle/wrapper/gradle/
 gradle/wrapper/gradlew*
 
 # Go native dependencies
-ext/golang/dist
+ext/golang
+go-latest.tar.gz
+go-1.4.tar.gz
```
xduugu: If the real archive names are used (see comments below), this two lines need to be changed to something like `/go*.src.tar.gz`.

Zillode:
> what is the purpose you are aiming for?

To fix the timeout issues on jenkins. The cloning of golang takes a very long time.

make-go.bash
```diff
@@ -2,20 +2,36 @@
...
-            # Add Go 1.4 to the environment
-            export GOROOT="$(pwd)/ext/golang/go1.4"
+      if [ ! -f go-1.4.tar.gz ]; then
+        wget -O go-1.4.tar.gz "$GO_SOURCE_URL$GO_BOOTSTRAP_SOURCE"
+      fi
+      sha1sum=$(sha1sum go-1.4.tar.gz | awk '{print $1}')
+      if [ "$GO_BOOTSTRAP_SHA1" != "$sha1sum" ]; then
+        echo "Error: invalid checksum"
+        exit 1
+      fi
```
AudriusButkevicius: You could technically do the sha check after the wget, and trust that it's all good ;)

Nutomic:
Thanks for all the suggestions @xduugu. I'm not familiar with bash at all, so they're very welcome :)

@Zillode I don't see anything wrong with the code, so I'll just wait until you addressed the comments here, then test it and merge it.

make-go.bash
Zillode: I actually don't fully agree with this change. Readability should highly be optimized in a bash script and all these variables make it harder to read imo. But updated as requested..

xduugu: The issue I see when you use a hardcoded archive name is the following:
* The script is run, downloads go 1.6.3 and saves it as go-latest.tar.gz
* The script gets update to use go 1.6.4 --> the sha1 hash changes as well
* go-latest.tar.gz that was previously downloaded and contains go 1.6.3 will now make the build script stop because of a checksum mismatch.

xduugu:
At some point, we probably want to use functions, so we can use something like

```
download "${GO_SOURCE_URL}/${GO_SOURCE}" "$GO_SOURCE"
verify "$GO_SOURCE" "GO_SHA1"
extract "$GO_SOURCE" ext/golang/go
```

Not sure, if it needs to be in this PR, though.

.gitignore
```diff
@@ -39,7 +39,8 @@ gradle/wrapper/gradle/
 gradle/wrapper/gradlew*
 
 # Go native dependencies
-ext/golang/dist
+ext/golang
+go*.tar.gz
```
xduugu: You can even change it to `go*.src.tar.gz`.

Nutomic:
Can you change the location for the downloaded files to `build/`, or `build/tmp`? That would avoid some unnecessary clutter. Of course, they shouldn't be deleted on `gradle clean` (so you can rebuild everything without internet).

AudriusButkevicius:
I think GOOS=android binary used to crash for some versions of the OS.

Nutomic:
@capi Afaik, DNS on Android is still not working correctly in the latest Golang versions. But maybe that has been fixed in the meantime.

Nutomic:
Will merge this after you change the location of the temp files (see my previous comment).

Nutomic:
Closed due to inactivity. Feel free to reopen.

Issue #706, opened 2016-07-26
Syncthing binary: Build with GOOS=linux crashes > v0.14.1


capi:
Since `v0.14.1`, the Syncthing binary requires a temporary directory, which for the `linux` flavor is `/tmp`. Unfortunately, this does not exist on Android, causing a panic.

This can be worked around, by setting an environment variable `TMPDIR` to `mContext.getCacheDir().getAbsolutePath()`, but I think it would be sensible to work using the `android` flavor of Go over the `linux`, since I can imagine, there are other subtle differences to various aspects.

Unfortunately I don't know much about Go and the native binary for Android, so I don't feel quite up to the task of changing that. Maybe this could be looked into together with #697?

Nutomic:
@Zillode switched to `GOOS=linux` some time ago due to a problem. I don't remember what that was, though.

AudriusButkevicius:
I think it was related to the cgo chain which we do not use anymore.

Nutomic:
@capi I guess it would help if you post a crash log here.

AudriusButkevicius:
It can't write to /tmp as it does not exist on android.

capi:
It's already fixed in my PR. We just need to set `TMPDIR` (which I did) to something that exists. The Go temp-directory function takes that into account, and defaults for `/tmp` for Linux, and `/data/local/tmp` on Android. It's just one example where Linux uses other defaults than Android.

(Footnote: `/data/local/tmp` doesn't work reliably accross ROMs anyway, `context.getCacheDirectory()` works reliably, that's what I use.)

capi:
@Nutomic I'd consider this more an "enhancement" than a "bug". The building as "linux" doesn't cause the crash, it would still happen with "android" on most ROMs, it was just a suggestion by @calmh that we should working towards compiling with GOOS=android to avoid subtle problems caused by differences between Linux and Android.

Nutomic:
So this is an issue for something that you already fixed in your branch? I'm more or less on vacation now, so I don't have much time to look into this :p

capi:
It's a ticket for something that we should change in the future, since it caused an issue and likely will cause more issues in the future.

Enjoy your well-earned vacation.

capi:
[...]

PR #707, opened 2016-07-26
[WIP] Updated Syncthing to v0.14.2

capi:
I have updated the Syncthing binary to 0.14.2, since 0.14.1 introduces the delta-index feature, which could be a huge game-changer for mobile users.

Unfortunately, the first run showed that our build-process builds for `linux` and not for `android`, causing a crash, since the delta-index feature requires a temporary directory. (See also #706).

I have added the `TMPDIR` environment variable, pointed it to the application's "cache" directory, this seems to be working quite well.

Still I'd like someone else to have a look before merging this on the official repository.

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
@@ -117,6 +117,9 @@ public class SyncthingRunnable implements Runnable {
             Map<String, String> env = pb.environment();
             // Set home directory to data folder for web GUI folder picker.
             env.put("HOME", Environment.getExternalStorageDirectory().getAbsolutePath());
+            // workaround for #706 (building the binary with arm/linux instead of arm/android)
+            Log.d(TAG, "Using " + mContext.getCacheDir().getAbsolutePath() + " as TMPDIR");
+            env.put("TMPDIR", mContext.getCacheDir().getAbsolutePath());
             env.put("STTRACE", sp.getString("sttrace", ""));
             env.put("STNORESTART", "1");
             env.put("STNOUPGRADE", "1");


capi:
**We should not merge this** for now, since 0.14.1+ have issues with crashes during the delta-index exchange. I leave this open for now, so that others see that we are working on it (and we know how to get 0.14.1+ running on Android).

See https://github.com/syncthing/syncthing/issues/3448 for further information on the crash of 0.14.2

xduugu:
Since syncthing 0.14.3 creates the temporary directories in syncthing's home folder, this shouldn't be required anymore, right?

capi:
As soon as 0.14.3 is released, I'll update the submodule and open a new pull-request. I do think that it would be sensible to set the TMPDIR to something that's writeable, as in the future other parts might start using temporary files. Having Go fall-back to a non-existing /tmp is not optimal in my opinion. What do you think, @xduugu?

xduugu:
I wouldn't tinker with the TMPDIR environment variable at all. I don't see that syncthing will use the default temp directory in the near future. I rather expect that at some point a cache directory is introduced which can be adjusted by commandline parameter and environment variable.

capi:
@xduugu I disagree, since Go uses this variable internally for stuff, so having it knowingly at something wrong, is _wrong_ in my opinion. All based on the fact, that we are compiling for Linux and not for Android...

capi:
Obsoleted by #710, which will use Syncthing v0.14.3 and doesn't require the `TMPDIR` fix.

xduugu:
@capi
But the cause, in my opinion, is not the TMPDIR variable, but the use of a "linux" binary. When you look at [os.TempDir()](https://golang.org/src/os/file_unix.go?s=9565:9586#L336), you see that the default return value is `/data/local/tmp` for an "android" binary, not `/tmp`.

```
func TempDir() string {
	dir := Getenv("TMPDIR")
	if dir == "" {
		if runtime.GOOS == "android" {
			dir = "/data/local/tmp"
		} else {
			dir = "/tmp"
		}
	}
	return dir
}
```

capi:
@xduugu Yes, that's exactly my point. We know that we currently will enter the wrong branch here. Always. It doesn't help, that "/data/local/tmp" is not writable in most ROMs anyways. :) Regardless, it's now moot, my new PR doesn't set it any more, since it's not required in 0.14.3+ for now.

PR #710, opened 2016-07-29
Update Syncthing to v0.14.3

capi:
Provides delta-indexes (which could hugely improve usability while on mobile data).

Obsoletes #707 (v0.14.2) in favor of this bug-fix release. This version also does not require the `TMPDIR` workaround of #707 (see #706).


fti7:
Please merge! :-)

PR #711, opened 2016-08-02
Disable sync if Android sync setting is disabled

Nutomic:
This is completely untested at the moment.

Also, should we add a preference whether to consider the master sync setting?

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -12,6 +12,7 @@
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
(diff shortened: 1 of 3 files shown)


Nutomic:
Turns out we have to check seperately for battery saving mode. I've added that in a new commit. 

Should we add settings to disable Syncthing according to sync/battery saver status? The problem is that it would make everything more complicated, so I'm not sure.

Also, it would probably be nice to change the disabled dialog to include why Syncthing was disabled.

capi:
I think that it is sensible to add it like this for now. The only problem I see right now is that it will not sync even though the application is opened and in foreground. And the guide-lines for auto-sync say, it should prevent _background sync_, but applications should work normally if they are in foreground. so this would complicate this, I suppose.

Nutomic:
@capi I'm not quite sure how that should work, especially together with the "only on internet"/"only while while charging" settings. So I think we should discuss that first, and then maybe implement it in the future.

PR #713, opened 2016-08-02
Set user/password in config instead of removed STGUIAUTH (fixes #704).

### LINKED ISSUE #704: WebGUI username and password are not saved
Hi,
I have changed the listener to 0.0.0.0:8384 and also set username and password to access the website from other devices. 
But i noticed that there is no logon. i can directly access webpage without the need to logon. (Also the warning will be shown on the Website)
Tried to restart app/phone but it doesn't make any difference. (Android 6.0.1 App V0.8.0)

Nutomic:
The web interface isn't working yet, I don't know why that is. Everything else seems to work.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,6 +25,7 @@ dependencies {
     compile 'com.android.support:design:24.1.1'
     compile 'com.google.zxing:android-integration:3.2.1'
     compile 'com.google.code.gson:gson:2.7'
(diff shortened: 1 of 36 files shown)


Nutomic:
@Zillode: Maybe you can check why the web interface doesn't work? I'm quite clueless.

Nutomic:
Merged, not sure why Github doesn't pick that up...

Issue #700, opened 2016-07-22
WLAN Filter list does not Work


Wombatsmann:
I activatec Sync only in WLAN. But if I select my Network Name in the WLAN filter list the app asks to end or change settings. Tried this in different Wlans.

capi:
Does it work with an empty list (i.e. sync on all WiFis)? If not, your Android is not detecting that it has Internet connection via the WiFi and does not consider the WiFi as connected.

I'm using the WiFi white-list every day and I'm surprised it's not working. Which version are you using? (I'm still on the 0.7.x version (0.13 Syncthing)).

Wombatsmann:
Yes it works with an empty list. I updated to Syncthing 0.14 and this problem is still there.
Its an Samsung GT-S7710 on Android 4.1.2

capi:
Can you provide me with an `adb log` of the device when switching on/off the entries in the list? I just compiled and installed this version and it works flawlessly in my environment. So it might be that the device does not report the WiFi name correctly but I'd need to get the logs to have a look.

Wombatsmann:
This is the output of adb logcat. I activated the Network filter and selected my Wlan. Then the popup came up and asked me to end it or change settings. Thats when I started the log to adb_logcat.txt. The other file is the way back.

Am 22. Juli 2016 21:36:54 MESZ, schrieb Martin Carpella <notifications@github.com>:

Wombatsmann:
Sorry I didnt know that the reply would post this here. Did you get the logcat.zip? I don't want to post it here in fact of my personal informations like ST folder Structure etc.

Nutomic:
I don't think Github supports email attachments. But you should be able to send it to @capi and me on the [forum](https://forum.syncthing.net/). Usernames are the same.

capi:
@Wombatsmann I finally had time to review the adb-logs you submitted. Everything works fine with `All SSIDs allowed for syncing` until suddenly `SSID <redacted> not whitelisted`.

Are you sure you did not change any settings in "Restrict to certain wifi networks" option and removed the SSID in question?

Another possibility would be that there are some characters in the WiFi name, which I never tested (space and `+` sign, actually). Maybe the selector UI messes up the SSID name somehow. I'll add additional logging, so we can be diagnose this better.

capi:
I think I found something:

`WifiInfo.getSSID()` specifies:
[...]

PR #720, opened 2016-08-11
Log SSID whitelist for better troubleshooting of #700

capi:
To better diagnose issue #700, log the whitelist alongside the connected SSID when `isAllowedWifiConnected` is not connected.


Nutomic:
I guess that makes sense, thanks :)

PR #721, opened 2016-08-11
Explain requirement for notification in "Run as foreground" setting (fixes #715)

### LINKED ISSUE #715: Notification types should depend on foreground service state
As of 0.8.1

If Experimental/Foreground is disabled (this is how it looks now):
* Normal
* Low Prio
* None

If Experimental/Foreground is enabled (there is no None as a Notification is needed for the foreground service anyway):
* Normal
* Low Prio (activated by default)

capi:
Add suggested explanation of #715 to `run_as_foreground_service_summary`.

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -302,7 +302,7 @@ Please report any problems you encounter via Github.</string>
 
     <string name="run_as_foreground_service">Run service with foreground priority</string>
 
-    <string name="run_as_foreground_service_summary">If enabled, Syncthing will run with foreground priority and is less likely to be stopped by Android. This might cause other services to be stopped if available memory is low.</string>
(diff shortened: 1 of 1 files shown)


Nutomic:
@licaon-kter Do you think this is enough to fix the issue?

@capi Can you try to keep commit messages under 80 characters, so Github and other tools don't cut the end off? That should make it easier to read logs :)

AudriusButkevicius:
`other tools`  - your VT100?

capi:
@Nutomic Ups, sorry, yeah, normally I commit via git-gui, which warns about this. This one was done via shell and `-m`, so I missed out. Will try to behave better in the future :)

@AudriusButkevicius :D

licaon-kter:
@Nutomic It's not a fix but a clarification, maybe when foreground exits experimental state this needs a re-evaluation.

Nutomic:
@capi Okay, I'll merge this after you shorten the commit message.

@licaon-kter Please remind us then ;)

capi:
@Nutomic commit message shortened

Nutomic:
Thanks :)

PR #724, opened 2016-08-19
Support TOR connectivity (fixes #589).

### LINKED ISSUE #589: Support Tor connectivity
There are three parts

1. Support connecting via SOCKS5 proxy. Not so Tor specific and may be useful anyway.

2. Support .onion names as described in https://tools.ietf.org/html/rfc7686

3. Provide a single checkbox labeled "Use Tor" that: a) configures socks5 proxy at localhost:9050, b) disables connecting to any non-.onion adresses.

Tor solves two problems both of which apply IMHO to Syncthing more or less. First, Tor provides hides communication from evasdroppers. Second, hidden services are quite convinient way to bypass NAT.

I suppose some of the above tasks apply to the syncthing engine rather than android app but as I am unfamiliar with syncthing internals please allow me to report them here.

Nutomic:
I did a quick test and it seems to work. Would be nice if someone can confirm this though.


Nutomic:
Actually it should definitely work, so I'm just gonna merge this.

PR #738, opened 2016-09-06
[WIP] Speed up Hashing for ARMv8

Nutomic:
See my [forum post](https://forum.syncthing.net/t/faster-hashing-package/8186/15?u=nutomic).

In short, the changes in this PR improve the hashing performance from 33 MB/s to 817 MB/s on my LG G4. 

TODO:
- [x] Figure out why no arm64-v8a apk is generated
- [x] Figure out why Syncthing binary is not extracted from apk on ARMv8 devices
- [x] Wait for sha256-simd branch to be merged into Syncthing?

### DIFF
--- a/make-all.bash
+++ b/make-all.bash
@@ -8,3 +8,5 @@
 ./make-go.bash 386
 ./make-syncthing.bash 386
 
+./make-go.bash arm64
+./make-syncthing.bash arm64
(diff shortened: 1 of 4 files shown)


Nutomic:
I still have the problem that the ARMv8 binary is not extracted from the apk. I've also reproduced the problem in the emulator, so it's not something device specific. I posted a question [on stackoverflow](http://stackoverflow.com/q/39478884/1837158), but nothing that helped so far.

Nutomic:
Yay I figured it out :D

lkwg82:
Great! Thx for your effort.

PR #740, opened 2016-09-12
Update to Syncthing v0.14.6

capi:
Updates the Syncthing binary to v0.14.6, and also addresses the change in device completion calculation introduced by this Syncthing version.


capi:
@Nutomic Updated comments as per your request.

Nutomic:
Thanks :)

PR #764, opened 2016-10-19
Set STGUIASSETS to let user add custom themes without root

wweich:
Set the STGUIASSETS env var to primary storage  (e.g. `/storage/emulated/0/Android/data/com.nutomic.syncthingandroid/gui`) to be able to add custom themes.

As I am not very fond of the new dark / gray theme (especially on an OLED display), I would like to add the old dark theme, which isn't possible on Android at the moment (without root).

This should fix that.

The path defaults to $STHOME/gui and Syncthing has no problem if that doesn't exist, so there should be no problem here either. That's also why I don't think there should be a setting for that.

I'm not an Android dev and have no Android dev environment to test this, but from what I read, it should work as expected ;)

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
+++ b/src/main/java/com/nutomic/syncthingandroid/syncthing/SyncthingRunnable.java
@@ -118,6 +118,7 @@ public class SyncthingRunnable implements Runnable {
             // Set home directory to data folder for web GUI folder picker.
             env.put("HOME", Environment.getExternalStorageDirectory().getAbsolutePath());
             env.put("STTRACE", sp.getString("sttrace", ""));
+            env.put("STGUIASSETS", getExternalFilesDir(null).getAbsolutePath() + "/gui");
             env.put("STNORESTART", "1");
             env.put("STNOUPGRADE", "1");
             if (sp.getBoolean("use_tor", false)) {


AudriusButkevicius:
Can you make a PR adding back the dark theme back as an option and we end this bullshit once and forever

wweich:
I can do that, sure.

But @calmh was the one, who didn't want a new theme but an updated dark theme, in the first place.

> I think two themes shipped by default are enough for the moment [(source)](https://forum.syncthing.net/t/0-14-7-refined-dark-theme-is-really-bad/8292/4?u=wweich)

AudriusButkevicius:
Sure, yet I think the amount of noise this produces is too much when compared to shipping an extra file, and I think @calmh will agree with me.

wweich:
Giving the users of the Android App the possibility to add custom themes is about more than just getting back the old dark theme, so it could be done in addition to resurrecting the old theme.

capi:
Any drawbacks in always setting the environment variable vs. only setting it when an option is set?

wweich:
The internal setting is always set. If STGUIASSETS isn't set, it will use the default.
https://github.com/syncthing/syncthing/blob/29ccf10d0b8b2b4ef90e912f5affbdb6696a6f10/cmd/syncthing/main.go#L314
So I don't think there are any drawbacks.

The only thing I can think of is, if someone doesn't know about this and creates the gui folder for some reason and the content of that folder breaks Syncthing.

Nutomic:
I don't see any problem with having this. And we already have tons of options, no point adding another one that does basically nothing. I'm just gonna wait until Travis is done (no idea why that failed).

@calmh Can you disable @st-review for this repo? Cause we don't need it, and it just spams extra branches.

capi:
Yeah, I also don't see that it's a real-world risk that another process could inject malicious JS files into the directory and mess with our users. A process with write-access to the external storage could most likely simply read the data Syncthing is syncing anyways. Just wanted to make sure, what's the worst that could happen.

Nutomic:
That same risk also exists on a desktop, and no one complains there. As long as the app has the same security as Syncthing on its own, that should be fine.

Travis seems broken, will look into that later.

@wweich Can you add some documentation for this, eg in the [Android FAQ](https://github.com/syncthing/syncthing-android/wiki/Frequently-Asked-Questions). Or wherever else themes are documented.

wweich:
Done. [here](https://docs.syncthing.net/users/faq.html#i-don-t-like-the-gui-theme-can-it-be-changed) and [here](https://github.com/syncthing/syncthing-android/wiki/Frequently-Asked-Questions#how-can-i-add-a-custom-webui-theme).

capi:
This failed for me on a 5.1.1 emulator, which doesn't have a simulated SD-card. I'll investigate:
```
10-20 17:51:55.985 2735-2862/com.nutomic.syncthingandroid.debug E/AndroidRuntime: FATAL EXCEPTION: AsyncTask #1
  Process: com.nutomic.syncthingandroid.debug, PID: 2735
  java.lang.RuntimeException: An error occured while executing doInBackground()
      at android.os.AsyncTask$3.done(AsyncTask.java:304)
      at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:355)
      at java.util.concurrent.FutureTask.setException(FutureTask.java:222)
      at java.util.concurrent.FutureTask.run(FutureTask.java:242)
      at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:231)
      at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112)
      at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587)
      at java.lang.Thread.run(Thread.java:818)
   Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String java.io.File.getAbsolutePath()' on a null object reference
      at com.nutomic.syncthingandroid.syncthing.SyncthingRunnable.run(SyncthingRunnable.java:121)
      at com.nutomic.syncthingandroid.util.ConfigXml.generateKeysConfig(ConfigXml.java:83)
      at com.nutomic.syncthingandroid.util.ConfigXml.<init>(ConfigXml.java:65)
      at com.nutomic.syncthingandroid.syncthing.SyncthingService$StartupTask.doInBackground(SyncthingService.java:331)
      at com.nutomic.syncthingandroid.syncthing.SyncthingService$StartupTask.doInBackground(SyncthingService.java:326)
      at android.os.AsyncTask$2.call(AsyncTask.java:292)
      at java.util.concurrent.FutureTask.run(FutureTask.java:237)
      at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:231) 
      at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1112) 
      at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:587) 
      at java.lang.Thread.run(Thread.java:818) 
10-20 17:51:56.119 1472-2248/system_process W/ActivityManager:   Force finishing activity 1 com.nutomic.syncthingandroid.debug/com.nutomic.syncthingandroid.activities.MainActivity
````

The crashing line is the

    env.put("STGUIASSETS", mContext.getExternalFilesDir(null).getAbsolutePath() + "/gui");

Don't know if this will have impact on any other "real-world" device.

PR #768, opened 2016-10-20
Fix NullPointerException for STGUIASSETS

capi:
On devices without SD-card / shared-storage, `getExternalFilesDir()` may return `null`.

I managed to get this on an 5.1.1 emulator without SD-card, and it's part of the JavaDoc for `getExternalFilesDir()`.


AudriusButkevicius:
You are making a call twice for no good reason.

capi:
Yes, I clicked the wrong refactoring, sorry :) Will be fixed in a minute.

capi:
Fixed.

Nutomic:
I didn't think this would actually happen. Thanks for fixing it!

PR #779, opened 2016-11-05
Reload URL and API key when importing config

### LINKED ISSUE #777: Import settings dialog are not completing in new layout
For clean instalation of 0.9.0-beta1 in STB with Android 5.1.1  there is a problem after "Backup -> Import Configuration" command.

- When importing config from old 0.8.8 release it shows "Generating security key" window which is never ending.
- When importing config from the same version 0.9.0-beta1 it shows "Loading" window which is never ending to.

For both cases application starts immediately. There is access from web GUI with correct new settings.

### Version Information

   App Version: 0.9.0-beta1
   Syncthing Version: v0.14.9
   Build platform: Ubuntu 16.04 LTS
   Android Version:  5.1.1

Unrud:
Fixes #777


Nutomic:
Thank you :)

Issue #807, opened 2016-12-10
Dialog disappear on orientation change


reverse-unina:
When i perform these sequences of events, the dialog on screen disappears

Sequence 1: 
1. Click on Folder item
2. Click on Remove icon
3. Orientation Change

Sequence 2:
1. Open Side Drawer
2. Click on Restart
3. Orientation Change

Sequence 3:
1. Click on Add Folder icon
2. Click on Back
3. Orientation Change

Sequence 4:
1. Click on Devices
2. Click on Add Device icon
3. Click on back
4. Orientation Change

Sequence 5:
1. Click on Devices
2. Click on Add Device icon
3. Click on Compression
4. Orientation Change

licaon-kter:
You must love to rotate your device. :)

reverse-unina:
Yes, indeed 😃 
I am rotating everything I can and exploring mobile GUI failures for research purposes.

licaon-kter:
And thank you for that. The whole issue is about Android recreating the view on rotation, iirc.

Issue #804, opened 2016-12-10
Dialog appear on device rotation


reverse-unina:
When I perform this sequence of events, an unwanted dialog appear on the screen

1.  Click Later in Battery Optimization dialog
2. Orientation Change

shobhitagarwal1612:
I'd like to tackle this issue

Nutomic:
Cool! You should basically follow [this approach](https://stackoverflow.com/questions/151777/saving-android-activity-state/151940#151940). Let me know if you need more help ;)

shobhitagarwal1612:
@Nutomic  I have built the project but it crashes just after pressing continue button at the starting screen. I have checked on Android 5.0 and Android 6.0 . Please help!


> 02-03 11:31:33.533 8472-8937/com.nutomic.syncthingandroid.debug W/SyncthingRunnable: Failed to trim log file
>                                                                                      java.io.FileNotFoundException: /storage/emulated/0/Android/data/com.nutomic.syncthingandroid.debug/files/syncthing.log: open failed: ENOENT (No such file or directory)
>                                                                                          at libcore.io.IoBridge.open(IoBridge.java:452)
>                                                                                          at java.io.FileInputStream.<init>(FileInputStream.java:76)
>                                                                                          at com.google.common.io.Files$FileByteSource.openStream(Files.java:120)
>                                                                                          at com.google.common.io.Files$FileByteSource.openStream(Files.java:110)
>                                                                                          at com.google.common.io.ByteSource$AsCharSource.openStream(ByteSource.java:456)
>                                                                                          at com.google.common.io.CharSource.readLines(CharSource.java:311)
>                                                                                          at com.google.common.io.Files.readLines(Files.java:553)
>                                                                                          at com.google.common.io.Files.readLines(Files.java:520)
>                                                                                          at com.nutomic.syncthingandroid.service.SyncthingRunnable.trimLogFile(SyncthingRunnable.java:344)
>                                                                                          at com.nutomic.syncthingandroid.service.SyncthingRunnable.run(SyncthingRunnable.java:94)
>                                                                                          at com.nutomic.syncthingandroid.util.ConfigXml.generateKeysConfig(ConfigXml.java:83)
>                                                                                          at com.nutomic.syncthingandroid.util.ConfigXml.<init>(ConfigXml.java:64)
>                                                                                          at com.nutomic.syncthingandroid.service.SyncthingService$StartupTask.doInBackground(SyncthingService.java:335)
[...]

Issue #80, opened 2014-08-08
Folder picker lets only add folders from SDCARD0


Eucyon:
When adding a new repository. Folder from SDCARD1 can only be added in the webgui.

PR #828, opened 2017-01-07
Stop screen rotation from redrawing activity.  (fix #805 , #807 & #804)

### LINKED ISSUE #805: Picker changes state on device rotation
When I perform this sequence of events, the picker on screen loose the user input and go back to its default value;

1. Click on Folder item
2. Click on Keep Versions
3. Pick a versions number
4. Orientation Change

jmintb:
The changes to the manifest tell android that we want to handle screen changes and hence the activity if not recreated apon rotating the screen. The same goes for the folder and device activities. (fix #805 , #80  7 & #804)

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -40,7 +40,9 @@
         <activity
             android:name=".activities.MainActivity"
             android:label="@string/app_name"
-            android:launchMode="singleTop">
+            android:launchMode="singleTop"
+            android:configChanges="orientation|screenSize"
+            >
         </activity>
         <activity
             android:name=".activities.WebGuiActivity"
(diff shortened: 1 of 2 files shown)


Nutomic:
What you're doing is basically a hack to work around the symptom, but doesnt really fix the problem. [See this stackoverflow post](http://stackoverflow.com/a/7990543). 

I would prefer if we can solve this properly, using `onSaveInstanceState()` and `onRestoreInstanceState()` ([see here](http://stackoverflow.com/a/151940)).

Nutomic:
Closed due to inactivity. Feel free to reopen.

jmintb:
Hi, sorry for the radio silence, it looks like the issues are still open, so I will have another crack solving them  in a proper way.

jmintb:
I have a patch to submit now, should I open a new pull request?

Nutomic:
Sure, go ahead.

PR #830, opened 2017-01-08
Click region of Open folder button

PaulAnnekov:
Open folder button click region is incredibly small. It's almost impossible to catch it.

### DIFF
--- a/src/main/res/layout/item_folder_list.xml
+++ b/src/main/res/layout/item_folder_list.xml
@@ -5,32 +5,69 @@
     <RelativeLayout
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        android:paddingBottom="8dp"
-        android:paddingLeft="@dimen/abc_action_bar_content_inset_material"
-        android:paddingRight="@dimen/abc_action_bar_content_inset_material"
-        android:paddingTop="8dp"
         android:descendantFocusability="blocksDescendants">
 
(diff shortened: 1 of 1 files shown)


Nutomic:
From the diff, it seems like you are changing a lot of things for no reason. I'd prefer if you just change the padding of the ImageBtton, because that's all you need to do.

PaulAnnekov:
@Nutomic that's what I initially did (1st commit). But these top and right paddings make bad UX when clicking the button. Sometimes you click on them instead of button, so folder editor opens instead of what user needed.

I have just created a new RelativeLayout and moved all items to it, except ImageButton. So it takes all left space, while other items are still correctly padded. Real changes are much less, than what diff shows.

Nutomic:
Okay that makes sense. I didnt try it yesterday because it was late, but it actually works well on the device. What do you think about adding a vertical seperator, to make it clear that the entire right area opens the folder?

PaulAnnekov:
I'm not UX expert, but when I was developing Android apps I had a rule - "if you don't know how to design this piece of UI - check Android design guides or Google apps". 

So just open https://material.io/guidelines/layout/metrics-keylines.html#metrics-keylines-keylines-spacing and scroll to `Detail view` and below. You will find right buttons ("Info" and "SMS")  w/o any borders are quite common thing. 

android_no_borders

PR #843, opened 2017-02-05
[WIP] Ability to share to syncthing

PaulAnnekov:
Inspired by Google Drive's "Save to Drive".

TODO:
- [x] Test `file://` share (only `content://` was tested now)
- [x] Test on another OS version (was test on 5.1.1) and device
- [ ] Check if folders getter implemented correctly
- [ ] Tests (?) 

Screenshot

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -1,2 +1,0 @@
-ownCloud Android (GNU GPL v2.0 https://github.com/owncloud/android/blob/master/LICENSE.txt)
-    Part of Sharing-related code in ShareActivity
--- a/CREDITS
+++ b/CREDITS
@@ -1,0 +1,2 @@
+ownCloud Android (GNU GPL v2.0 https://github.com/owncloud/android/blob/master/LICENSE.txt)
+    Part of Sharing-related code in ShareActivity
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -51,6 +51,15 @@
                 android:value=".activities.MainActivity" />
         </activity>
         <activity
+            android:name=".activities.ShareActivity"
+            android:label="@string/title_activity_share">
+            <intent-filter>
+                <action android:name="android.intent.action.SEND" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <data android:mimeType="*/*" />
+            </intent-filter>
+        </activity>
+        <activity
             android:name=".activities.LogActivity"
             android:label="@string/log_title"
             android:parentActivityName=".activities.SettingsActivity">
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -52,7 +52,9 @@
         </activity>
         <activity
             android:name=".activities.ShareActivity"
-            android:label="@string/share_activity_title">
+            android:label="@string/share_activity_title"
+            android:excludeFromRecents="true"
+            android:taskAffinity="">
             <intent-filter>
                 <action android:name="android.intent.action.SEND" />
                 <category android:name="android.intent.category.DEFAULT" />
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -58,6 +58,11 @@
                 <category android:name="android.intent.category.DEFAULT" />
                 <data android:mimeType="*/*" />
             </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.SEND_MULTIPLE" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <data android:mimeType="*/*" />
+            </intent-filter>
         </activity>
         <activity
             android:name=".activities.LogActivity"
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -52,7 +52,7 @@
         </activity>
         <activity
             android:name=".activities.ShareActivity"
-            android:label="@string/title_activity_share">
+            android:label="@string/share_activity_title">
             <intent-filter>
                 <action android:name="android.intent.action.SEND" />
                 <category android:name="android.intent.category.DEFAULT" />
--- a/src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
+++ b/src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
@@ -127,6 +127,9 @@ public class ShareActivity extends SyncthingActivity
 
     // Taken from ownCloud Android
     // (GNU GPL v2.0 https://github.com/owncloud/android/blob/master/LICENSE.txt)
+    /**
+     * Generate file name for new file.
+     */
     private String generateDisplayName() {
         Date date = new Date(System.currentTimeMillis());
         DateFormat df = DateFormat.getDateTimeInstance();
@@ -134,6 +137,9 @@ public class ShareActivity extends SyncthingActivity
                 df.format(date));
     }
 
+    /**
+     * Get file name from uri.
+     */
     private String getDisplayNameForUri(Uri uri) {
         String displayName;
 
@@ -167,6 +173,9 @@ public class ShareActivity extends SyncthingActivity
         return displayName != null ? displayName.replaceAll("/", "-") : null;
     }
 
+    /**
+     * Get file name from content uri (content://).
+     */
     private String getDisplayNameFromContentResolver(Uri uri) {
         String displayName = null;
         String mimeType = getContentResolver().getType(uri);
--- a/src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
+++ b/src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
@@ -189,26 +189,19 @@ public class ShareActivity extends SyncthingActivity
                 displayNameColumn = MediaStore.Files.FileColumns.DISPLAY_NAME;
             }
 
-            Cursor cursor = null;
-            try {
-                cursor = getContentResolver().query(
-                        uri,
-                        new String[]{displayNameColumn},
-                        null,
-                        null,
-                        null
-                );
-                if (cursor != null) {
-                    cursor.moveToFirst();
-                    displayName = cursor.getString(cursor.getColumnIndex(displayNameColumn));
-                }
-            } catch (Exception e) {
-                Log.e(TAG, "Could not retrieve display name for " + uri.toString(), e);
-                // nothing else, displayName keeps null
-            } finally {
-                if (cursor != null) {
-                    cursor.close();
-                }
+            Cursor cursor = getContentResolver().query(
+                    uri,
+                    new String[]{displayNameColumn},
+                    null,
+                    null,
+                    null
+            );
+            if (cursor != null) {
+                cursor.moveToFirst();
+                displayName = cursor.getString(cursor.getColumnIndex(displayNameColumn));
+            }
+            if (cursor != null) {
+                cursor.close();
             }
         }
         return displayName;
--- a/src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
+++ b/src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
@@ -249,7 +249,7 @@ public class ShareActivity extends SyncthingActivity
                     try {
                         if (inputStream != null)
                             inputStream.close();
-                    } catch (Exception e) {
+                    } catch (IOException e) {
                         Log.w(TAG, "Exception on input/output stream close", e);
                     }
                 }
--- a/src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
+++ b/src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
@@ -1,0 +1,100 @@
+package com.nutomic.syncthingandroid.activities;
+
+import android.content.Intent;
+import android.net.Uri;
+import android.os.Bundle;
+import android.text.TextUtils;
+import android.widget.ArrayAdapter;
+import android.widget.Spinner;
+import android.widget.Toast;
+
+import com.nutomic.syncthingandroid.R;
+import com.nutomic.syncthingandroid.model.Folder;
+import com.nutomic.syncthingandroid.service.SyncthingService;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ *
+ */
+public class ShareActivity extends SyncthingActivity
+        implements SyncthingActivity.OnServiceConnectedListener, SyncthingService.OnApiChangeListener {
+    @Override
+    public void onApiChange(SyncthingService.State currentState) {
+        if (currentState != SyncthingService.State.ACTIVE)
+            return;
+
+        Toast.makeText(this, "test", Toast.LENGTH_SHORT).show();
+        if (getApi() == null)
+            return;
+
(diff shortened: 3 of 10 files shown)


Nutomic:
Cool :)

Regarding tests, they are somewhat broken right now, so I won't mind if you skip them. We should probably rewrite them at some point. Also, don't add any translations other for english, because they will be overwritten by the Transifex tool.

edrex:
@PaulAnnekov I'm happy to test this a bit in the emulator once you feel like it's ready.

I should mention that I've been using [Cabinet](https://play.google.com/store/apps/details?id=com.afollestad.cabinet&rdid=com.afollestad.cabinet)'s local folder sharing intent to share things to Syncthing.

PaulAnnekov:
@edrex I hope I will find some time to complete it.

PaulAnnekov:
Checked both schemas and checked on Android 7.1.1 emulator. Works pretty good. Checked share from Google Drive/Total Commander/Gallery...
@edrex You can test a new version by downloading [this file](https://github.com/syncthing/syncthing-android/files/826059/syncthing-android-debug.zip). Rename it `.zip` -> `.apk`, install and test.
@Nutomic you can test it too ^^.

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
```diff
@@ -0,0 +1,283 @@
...
+                Log.e(TAG, "Could not retrieve display name for " + uri.toString(), e);
+                // nothing else, displayName keeps null
+            } finally {
+                if (cursor != null) {
+                    cursor.close();
+                }
+            }
+        }
+        return displayName;
+    }
```
Nutomic: I'm not sure what all the `getDisplayName*()` methods do. Do we really need all of them? If so, can you add some comments to describe what they do?

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
```diff
@@ -0,0 +1,283 @@
...
+                        new String[]{displayNameColumn},
+                        null,
+                        null,
+                        null
+                );
+                if (cursor != null) {
+                    cursor.moveToFirst();
+                    displayName = cursor.getString(cursor.getColumnIndex(displayNameColumn));
+                }
+            } catch (Exception e) {
```
Nutomic: Please don't catch `Exception` directly, always catch the specific exception class(es). Actually, this code and some other dont even throw anything.

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
```diff
@@ -0,0 +1,283 @@
...
+            }
+        }
+        return displayName;
+    }
+
+    private class CopyFilesTask extends AsyncTask<Void, Void, Boolean> {
+        ProgressDialog progress;
+        Map<Uri, String> files;
+        Folder folder;
+        int copied = 0, ignored = 0;
```
Nutomic: These should be private. And naming convention for member variables is eg `mProgress`.

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
```diff
@@ -0,0 +1,283 @@
...
+                        ignored++;
+                        continue;
+                    }
+                    inputStream = getContentResolver().openInputStream(entry.getKey());
+                    outputStream = new FileOutputStream(outPath);
+                    byte[] buffer = new byte[4096];
+                    int count;
+                    while ((count = inputStream.read(buffer)) > 0) {
+                        outputStream.write(buffer, 0, count);
+                    }
```
Nutomic: You should be able to replace all this with `Files.asByteSink(outFile).writeFrom(inputStream);`.

src/main/res/values/themes.xml
```diff
@@ -19,6 +19,8 @@
         <item name="drawerArrowStyle">@style/Widget.Syncthing.DrawerArrowToggle</item>
 
         <item name="android:actionBarSize">@dimen/abc_action_bar_default_height_material</item>
+
+        <item name="android:spinnerStyle">@android:style/Widget.Holo.Light.Spinner</item>
```
Nutomic: Whats the reason you're setting Holo spinner here? It should use the material spinner by default, which es exactly what we want.

Nutomic:
Thanks! General functionality looks good, I've added some comments regarding the code.

src/main/AndroidManifest.xml
```diff
@@ -51,6 +51,20 @@
...
+            android:label="@string/share_activity_title">
+            <intent-filter>
+                <action android:name="android.intent.action.SEND" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <data android:mimeType="*/*" />
+            </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.SEND_MULTIPLE" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <data android:mimeType="*/*" />
```
Nutomic: Can we limit the mime type somehow? Right now Syncthing shows up for sharing everywhere (even for plain text), and then just shows the "Nothing to share" toast.

Nutomic:
Oh and from your comment (line 86), you also want to support plain text sharing? I dont really think that makes sense for Syncthing.

CREDITS
```diff
@@ -0,0 +1,2 @@
+ownCloud Android (GNU GPL v2.0 https://github.com/owncloud/android/blob/master/LICENSE.txt)
+    Part of Sharing-related code in ShareActivity
```
PaulAnnekov: I really didn't find any guides on how to make credits, so taken general approach. Will move to activity code, but when we will have >1 credits - better to move them to separate file, like this one.

src/main/res/layout/activity_share.xml
```diff
@@ -0,0 +1,91 @@
...
+
+            <EditText
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:ems="10"
+                android:id="@+id/name"
+                android:inputType="text|textMultiLine"
+                android:layout_below="@+id/namesTitle"
+                android:layout_alignParentLeft="true"
+                android:layout_alignParentStart="true" />
```
PaulAnnekov: 1. Why not? That's like Google Drive do. You can change file name on share.
2. There can be conflicts with existing files. As for now - copy is ignored, but in future you should change the file name.

CREDITS
Nutomic: I think it makes more sense to have the credits right where we use the code. Besides, we already have lots of files in the repo root, and I would prefer to keep that clean.

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
```diff
@@ -0,0 +1,283 @@
...
+
+    @Override
+    public void onApiChange(SyncthingService.State currentState) {
+        if (currentState != SyncthingService.State.ACTIVE || getApi() == null)
+            return;
+
+        List<Folder> folders = getApi().getFolders();
+
+        ArrayAdapter<Folder> adapter = new ArrayAdapter<>(
+                this, android.R.layout.simple_spinner_item, folders);
```
PaulAnnekov: I wanted to get rid of one more `.xml` creation. It's fixed very fast: 
```
public String toString() {
        return label != null ? label : id;
    }
```
Still want separate file? I agree, it's more correct way, but implies new file creation, when anyway `toString()` is not used by anybody + it's still rather good return value.

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
PaulAnnekov: It's a copy-paste from https://github.com/owncloud/android/blob/79664304fdb762b2e04f1ac505f50d0923ddd212/src/com/owncloud/android/utils/UriUtils.java#L193. Didn't want refactor it. Do you?

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
PaulAnnekov: Again, copy-paste. I thought they are self-explanatory: 
- `getDisplayNameFromContentResolver` - gets file name from content resolver. Uses special calls to it to get nice assigned name.
- `getDisplayNameForUri` - get display name for uri (bith `file://` and `content://`).
- `generateDisplayName` - generate new display name. Used if we can't determine shared file name. Can happen for some apps, which don't provide such data.

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
PaulAnnekov: Will fix.

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
PaulAnnekov: Yep. Didn't find this example.

src/main/res/values/themes.xml
PaulAnnekov: Will fix. Forgot to change during experiments.

src/main/AndroidManifest.xml
PaulAnnekov: When you share plain text file - you get its contents from `Intent.EXTRA_STREAM` which will be a url. You can then get file name based on schema.
When you share some text or memo the contents will be inside `Intent.EXTRA_TEXT`.
And both these cases will be with `mimeType = 'text/plain'`. You can't pre-filter intent by EXTRA_* fields.
Google Drive handles everything. And that's what I want too.

PaulAnnekov:
@Nutomic for example, why can't I share apps log files? Or some text, my friend written me in whatsapp?
We don't have any usage stats. We can't say if it will be useful for our users. But Google Drive have. And Google Drive can handle plain text. So, probably, their metrics say it's useful, if this feature is still there :).

PaulAnnekov:
Made changes you requested. Attached new apk.
[syncthing-android-debug.zip](https://github.com/syncthing/syncthing-android/files/832074/syncthing-android-debug.zip)

src/main/res/layout/activity_share.xml
Nutomic: I dont see a rename option in Google Drive ([image](http://i.imgur.com/XCIiitO.png)). If you want to support renaming, there should probably be seperate EditTexts for each file, or dialogs.

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
Nutomic: Yes that seems fine.

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
Nutomic: Yes please. This code will silently ignore NPEs and other errors.

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
Nutomic: I meant comments in the source code ;)

Nutomic:
Right, I dont know how Google Drive handles all this. If they support text files, it makes sense that we also support it.

src/main/res/layout/activity_share.xml
PaulAnnekov: If you share single file - you can rename it in Google Drive. If you share multiple files - you can't and they are all in the same text edit, but in disabled state. We mimic their behavior.

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
PaulAnnekov: I understand. But I thought you will understand that these names are really self-explanatory. Will add them in code.

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
PaulAnnekov: Will do.

PaulAnnekov:
@Nutomic done fixing, check, please.

Nutomic:
You're still catching `Exception` in lines 205 and 259. Another thing I noticed is that the activity still shows up in recents after clicking cancel or save. It might be appropriate to [exclude the activity from recents](http://stackoverflow.com/a/11059252).

Functionality-wise everything looks good, so I will merge after you fix these minor style issues.

PaulAnnekov:
> You're still catching Exception in lines 205 and 259

Yep, didn't mention fixed.

> Another thing I noticed is that the activity still shows up in recents after clicking cancel or save. It might be appropriate to exclude the activity from recents.

Copied activity settings from owncloud's manifest.

PaulAnnekov:
@Nutomic any chance to do a last review?

Nutomic:
Oh it looks like I started a review, but didnt finish it, so you couldnt see my comments. Sorry about that.

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
```diff
@@ -0,0 +1,282 @@
...
+            if (files.size() == 1)
+                files.entrySet().iterator().next().setValue(mShareName.getText().toString());
+            Folder folder = (Folder) mFoldersSpinner.getSelectedItem();
+            new CopyFilesTask(files, folder).execute();
+        });
+        mCancelButton.setOnClickListener(view -> finish());
+    }
+
+    // Taken from ownCloud Android
+    // (GNU GPL v2.0 https://github.com/owncloud/android/blob/master/LICENSE.txt)
```
PaulAnnekov:
```
/**
 * Shares incoming files to syncthing folders.
 *
 * {@link #getDisplayNameForUri} and {@link #getDisplayNameFromContentResolver} are taken from
 * ownCloud Android {@see https://github.com/owncloud/android/blob/79664304fdb762b2e04f1ac505f50d0923ddd212/src/com/owncloud/android/utils/UriUtils.java#L193}
 */
public class ShareActivity extends SyncthingActivity
```
Is it ok?

src/main/java/com/nutomic/syncthingandroid/model/Folder.java
```diff
@@ -60,4 +60,8 @@ public void removeDevice(String deviceId) {
             }
         }
     }
+
+    public String toString() {
+        return label != null ? label : id;
```
PaulAnnekov: Done

src/main/java/com/nutomic/syncthingandroid/model/Folder.java
```diff
@@ -60,4 +60,8 @@ public void removeDevice(String deviceId) {
             }
         }
     }
+
+    public String toString() {
```
PaulAnnekov: Done

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
Nutomic: Yes that looks good.

PaulAnnekov: Fixed.

PaulAnnekov:
@Nutomic Fixed everything.

Nutomic:
Merged, thank you :)

PaulAnnekov:
Great. You're welcome.

PR #853, opened 2017-03-08
Update to Go 1.8, syncthing v0.14.24, fix dns patch for ipv6

AudriusButkevicius:

PR #854, opened 2017-03-09
Fix DNS resolution on IPv6 networks/resolvers

### LINKED ISSUE #817: IPv6 not supported by DNS
Cannot discover devices. The log contains an unsettling comment: 
`dynamic.go:65 ... failed to lookup dynamic relays ... dial tcp: lookup relays.syncthing.net on 1111:2222:3333::33:53: dial udp: too many colons in address 1111:2222:3333::33:53:
`

(It's the DNS server's ipv6 address.)

### Version Information

    App Version: 0.9.2
    Syncthing Version: v0.14.14

uqs:
With IPv6, you must never just append ":port" to a possible IP, as for
IPv6 it must be wrapped in brackets, e.g. [::1]:53

This should fix #817

### DIFF
--- a/patches/golang/all/dns.patch
+++ b/patches/golang/all/dns.patch
@@ -27,7 +27,7 @@
 +		}
 +		output := strings.Trim(string(outputBytes), "\n")
 +		if ParseIP(output) != nil {
-+			servers = append(servers, output + ":53")
++			servers = append(servers, JoinHostPort(output, "53"))
 +		}
 +	}
 +


Nutomic:
Have you seen #853? That should fix the same problem, plus some others.

uqs:
Didn't see it, no. But apparently it's pretty much the same fix, with the problem that it currently doesn't pass the checks and who knows what else is going to fail.

I would appreciate it, if we could merge my hotfix first so that Ipv6 is back in a working state. Updating to a more recent Go seems like a lot more work (which has to be done at some stage anyway).

AudriusButkevicius:
So I suggest you produce yourself a local build with the fix applied if you are in such a hurry, as the fact that you've produced a fix does not imply that the release will go out straight away.

uqs:
I find this an odd approach to things. The currently live version of Syncthing on Android is broken for a large number of users (at least those that don't want to rely on "dynamic"). There is a 1-line patch that is safe and very unlike to break other things.

Shouldn't it be of highest importance to get this fix in and releasing a new, fixed version to the public?

I'm in a hurry only, because I'm dead in the water, am currently switching to a new phone and getting the files sync'ed over would be *really* helpful :)

AudriusButkevicius:
I'd argue that if we release every time there is some fix to something, we'd run out of version numbers. So far we had only a few users complain about this, so you might have wrong assumptions on how widespread this is. As I said, you can always use a local build if you are in a hurry.

Nutomic:
Fixed in #853. And I completely agree with Audrius, if you need a fix this urgent, you will have to compile it yourself.

grinapo:
(I never heard "we'd running out of version numbers" argument before, but I can send you a few numbers in case you're short on them. Also this bug has been rendering syncthing on ipv6 *absolutely broken*: nothing ever connects. All sync happened whenever the device happened to come up on a non-ipv6 network. If ever. Obviously I am not in a hurry, I **can** wait for the next release but it's a **grave** bug nevertheless.)

Nutomic:
It's not "running out of version numbers", it's that we font want to make a new release for every single bug fix. Also, I never heard complaints about missing IPv6 connectivity, so it doesn't seem so urgent. Don't all IPv6 internet connections have dome way to connect to IPv4 addresses?

AudriusButkevicius:
The problem here is that the dnsmasq service on the phone exposes itself via v6, so dns resolution would be broken all together.

uqs:
Yeah, it's not that IPv6 connections fail, it's that somehow syncthing doesn't use the system resolver (I guess because it has to special case "dynamic"??). And instead it looks at the list of configured DNS resolvers and then tries to contact them directly. This fails if all you have are IPv6 resolvers.

I have to check this at my home network, but you have 2 resolvers typically, so if 1 is ipv6 and the other 1 is ipv4, it should try to talk to both of them and at least one should work.

In the end, I disabled ipv6 at my home network temporarily, because "build your own apk" is just a nice way to write "fuck off". It would take me hours to get the SDK and whatnot running and an apk built and then sideloaded onto my phone. Whereas for the project's release engineer it should be a simple press of a button.

Nutomic:
Okay I wasnt aware of this at all. Will make a new release in a bit.

uqs:
So at my home network, dual stack with IPv4 via DHCP and IPv6 via SLAAC and RDNSS (or however the acronyms are spelled), I have this when IPv6 is active:

% adb shell getprop net.dns1
2001:REDACTED::151
% adb shell getprop net.dns2
2001:4860:4860::8844

So even though the device has dual-stack connectivity, it will only ever use IPv6 for DNS resolution, hence the total failure with this bug.

I consider this a short-coming in android, it should keep the primary DNS via DHCP and the primary DNS via RDNSS/SLAAC and fail back accordingly. Oh well, I guess this problem will fix itself eventually over time.


I also tried advertising both an IPv4 and IPv6 address for my local resolver here, but then android only accepts the IPv4 one and ignores the IPv6 one? Or maybe it's a bug in my router advertisements, I didn't sniff them to check whether it simply refuses to send out an IPv4 resolver IP, so android only got the IPv4 via DHCP.

Bloody computers.

In any case, thank you very much for producing a new release and to endure the whining of your users :)

Issue #832, opened 2017-01-10
Use the same description for Master / Sent Only


PanderMusubi:
Use the same description in the app for Master or no Master as in the webgui (on smartphone or computer) which is Sent Only or Sent & Receive. These different names confuse users.

### Version Information

    App Version: 0.9.3 (installed from F-Droid)
    Syncthing Version: v0.14.18
    Android Version: Android 6.0.1 / Cyanogenmod 13.1.2

PanderMusubi:
See also https://github.com/syncthing/syncthing/issues/3889

super9mega:
what would you change it to? i see the file that you would change is 
[https://github.com/syncthing/syncthing-android/blob/master/src/main/res/values/strings.xml](url)
but you have to change it to a single name, would it be called "receive?"

super9mega:
resolved #832 by #878

PanderMusubi:
Exactly the same 1-of-n selection as in the GUI (now two options but will probably be 3):
- Send only (Master)
- Receive only
- Send & Receive

PR #878, opened 2017-04-20
changed Folder Master to 'Receive Changes?'

super9mega:
made naming more universal across all versions, rather then use confusing terminology used no where else


super9mega:
in order to solve issue #832 at [https://github.com/syncthing/syncthing-android/issues/832](url)

Nutomic:
Syncthing calls it `Send Only`. Its probably better if we use the same term.

super9mega:
oops sorry, should be correct now

Nutomic:
Thanks :)

PR #883, opened 2017-04-30
Fix missing path separator in folder copy target

anrp:
When a file is shared to the application, the destination path is missing a trailing / (as tested by out of the box use with the GUI to create and share a folder). So, when a file is copied in to a share directory (example /sdcard/Sync), the file is copied instead to "/sdcard/Sync${SOURCEFILENAME}" instead of "/sdcard/Sync/${SOURCEFILENAME}".


Nutomic:
I think the cleaner solution would be `new File(mFolder.path, entry.getValue());`. Then you can also get rid of the variable `outFile`. Thanks for finding this!

Nutomic:
Thanks :)

PR #892, opened 2017-05-22
Building syncthing using android ndk toolchain

somearth:
**There are some changes in system requirements with these changes:**

1. **Standalone Android Toolchain** - The user should have installed the standalone Android toolchain (using Android NDK) and generated  the Android cross compiler for C.

Please follow the [steps described here](https://gist.github.com/calmh/20c24afc283656695b76f7822c8c8997) by @calmh to generate the Android toolchain

2. **Latest GO** - The user should have a working latest installation of go


AudriusButkevicius:
What it's worth, perhaps we should keep the old build method and print a big ass warning in case it's not built that way?

AudriusButkevicius:
Also I wonder if the tool chain looks the same on Windows.

somearth:
The previous build method builds `go` for all architectures (`ARM`, `ARM64` and `x86`), which is not really required to create `Syncthing` executable for `Android+ARM/ARM64`. Haven't tried for `Android+x86` though. Do we have users using `Android` on `x86` architecture?

Zillode:
The F-Droid build server does not have Go. Therefore `./make-go.bash` cannot be removed.
Yes,

Nutomic:
@Zillode I [asked about this some time ago](https://gitlab.com/fdroid/fdroidserver/issues/215), and apparently they can provide Golang on the F-Droid build server, as long as that version is in Debian backports. But yes, we need to make sure that this new build process works in F-Droid.

For testing, I would just make a beta release via Google Play.

Zillode:
True, but it'll never be fully up to date. Given that upstream prefers to pin to recent Go versions, I'd actually keep it building it (unless this causes actual problems)?

Nutomic:
Well it depends which version of Go is required. Right now 1.7 would be available on F-Droid, so if we can stay behind by one or two releases, that should be fine. Otherwise, we still need to build it from source.

somearth:
Currently I am using syncthing compiled with `go1.7.5 linux/amd64` without any issues on android.

somearth:
Also, will it be a good idea to directly host the `Syncthing so` files and just `wget` them in the buildscript?

Zillode:
I like the DNS fixing, so this PR should be merged.
However, it's just a little curious that it changes a part of the build script that already works with no clear benefits? Your call to accept or reject this part.

> Also, will it be a good idea to directly host the Syncthing so files and just wget them in the buildscript?

No, why would you do this? The idea is that we do not trust compiled binaries and we can build "all" source from scratch. If you want to speed up the building process (why? you only need to do this once), you should add a make-quick.sh script.

make-syncthing.bash
```diff
@@ -7,15 +7,19 @@ JNIDIR="/src/main/jniLibs"
 
 case "$1" in
     arm)
-        export CGO_ENABLED=0
-        export GOOS=linux
+        export CGO_ENABLED=1
+	# The executable "arm-linux-androideabi-gcc" can be found in the bin folder of your standalone android ndk toolchain
+	export CC=arm-linux-androideabi-gcc
+        export GOOS=android
```
AudriusButkevicius: Indentation.
Also, what is that toolchain is not available?

We should check the existence of the toolchain and print a friendly error.

make-syncthing.bash
```diff
@@ -7,15 +7,19 @@ JNIDIR="/src/main/jniLibs"
...
         export GOARM=5
         export TARGETDIR=$MYDIR$JNIDIR/armeabi
         ;;
     arm64)
-        export CGO_ENABLED=0
-        export GOOS=linux
+        export CGO_ENABLED=1
+	# The executable "aarch64-linux-android-gcc" can be found in the bin folder of your standalone android ndk toolchain
+	export CC=aarch64-linux-android-gcc
+        export GOOS=android
```
AudriusButkevicius: See above.

make-syncthing.bash
```diff
@@ -7,15 +7,19 @@ JNIDIR="/src/main/jniLibs"
...
         export GOARCH=arm64
         unset GOARM
         export TARGETDIR=$MYDIR$JNIDIR/arm64-v8a
```
AudriusButkevicius: Also, I think we should build x86 with the android toolchain too, otherwise we'd ship a broken release for some.

make-all.bash
```diff
@@ -1,12 +1,6 @@
 #!/bin/bash -e
 
 # Build the syncthing library
-
-./make-go.bash arm
```
AudriusButkevicius: I still think we should bootstrap latest go as part of the build, as this way we become reliant on Debian to release go fast enough to pick up security fixes and performance improvements.

calmh: Are the trust issues such that a binary download of https://storage.googleapis.com/golang/go1.8.3.linux-amd64.tar.gz directly doesn't work? This is presumable the same place you download the source from anyway?

calmh:
Would it be worth it to separate the f-droid "paranoid" build process from the "normal" build process? We could have a docker image that builds the app to apks etc quickly and simply for every push on the build server. Rebuilding go for every build and so on seems pointless though.

AudriusButkevicius:
That is true, we could just download go as part of the build (if its unavailable at the version we want) and be done with it.

somearth:
I have added a new commit adding support for x86 architecture as well. However, I cannot test this since I don't have such a machine available. Hence, I have kept it commented for now. I have left instructions on what needs to be done in the file itself. Let me know if anything else needs to be done.

make-all.bash
Zillode: Yes, there is no guarantee that the source corresponds to the given binary. This is a guarantee F-Droid wishes to provide. The apk & source can be downloaded for every application. The only binary upstream they trust is Debian (as far as I know).

Nutomic:
Sorry I was kind of busy and didnt have time to check this earlier. 

For the NDK, it seems I have to add `toolchains/aarch64-linux-android-4.9/prebuilt/linux-x86_64/bin/` and other folders to my PATH, right? It would be easier if the script could instead use `NDK_HOME` (or similar), and find the gcc binaries from there.

Even with the PATH set, the build failed for me with this error:
```
$ ./make-syncthing.bash 386
~/workspace/syncthing-android/ext/syncthing/src/github.com/syncthing/syncthing ~/workspace/syncthing-android

GOPATH is /home/felix/workspace/syncthing-android/ext/syncthing
GOPATH is /home/felix/workspace/syncthing-android/ext/syncthing
runtime/internal/sys
go install runtime/internal/sys: mkdir /usr/lib/go-1.7/pkg/linux_386: permission denied
exit status 1
exit status 1
```
Running on Ubuntu 17.04, Go 1.7.4 installed via apt-get. 

Regarding x86 support, please uncomment that code so we can make sure it builds. Once ARM works and this PR is merged, I will make a beta release to make sure x86 also works correctly.

@calmh I think we could use a bootstrap step, like we do in the current build. So if there is a high enough Go version installed, we just use that. If not (eg on F-Droid), the installed Go version is used to build the latest version from source.

Zillode:
> Would it be worth it to separate the f-droid "paranoid" build process from the "normal" build process?

That's the thing, you only need to run the build script _once_. The application development happens using the gradle file with is automatically handled by Android Studio. Thus it merely costs you a few minutes the first time you check out this repository or when we update the Syncthing version.

somearth:
@Nutomic I have uncommented the x86 part of the script.
Also, the error should go away if you run the build as `sudo`.

Nutomic:
Why does the build script need root?

somearth:
I am not sure, but `go` might be doing some one time changes in its installation. It may be because of the following properties that have changed in the build script:
```
export CGO_ENABLED=1
export CC=$x86AndroidGcc
```

somearth:
It gave me the exact same error on first time run. It works perfectly fine though on subsequent runs even without sudo. Following is the output for that:
```
[ec2-user@ip-172-31-28-233 syncthing-android]$ ./make-syncthing.bash 386
~/batnpad/syncthing-android/ext/syncthing/src/github.com/syncthing/syncthing ~/batnpad/syncthing-android
GOPATH is /home/ec2-user/batnpad/syncthing-android/ext/syncthing
GOPATH is /home/ec2-user/batnpad/syncthing-android/ext/syncthing
github.com/syncthing/syncthing/cmd/syncthing
gometalinter is not installed
~/batnpad/syncthing-android
Build Complete
```

calmh:
On 28 May 2017, at 20:58, Zillode <notifications@github.com> wrote:
> 
> Thus it merely costs you a few minutes the first time you check out this repository or when we update the Syncthing version.

I was envisioning the build server building the app for the master branch of both syncthing-android and syncthing, like the build server does today for Syncthing.

Nutomic:
I don't think that's necessary. Syncthing is quite stable so regressions are unlikely, and I also test the app every time I upgrade Syncthing. Besides, the tests for sync things android aren't even working at this point.

Nutomic:
Okay I tried the build again after adding `sudo` in front of lines 70 and 71 and using Go 1.8.1, but I got another error:
https://gist.github.com/Nutomic/53eda4f917ce94dedbb32f9d94029e27

Any idea what that means?

somearth:
I think the variables `CC` and `CGO_ENABLED` are not getting used here, since you did not run the entire script as `sudo`.

Please try running the entire script as sudo once.

imsodin:
I assume the root privilege problem is the same that I encountered earlier when cross-compiling something with go: Internal packages (which need to be compile once when cross-compiling) are always installed to $GOROOT, not $GOPATH. Moving $GOROOT to a path writable without root privileges should resolve the problem.

Nutomic:
I still don't have much luck with this, getting another random error:

https://gist.github.com/Nutomic/ef1cea66b9b78aab8b01e9f19910b5a4

@somearth Regardless, there are a few things in this PR that should be changed:
1. bring make-go.bash back, which should be used if the installed Go version is too old
2. have make-syncthing.bash detect NDK binaries from $NDK_HOME env var, so we dont have to add all the subfolders to $PATH
3. set the env vars as suggested by @imsodin so that we dont need root to compile

AudriusButkevicius:
I'd also vouch for falling back to old build if ndk not available.

Nutomic:
@AudriusButkevicius I think that depends how complex it would be. If we have to keep patching the Go source, I'd say its not worth the effort.

somearth:
@Nutomic Will make these changes.

@Zillode I am not very clear on what you meant in your comment on hosting `Syncthing` binaries for `Android`.

1. What is the trust issue in downloading a binary over SSL from the same place where a developer is downloading the source from?

1. A developer downloading `syncthing-android` source, is not looking for `Syncthing` source in most cases. Why do we want to build it from scratch for everybody? If we are supporting 3 architectures for android right now, `arm`, `arm64`,`386`, why can't we just host the 3 binaries? Won't it be better to provide instructions to build `Syncthing` for `Android` separately.

Zillode:
We can add a script that downloads Syncthing or Go binaries, I have no problem with that. However, the existing full build process should remain intact for our F-Droid release.

> Yes. In all but a very few cases (for technical or historical reasons), we build all applications directly from the source code. This guarantees that the source code is available for the version people have installed.

https://f-droid.org/wiki/page/Deterministic,_Reproducible_Builds
https://wiki.debian.org/ReproducibleBuilds/About

Verifiable trust is an important aspect of F-Droid, as well as Syncthing.

calmh:
I think there are multiple issues conflated here. I would like the Android app to built with GOOS=android because I would like

- to take advantage of any hacks and compatibility tweaks that may be there (this is theoretical, but I'm thinking of things like workarounds for odd filesystem / syscall behavior etc)

- to run on the officially correct platform so we can file bugs on broken Android behavior that should be abstracted away by the Go library

- to not have to run a patched Go, because it's cleaner and see above about being supported and reporting bugs

- to get access (hopefully?) to the CA bundle and be able to remove hacks around using InsecureSkipVerify for usage reporting

- to get correct usage reporting and not bundle everything under "Linux".

And probably other things I can't think of at the moment. Given that you wouldn't need to patch Go it would seem simpler to use a binary distribution of Go, but I don't particularly care.

Certainly build Syncthing from source. Anything else would be odd.

A secondary goal could be to offer an APK with latest-master syncthing-android and latest-master syncthing, or at least latest-release syncthing. People have asked for this for testing and it doesn't immediately make sense to me that we are *not* at least *building* the latest-greatest of everything for ourselves. But I don't really have a horse in that race either.

Nutomic:
@calmh There is continuous integration at [Travis CI](https://travis-ci.org/syncthing/syncthing-android), but that doesn't provide apk downloads unless we manually upload the apks to another server. We could also start building syncthing-android on our own build server again, as long as there is a volunteer to administrate that. But this discussion belongs in a new issue or on the forum.

Nutomic:
@somearth Are you still planning to work on this? It's definitely something we want to do.

Issue #718, opened 2016-08-05
Support Staggered Versioning


scienmind:
Seems like the native in-app UI doesn't support all versioning options..
1. Enable "Staggered Versioning" via WebUI 
2. Restart to apply config.
3. Review folder's details in-app

Result: "Keep Versions = off"
Expected result: something more reflective of the actual config in place
### Version Information

```
App Version: 0.8.1
Syncthing Version: v0.14.3
Android Version: Android 5
```

jmintb:
As fare as I can tell there isn't a framework currently in place to get the versioning from the config file. Is this correct or am I missing something?
Not to say that it is not possible to get the versioning from the config, just that it hasn't been done yet.

jmintb:
I have started adding functionality that would support all the types of file versioning, I just remembered that this issue only mentions staggered file versioning. 
Should I just stick to adding staggered file versioning or can I add an interface to choose between all the types of file versioning and configuration options?

Nutomic:
I guess it might be easier if you do only staggered versioning at first. But thats your decision. It would be great to have all types of versioning supported :)

scienmind:
I filed this issue when I discovered that staggered versioning (that I use most often) is missing from the UI.
There was no intention no imply that "only" staggered versioning should be implemented. 

Basically - if it is available in the web-ui, it should be available in the native UI as well :-)

jmintb:
Question, if Syncthing is not given a parameter like amount of versions to keep for simple versioning, will it revert to a default value or just not work?

jmintb:
Btw I forgot to mention that I chose to continue working on functionality that would support all versioning types. I have a version which works now, should I post an image of what the UI looks like before submitting a pull request?

scienmind:
> should I post an image of what the UI looks like before submitting a pull request?

it's always useful for any kind of UI-altering PRs

Nutomic:
You can just open a PR with the code you have now, and attach screenshots in the PR.

PR #896, opened 2017-05-29
Added support for all versioning types in the app UI. Addressing #718

jmintb:
The UI consists of a dialog fragment with a spinner to select the file versioning type and a scroll view containing a frame layout and finish button. depending on the selected versioning type a fragment containing the appropriate configuration options replaces the frame layout.

I used the descriptions from the web UI, maybe we should shorten them a bit to avoid scrolling so much in the staggered versioning configuration, although I think it is a good idea that the concept of staggered versioning is properly explained. Perhaps the description can be collapsed and expanded if the user wants to read it.

I created fragments for the number picker and folder path view to avoid duplicating code.

screenshot_20170529-225637
screenshot_20170529-225611
screenshot_20170529-225625
screenshot_20170529-225629
screenshot_20170529-225633


Nutomic:
Thanks for this! 

The functionality seems good, but the dialogs seem quite odd. Why aren't you using the default title and buttons for the dialogs? I think you need to use `onCreateDialog()` for that, instead of `onCreateView()` ([documentation](https://developer.android.com/reference/android/app/DialogFragment.html#AlertDialog)).

jmintb:
I agree that the dialogs seem a bit odd. The reason I havn't used the default title and buttons is I needed the custom xml layout to contain the framelayout in order to switch between fragments. When I tried onCreateDialog() it gives me an error "java.lang.IllegalStateException: Fragment does not have a view". The problem is described in this stackoverflow [post](https://stackoverflow.com/a/21418952/6639408) Should I try to add the default title and buttons to the custom dialog?

Nutomic:
Sorry I'm not familiar with DialogFragment, so I'm not sure what the problem is. Are you sure you are using the class correctly?

If DialogFragment doesnt work, you could also use an AlertDialog, or even an Activity with dialog style.

jmintb:
An Activity with the dialog style might fix the issue. I will give that a go. I think I have figured out how to apply the default dialog style to the custom dialog. I will upload pics when I have something to show :)

Nutomic:
You probably want to use `startActivityForResult()`, and return the selection with `setResult()` in that case ;)

jmintb:
Yes, that would be the way to go :). Here's how the Activity with a dialog style looks combined with a little themeing of the layout file: 



screenshot_20170606-121931

Nutomic:
Android dialogs usually have the buttons on the bottom right, not centered (but it's not that important). Otherwise it looks good!

jmintb:
I can fix that. :)

src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
```diff
@@ -460,4 +425,64 @@ private Dialog createDiscardDialog() {
...
+                break;
+            case "staggered":
+                int maxAge = getMaxAgeInDays();
+                description = getString(R.string.type) + ": " + getString(R.string.staggered) + "\n"
+                        + getString(R.string.maximum_age) + ": " + (maxAge == 0 ? getString(R.string.never_delete) : maxAge) + "\n"
+                        + getString(R.string.versions_path)+ ": " + mFolder.versioning.params.get("versionsPath");
+                break;
+            case "external":
+                description = getString(R.string.type) + ": " + getString(R.string.external) + "\n"
+                        + getString(R.string.command) + ": " + mFolder.versioning.params.get("command");
```
Nutomic: You shouldn't concatenate strings in Java like that, it can mess up translations (pretty sure lint will show a warning for it). Instead, you should use [string format parameters](https://stackoverflow.com/a/3656409), and I would make seperate TextViews for `Type` and type specific info.

So you would have `type = getString(R.string.type_simple)` for the first line, and `getString(R.string.type_simple_info, keepVersions)` for the second line.

src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
```diff
@@ -460,4 +425,64 @@ private Dialog createDiscardDialog() {
...
+        }
+
+        return description;
+    }
+
+    //MaxAge is stored in seconds, since Syncthing requires it in seconds, but it is displayed in days. Hence the conversion below.
+    private int getMaxAgeInDays() {
+        int secInADay = 86400;
+
+        return Integer.valueOf(mFolder.versioning.params.get("maxAge"))/secInADay;
```
Nutomic: You can just use `TimeUnit.SECONDS.toDays()`.

src/main/java/com/nutomic/syncthingandroid/fragments/FolderPathTextView.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.nutomic.syncthingandroid.R;
+import com.nutomic.syncthingandroid.activities.FolderPickerActivity;
+
+/**
+ * Simply displays a clickable view that allows for selecting and displaying a folder path.
+ * The selected path can be accessed by adding a textwatcher in the parent activity/fragment to fragment_directory_text_view.
+ */
+
+public class FolderPathTextView extends Fragment {
```
Nutomic: If it extends Fragment, the class name should also end with *Fragment.

Nutomic: Actually, I think you can just extend TextView directly for what you're doing here.

src/main/java/com/nutomic/syncthingandroid/fragments/NumberPickerFragment.java
```diff
@@ -0,0 +1,39 @@
...
+    @Nullable
+    @Override
+    public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container, @Nullable Bundle savedInstanceState) {
+        mNumberPicker = (NumberPicker) inflater.inflate(R.layout.numberpicker_fragment, container, false);
+        mNumberPicker.setWrapSelectorWheel(false);
+
+        return mNumberPicker;
+    }
+
+    public void setValueChangeLisenter(NumberPicker.OnValueChangeListener onValueChangeListener){
```
Nutomic: typo :p

src/main/java/com/nutomic/syncthingandroid/fragments/dialog/StaggeredVersioningFragment.java
```diff
@@ -0,0 +1,103 @@
...
+        int maxAgeInSeconds = Integer.valueOf(mArguments.getString("maxAge"));
+        return convertToDays(maxAgeInSeconds);
+    }
+
+    private int convertToSeconds(int days) {
+        return days * mSecInADay;
+    }
+
+    private int convertToDays(int seconds){
+        return seconds != 0 ? seconds/mSecInADay : 0;
```
Nutomic: Same thing about days/seconds conversion.

src/main/java/com/nutomic/syncthingandroid/fragments/dialog/VersioningDialogFragment.java
```diff
@@ -0,0 +1,119 @@
...
+public class VersioningDialogFragment extends android.support.v4.app.DialogFragment {
+
+    View mView;
+    Spinner mVersioningTypeSpinner;
+
+    Fragment mCurrentFragment;
+
+    VersioningDialogInterface mVersioningDialogInterface;
+
+    List<String> mTypes = Arrays.asList( "none", "simple", "trashcan", "staggered", "external");
```
Nutomic: These members should probably all be private (same on other classes).

src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
jmintb: Okay, i thought that when I used getString(), that it would get the correct transalation. :(

jmintb: Is it for simplicity that two TextViews are better then one, or is there a technical reason? 
Just curious :)

src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
jmintb: In hindsight it seems obvious that there would be a TimeUnit tool. :+1:

src/main/java/com/nutomic/syncthingandroid/fragments/FolderPathTextView.java
jmintb: Would there be advantage to extending TextView instead of using a Fragment?

src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
jmintb: Is it alright to put \n newline in the string resource?
`<string name="staggered_versioning_info">Maximum Age: %1%s \nVersions Path: %2%d</string>`

Nutomic: You will get the correct translation, but lint should  show a warning if you do string concatenation in code like that, probably because other languages might have the text in a different order. You can run `./gradlew lint` on command line to check, though it might not catch it because its not assigned directly to a TextView.

Nutomic: Not really a technical reason, just seems cleanest to have one string resource for one TextView. And \n in string resources is fine :)

src/main/java/com/nutomic/syncthingandroid/fragments/FolderPathTextView.java
Nutomic: It should be simpler to use, and probably be better for performance.

jmintb: When I extend TextView I can call startActivityForResult(), but I cant find a way to override OnActivityResult. is it still possible to override OnActivityResult, if you extended TextView?

jmintb:
Should I squash the commits?
Sorry for all the questions, I am still quite new to all of this. :)

Nutomic:
Ah you will have to move the intent handling to the activity, it doesnt really belong in a view.

No need to squash, Github can do that automatically when merging :)

Nutomic:
I just tried it, and the dialog animations are kind of choppy when you first open the dialog. It would probably be smother if you call `updateFragmentView()` from `onCreate()` (and probably ignore `onItemSelected()` if the selection didnt change).

jmintb:
I agree that the intent handling does not really belong in a view, but then I think it makes more sense just 
 to use a normal view, or keep the fragment. What do you think?

Nutomic:
I would just put it into the activity. You can also put some static helper methods into `FolderPickerActivity`, eg `public static Intent createIntent(Context context, String currentPath)`.

jmintb:
The FolderPickerFragment is no more :).
Was the lag fixed on your phone?

Nutomic:
Yes this is great, thanks :)

Issue #790, opened 2016-11-26
Show device QR codes


JohnVeness:
Although we can show device QR codes in the web GUI, it would be nice (more discoverable, fewer taps) in we could also do so in the native GUI.

wweich:
You already need Barcode Scanner to add a device by scanning that QR code. You can share the device ID (button in the side menu) with Barcode Scanner to show your device ID as QR code.

JohnVeness:
Ah yes, so you can. Maybe not quite as intuitive as a button labelled "Show QR", but good enough.

licaon-kter:
Why is this closed? Can't a view be opened directly to the WebUI QR code? Say add a button near copy ID that opens that from the main interface.

Nutomic:
Its possible, but I'm not sure if its worth the time and screen space, when you can already access the qr code via Barcode Scanner. That said, we can definitely keep the issue open.

licaon-kter:
Screen space? What's the more probable action, share the ID by string or by QR?

Also, i'd vote for leaving only the Share ID entry (remove the current id and copy icon from top on slide menu), that opens up a view like the WebUI does, with string on top (add copy/share icons here) and a QR below.

Nutomic:
Right, it makes sense to have a dialog with QR code + device iD (basically what Syncthing does in "show ID").

JohnVeness:
Thanks for considering this. To be clear, for me this would be useful for showing the QR for both This Device and any Remote Devices. I was wishing to add a remote device that was already known by Android device 1, to Android device 2.

jmintb:
It should be possible to get the QRCode image from the web UI with a Get request, that is how Syncthing-gtk does it. I will start working on it. :)

Nutomic:
Its probably easier if you use ZXing to generate the barcode directly:

https://stackoverflow.com/questions/41606384/how-to-generate-qr-code-using-zxing-library

jmintb:
I have got it working with the Get request, but I can switch to ZXing if you want? :)

Nutomic:
If you already implemented it, and loading times aren't too long, that's fine. Can you make a PR?

jmintb:
I will soon, I am just reviewing the code. The image is loaded instantly.

PR #912, opened 2017-06-28
Added dialog displaying device ID and QRCode addressing #790

jmintb:
Instead of showing the device ID in the drawer, it is shown in a dialog with the QRCode. The QRCode is generated by using a Get request to send the device ID to syncthingURL/qr/?="deviceID" and getting the QRCode as a bitmap in the response.
The device ID was moved to the dialog with the QRCode since it makes sense in my opinion to have the device ID and QRCode displayed in the same place, in the same way that the web UI displays them in the same dialog. Potentially the share device ID option can also be moved to this dialog.

Edit: Added the share device ID option to the dialog.

screenshot_20170629-011049
screenshot_20170629-011334

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/activities/MainActivity.java
+++ b/src/main/java/com/nutomic/syncthingandroid/activities/MainActivity.java
@@ -337,6 +337,13 @@ public class MainActivity extends SyncthingActivity
         return mDrawerToggle.onOptionsItemSelected(item) || super.onOptionsItemSelected(item);
     }
 
+    public void showQrCode() {
+        new AlertDialog.Builder(this)
+                .setMessage("Qr code")
+                .create()
+                .show();
+    }
+
     /**
      * Handles drawer opened and closed events, toggling option menu state.
      */
--- a/src/main/java/com/nutomic/syncthingandroid/activities/MainActivity.java
+++ b/src/main/java/com/nutomic/syncthingandroid/activities/MainActivity.java
@@ -375,13 +375,22 @@ public class MainActivity extends SyncthingActivity
     private View createQrCodeView(String deviceId, Bitmap qrCode) {
         View qrCodeDialogView = this.getLayoutInflater().inflate(R.layout.dialog_qrcode, null);
         TextView deviceIdTextView = (TextView) qrCodeDialogView.findViewById(R.id.device_id);
+        TextView shareDeviceIdTextView = (TextView) qrCodeDialogView.findViewById(R.id.actionShareId);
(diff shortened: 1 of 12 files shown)


licaon-kter:
Nice, moving Share here is better too.

jmintb:
@licaon-kter I agree.

src/main/java/com/nutomic/syncthingandroid/activities/MainActivity.java
```diff
@@ -332,6 +359,40 @@ private Dialog createRestartDialog(){
...
+        TextView shareDeviceIdTextView = (TextView) qrCodeDialogView.findViewById(R.id.actionShareId);
+        ImageView qrCodeImageView = (ImageView) qrCodeDialogView.findViewById(R.id.qrcode_image_view);
+
+        deviceIdTextView.setText(deviceId);
+        deviceIdTextView.setOnClickListener(v -> Util.copyDeviceId(this, deviceIdTextView.getText().toString()));
+        shareDeviceIdTextView.setOnClickListener(v -> shareDeviceId(deviceId));
+        qrCodeImageView.setImageBitmap(qrCode);
+
+        return qrCodeDialogView;
+    }
```
jmintb: I thought that it would improve the readability and maintainability of the code.

src/main/res/layout/dialog_qrcode.xml
```diff
@@ -0,0 +1,45 @@
...
+
+    </LinearLayout>
+
+    <ImageView
+        android:id="@+id/qrcode_image_view"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center" />
+
+</LinearLayout>
```
jmintb: Should I copy the layout you made into my code ?

src/main/java/com/nutomic/syncthingandroid/activities/MainActivity.java
Nutomic: I think splitting it into 3 methods actually makes it more complicated, because you have to figure out which one is actually used, and how (especially as the names are all really similar).

src/main/res/layout/dialog_qrcode.xml
Nutomic: Yes please :)

src/main/java/com/nutomic/syncthingandroid/activities/MainActivity.java
jmintb: After combining them into one function I agree. :)

Nutomic:
Great, thanks :)

Issue #911, opened 2017-06-26
Outgoing Limit Rate setting being immediately forgotton


mlda065:
# Steps to reproduce:

1. Open  the app
2. Go to `Settings` -> `Syncthing Options` -> `Outgoing Rate Limit (KiB/s)`

<img src="https://user-images.githubusercontent.com/7035647/27532924-27bba3fe-5aa5-11e7-92c1-0d251d1bcb6a.png" width="200">

3. Set the value to something positive, e.g. 200

<img src="https://user-images.githubusercontent.com/7035647/27532930-2b94a28c-5aa5-11e7-8e4d-02df9a7946dd.png" width="200">

4. Click Ok
5. When asked to restart the app, click 'Restart Now'

<img src="https://user-images.githubusercontent.com/7035647/27532939-30518344-5aa5-11e7-918d-e80c855167f2.png" width="200">

<img src="https://user-images.githubusercontent.com/7035647/27532943-3297c582-5aa5-11e7-8200-30d2e822d7ea.png" width="200">

6. Go back and look at what the current `Outgoing Rate Limit` is

# Expected behaviour:

After that final step, I should see `200KiB/s` in that field.

<img src="https://user-images.githubusercontent.com/7035647/27532930-2b94a28c-5aa5-11e7-8e4d-02df9a7946dd.png" width="200">

If I set the `outgoing rate limit` to be 200KiB/s, it should stay 200KiB/s. 

# Observed behaviour

The `Outgoing rate limit` is back at 0KiB/s.

<img src="https://user-images.githubusercontent.com/7035647/27532924-27bba3fe-5aa5-11e7-92c1-0d251d1bcb6a.png" width="200">

I think that this is not just an issue with *displaying* the saved value. I think this is also a bug with *enacting* on the value, and actually rate limiting.

### Version Information

    App Version: 0.9.5
    Syncthing Version: v0.14.18
    Android Version: Android 7.0, Samsung S7, Stock ROM

mlda065:
I think the issue is due to [this line](https://github.com/syncthing/syncthing-android/blob/4e8142b6b8e4ef3e1edf1b31ecc19c702ceb067d/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java#L236). I'll try to build and test it when I get a chance.

Nutomic:
You're right @mlda065, that's clearly a typo. Do you want to make a PR?

PR #916, opened 2017-07-05
fixed typo with send/recv rate limit

mlda065:
To fix issue #911 

Given how simple the change is, we can be pretty confident that the patch works, by inspection.
However I haven't actually compiled and run this to check that it works. 

I'm quite new to Android, and I couldn't get the current master branch to work on my phone. It did compile, but I got an error message at runtime about not being able to create a config file. I gave up trying to physically run this version on my phone, but I'm pretty confident.


Nutomic:
Yes it works, thanks :)

To build the app from source, you have to [compile Syncthing first](https://github.com/syncthing/syncthing-android#building)

PR #920, opened 2017-07-14
Use API key as password for web GUI (fixes #868)

### LINKED ISSUE #868: After uninstalling-reinstalling and importing previous settings Web GUI fails to load
Steps to reproduce the problem: 

1. I have exported settings
1. Deleted v0.9.6 (which was installed from F-Droid)
1. Installed v0.9.8 (from Play Store)
1. Started application for the first time
1. Certificates are generated
1. Imported settings

After this procedure everything works as it should except for the Web GUI. When I select from the menu, the web page never loads. I strongly suspect that it has something to do with the https certificate. I can open the Web GUI from a browser.

App Version: 0.9.8
Syncthing Version: v0.14.25
Android Version: Android 6.01 / MIUI 8.2.1

Unrud:
A random password can't be retrieved after export and import of settings.

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/activities/WebGuiActivity.java
+++ b/src/main/java/com/nutomic/syncthingandroid/activities/WebGuiActivity.java
@@ -87,9 +87,7 @@ public class WebGuiActivity extends SyncthingActivity
         }
 
         public void onReceivedHttpAuthRequest(WebView view, HttpAuthHandler handler, String host, String realm) {
-            String password = PreferenceManager.getDefaultSharedPreferences(WebGuiActivity.this)
-                    .getString("web_gui_password", "");
-            handler.proceed(mConfig.getUserName(), password);
+            handler.proceed(mConfig.getUserName(), mConfig.getApiKey());
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


Nutomic:
I feel like you're making this to complicated (and breaking existing behaviour). Wouldn't it be enough to check if the password in SharedPreferences is correct, and if not, set a random new password?

Unrud:
> and breaking existing behaviour

I updated the PR, it doesn't break existing behavior anymore. The API key is used as the password for new configurations and existing configurations are updated, if the password is still the stored random password.

If the user has changed the password manually, the configuration is not touched.

It also doesn't repair imported configurations, that are already broken. The user has to set the password in the configuration to the API key manually. We can't do this automatically, because we don't know if the imported password is a old random password or was set deliberately.

> Wouldn't it be enough to check if the password in SharedPreferences is correct,

No, because we don't know why the password doesn't work. Either it's an old random password or the user changed the password deliberately. This would break existing behavior in the same way, as the initial version of this PR.

> and if not, set a random new password?

Using a random password instead of the API key doesn't provide any additional security. It's just more complicated.

Unrud:
I created the new PR #923 with the old approach (It also removes user and password from the configuration activity.) I recommend #923 over this PR, because it also protects old configurations without a password and repairs already broken configurations. Setting a custom user and password doesn't seem very useful anyway.

PR #921, opened 2017-07-14
Check battery saving on Android 5.0

Unrud:
isPowerSaveMode and ACTION_POWER_SAVE_MODE_CHANGED were added in API level 21.


Nutomic:
Thanks for finding all these :)

PR #923, opened 2017-07-17
Remove custom user and password (fixes #868)

### LINKED ISSUE #868: After uninstalling-reinstalling and importing previous settings Web GUI fails to load

Unrud:
This removes the ability to set a custom user and password.
The user is always set to "syncthing" and the password to the API key.
This ensure that the web GUI works and makes sure that the web GUI is protected.

This is an alternative approach to #920.

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
+++ b/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
@@ -65,8 +65,6 @@ public class SettingsActivity extends SyncthingActivity {
         private CheckBoxPreference mRelaysEnabled;
         private EditTextPreference mGlobalAnnounceServers;
         private EditTextPreference mAddress;
-        private EditTextPreference mUser;
-        private EditTextPreference mPassword;
         private CheckBoxPreference mUrAccepted;
 
         private CheckBoxPreference mUseRoot;
@@ -115,8 +113,6 @@ public class SettingsActivity extends SyncthingActivity {
             mRelaysEnabled          = (CheckBoxPreference) findPreference("relaysEnabled");
             mGlobalAnnounceServers  = (EditTextPreference) findPreference("globalAnnounceServers");
             mAddress                = (EditTextPreference) findPreference("address");
-            mUser                   = (EditTextPreference) findPreference("user");
-            mPassword               = (EditTextPreference) findPreference("password");
             mUrAccepted             = (CheckBoxPreference) findPreference("urAccepted");
 
             Preference exportConfig = findPreference("export_config");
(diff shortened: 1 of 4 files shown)


Nutomic:
@syncthing/android Is it okay if we remove the username/password setting, and always use "syncthing" as user and the api key as password? Or should we keep that functionality for any reason?

PaulAnnekov:
I don't see any reason why we shouldn't do this.

Nutomic:
Alright, merged. I will be on vacation for about 3 weeks, so I will release this change afterwards (because I expect some complaints).

scienmind:
Few questions:

How does it impact the ability to connect to ST GUI remotely?
I almost exclusively manage android devices via desktop+WebUI, will it break the workflow?

If not, and it's just "fixed" username and password, which API key is used? Does every "generate new API key" will change the password as well?

And does having an API key as a password degrades security in any way and would you recommend to not map GUI address to 0.0.0.0 for network availability anymore?

Unrud:
> How does it impact the ability to connect to ST GUI remotely?

You have to use the API key as the password.

> Does every "generate new API key" will change the password as well?

Yes.

> And does having an API key as a password degrades security in any way

No, because you can access the web GUI with the API key anyway. (By setting the X-API-KEY header.)

Nutomic:
I noticed that this slowed down the app startup considerably, because you used the default work factor for bcrypt. On my Nexus 5, this took about 25 seconds. I've changed it to a work factor of 4, which takes about 500ms on the same device. The password was also computed more than once at app start, which slowed it down even more.

679841a6acf0046ec6c75efa48083f1b27c7cc06

Nutomic:
@Unrud Actually, even with a work factor of 4, it causes some slowdown in the app start. With the password check, it takes 0.652s from service start until the Syncthing binary is started. Without, it's only 0.370s. I think it would be fine if we just check the password field against a SharedPreference, right?

Unrud:
> I've changed it to a work factor of 4

:+1:

> On my Nexus 5, this took about 25 seconds. I've changed it to a work factor of 4, which takes about 500ms on the same device. 

That can't be right. I measured it on a [Moto E (1st gen)](https://en.wikipedia.org/wiki/Moto_E_(1st_generation)), which is old, very cheap and slower than a Nexus 5. It took 1.2 seconds for a factor of 10 and 85 milliseconds for a factor of 4.

> I think it would be fine if we just check the password field against a SharedPreference, right?

Yes, we can cache the password hash in SharedPreference. But considering my test results, I wouldn't bother. These ancient devices are barely able to run **syncthing** in the first place.

Nutomic:
Hmm that's weird, its definitely way slower here, with this code:
```
Log.d("xxx", "before pw check");
passwordOk = !TextUtils.isEmpty(pw) && BCrypt.checkpw(apikey, pw);
Log.d("xxx", "after pw check");
```
```
00:57:21.977 D/xxx: before pw check
00:57:22.326 D/xxx: after pw check


00:58:53.837 D/xxx: before pw check
00:59:16.106 D/xxx: after pw check
```

Unrud:
I used ``System.nanoTime()`` for measuring it. With your method I get 846 milliseconds for a factor of 10 and 67 milliseconds for a factor of 4.

PR #925, opened 2017-07-18
Show error when folder is out of sync

Unrud:
If a folder is out of sync the app shows that the folder is idle. Only the web GUI shows the problem.
I copied the translations from Syncthing and also colorized the folder status message.


Nutomic:
:+1:

PR #929, opened 2017-07-24
Fixed and added german translations.

Dpunkt:


wweich:
We are using [transifex](https://www.transifex.com/syncthing-android/syncthing-android/) for translations.

PR #930, opened 2017-07-27
Switch to wget instead of heavyweight submodules

fqsrz:


Nutomic:
What's the point of this PR? It's just copied from an old PR.

Issue #899, opened 2017-06-02
"Save to Syncthing" could remember last folder


JohnVeness:
At the moment, every time you go into the "Save to Syncthing" window it defaults to the first listed share. It would be very useful if it could remember the last share that was saved to.

If issue #898 is implemented, this would include remembering the last subfolder saved into.

jmintb:
Should it remember the last subfolder for all of the Syncthing folders or just the most recent one?

JohnVeness:
Good question. I think the average used would expect at the very least by default for it to remember the last subfolder used in the last Syncthing folder used. But yes, it might be a nice addition if, if at this point the user picked a different Syncthing folder, that it remembered the last subfolder used in that Syncthing folder too.

jmintb:
I am also considering adding a button that clears the saved subdirectory so that you can easily get back to the main folder that is being synced, without having to navigate back to the top of the folder hierachy.
What do you think?

JohnVeness:
Well if you're considering how to navigate back up the folder tree, I'm a fan of "clickable breadcrumbs", where a single click can navigate instantly up to any parent or grandparent, up to the top level.

jmintb:
Well I was thinking outsite of the file navigation ui. If you want to share something to the synced folder, but last time you added a file to it you had navigated to a sub directory. Then instead of opening the file browseing ui, you press the button and the saved sub directory is cleared. Something like this: 
Before button press:
![screenshot_20170816-215321](https://user-images.githubusercontent.com/10332534/29382674-a28e2d22-82cd-11e7-8532-c8f7cd3a5bb3.png)

After button press:
![screenshot_20170816-215326](https://user-images.githubusercontent.com/10332534/29382698-bd048264-82cd-11e7-8695-cad9fd30dab4.png)

JohnVeness:
Oh I see. I was rather thinking that as soon as you entered the "Save to Syncthing" window, you would be presented with a folder browser, with a box for the user to modify the filename. You could then browse within the folders, with the top-level being all the Syncthing folders.

I realise this model contradicts my idea of remembering the last used subfolder of each Syncthing folder, because it would seem strange to the user if they jumped further down the tree when clicking on one of the top-level folders.

Nutomic:
The "Clear subfolder" button is way too complicated, and might be hard to understand. Its totally fine to use back navigation in that case.

Issue #898, opened 2017-06-02
"Save to Syncthing" could allow browsing into subfolders


JohnVeness:
At the moment, the "Save to Syncthing" window when sharing from another app only allows you to save in the top-level folder of a Syncthing share. It would be very useful if you could browse and save into a subfolder.

jmintb:
I will try implementing this and #899.

PR #939, opened 2017-08-21
Addressing #899 and #898.

jmintb:
Changes:
Added Subdirectory browsing
The previously selected folder is remembered.
The subdirectory is remembered for each Syncthing folder.
The saved subdirectory in the shared preferences is deleted when the corresponding folder is deleted from Syncthing.
The root directory of the FolderPickerActivity is set to the synced folder, so the user can only choose a sub folder within the folder that is being synced.
TheFolderPickerActivity was modified in order to allow for a custom root directory to be set.

Example with no sub folder selected:
screenshot_20170821-141950

Example with a sub folder selected:
screenshot_20170821-162521

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
+++ b/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
@@ -141,7 +141,7 @@ public class FolderActivity extends SyncthingActivity
         mEditIgnores = findViewById(R.id.edit_ignores);
 
         mPathView.setOnClickListener(view ->
-                startActivityForResult(FolderPickerActivity.createIntent(this, mFolder.path), FolderPickerActivity.DIRECTORY_REQUEST_CODE));
+                startActivityForResult(FolderPickerActivity.createIntent(this, mFolder.path, null), FolderPickerActivity.DIRECTORY_REQUEST_CODE));
 
         findViewById(R.id.versioningContainer).setOnClickListener(v -> showVersioningDialog());
         mEditIgnores.setOnClickListener(v -> editIgnores());
@@ -383,9 +383,6 @@ public class FolderActivity extends SyncthingActivity
                 .setMessage(R.string.remove_folder_confirm)
                 .setPositiveButton(android.R.string.yes, (dialogInterface, i) -> {
                     getApi().removeFolder(mFolder.id);
-                    //Remove saved data from share activity for this folder.
-                    SharedPreferences prefs = this.getSharedPreferences(ShareActivity.SHARED_PREFS_SAVED_DIRECTORIES, MODE_PRIVATE);
-                    prefs.edit().remove(mFolder.id).apply();
                     finish();
                 })
                 .setNegativeButton(android.R.string.no, null)
--- a/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
+++ b/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
@@ -5,7 +5,7 @@ import android.app.AlertDialog;
 import android.app.Dialog;
 import android.content.ActivityNotFoundException;
 import android.content.Intent;
-import android.net.Uri;
+import android.content.SharedPreferences;
 import android.os.Build;
 import android.os.Bundle;
 import android.support.v7.widget.SwitchCompat;
@@ -383,6 +383,9 @@ public class FolderActivity extends SyncthingActivity
                 .setMessage(R.string.remove_folder_confirm)
                 .setPositiveButton(android.R.string.yes, (dialogInterface, i) -> {
                     getApi().removeFolder(mFolder.id);
+                    //Remove saved data from share activity for this folder.
+                    SharedPreferences prefs = this.getSharedPreferences(ShareActivity.SHARED_PREFS_SAVED_DIRECTORIES, MODE_PRIVATE);
+                    prefs.edit().remove(mFolder.id).apply();
                     finish();
                 })
                 .setNegativeButton(android.R.string.no, null)
--- a/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
+++ b/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
@@ -5,7 +5,7 @@ import android.app.AlertDialog;
 import android.app.Dialog;
 import android.content.ActivityNotFoundException;
 import android.content.Intent;
-import android.content.SharedPreferences;
+import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
 import android.support.v7.widget.SwitchCompat;
--- a/src/main/java/com/nutomic/syncthingandroid/activities/FolderPickerActivity.java
+++ b/src/main/java/com/nutomic/syncthingandroid/activities/FolderPickerActivity.java
@@ -33,6 +33,7 @@ import com.google.common.collect.Sets;
 import com.nutomic.syncthingandroid.R;
 import com.nutomic.syncthingandroid.SyncthingApp;
 import com.nutomic.syncthingandroid.service.SyncthingService;
+import com.nutomic.syncthingandroid.util.Util;
 
 import org.w3c.dom.Text;
 
@@ -198,7 +199,7 @@ public class FolderPickerActivity extends SyncthingActivity
                 return true;
             case R.id.select:
                 Intent intent = new Intent()
-                        .putExtra(EXTRA_RESULT_DIRECTORY, mLocation.getAbsolutePath());
+                        .putExtra(EXTRA_RESULT_DIRECTORY, Util.formatPath(mLocation.getAbsolutePath()));
                 setResult(Activity.RESULT_OK, intent);
                 finish();
                 return true;
@@ -310,7 +311,7 @@ public class FolderPickerActivity extends SyncthingActivity
 
     /**
      * Goes up a directory, up to the list of roots if there are multiple roots.
-     *
+     * <p>
      * If we already are in the list of roots, or if we are directly in the only
      * root folder, we cancel.
      */
--- a/src/main/java/com/nutomic/syncthingandroid/activities/FolderPickerActivity.java
+++ b/src/main/java/com/nutomic/syncthingandroid/activities/FolderPickerActivity.java
@@ -13,6 +13,7 @@ import android.os.Environment;
 import android.os.IBinder;
 import android.preference.PreferenceManager;
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.support.v4.content.ContextCompat;
 import android.text.TextUtils;
 import android.util.Log;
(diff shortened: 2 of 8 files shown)


src/main/java/com/nutomic/syncthingandroid/activities/FolderPickerActivity.java
```diff
@@ -48,6 +49,9 @@
     public static final String EXTRA_INITIAL_DIRECTORY =
             "com.nutomic.syncthingandroid.activities.FolderPickerActivity.INITIAL_DIRECTORY";
 
+    public static final String EXTRA_ROOT_DIRECTORY =
+            "com.nutomic.syncthingandroid.activities.FolderPickerActivity.ROT_DIRECTORY";
```
AudriusButkevicius: ROT -> ROOT

src/main/java/com/nutomic/syncthingandroid/activities/FolderPickerActivity.java
```diff
@@ -101,7 +116,8 @@ protected void onCreate(Bundle savedInstanceState) {
     }
 
     /**
-     * Reads available storage devices/folders from various APIs and inserts them into
+     * If a root directory is specefied it is added to {@link #mRootsAdapter} otherwise
```
AudriusButkevicius: specefied -> specified

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
```diff
@@ -52,12 +65,24 @@ public void onApiChange(SyncthingService.State currentState) {
 
         List<Folder> folders = getApi().getFolders();
 
+        //Get the index of the previously selected folder.
```
AudriusButkevicius: Space after //

jmintb:
Thanks for the nitpicking :).

src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
```diff
@@ -360,6 +361,9 @@ private Dialog createDeleteDialog(){
                 .setMessage(R.string.remove_folder_confirm)
                 .setPositiveButton(android.R.string.yes, (dialogInterface, i) -> {
                     getApi().removeFolder(mFolder.id);
+                    //Remove saved data from share activity for this folder.
+                    SharedPreferences prefs = this.getSharedPreferences(ShareActivity.SHARED_PREFS_SAVED_DIRECTORIES, MODE_PRIVATE);
+                    prefs.edit().remove(mFolder.id).apply();
```
Nutomic: it's probably better to handle this in RestApi.removeFolder(), in case that function is called from somewhere else. And you dont need to set MODE_PRIVATE, because that's the default.

src/main/java/com/nutomic/syncthingandroid/activities/FolderPickerActivity.java
```diff
@@ -74,6 +78,17 @@ public static Intent createIntent(Context context, String currentPath) {
         return intent;
     }
 
+    //Creates intent with a specified root directory that is used to limit what directories the user can browse.
+    public static Intent createIntentWithRootDir(Context context, String currentPath, String rootDir) {
+        Intent intent = createIntent(context, currentPath);
+
+        if(!TextUtils.isEmpty(rootDir)) {
+            intent.putExtra(EXTRA_ROOT_DIRECTORY, rootDir);
+        }
```
Nutomic: If the function is called createIntentWithRootDir(), there is no reason to ever call it without a rootDir parameter. So I dont see the reason for this check.

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
```diff
@@ -52,12 +65,24 @@ public void onApiChange(SyncthingService.State currentState) {
 
         List<Folder> folders = getApi().getFolders();
 
+        // Get the index of the previously selected folder.
+        int folderIndex = 0;
+        String savedFolderId = getSharedPreferences(SHARED_PREFS_SAVED_DIRECTORIES, MODE_PRIVATE)
```
Nutomic: Just use PreferenceManager.getDefaultSharedPreferences() instead.

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
```diff
@@ -123,7 +150,34 @@ protected void onCreate(Bundle savedInstanceState) {
             if (files.size() == 1)
                 files.entrySet().iterator().next().setValue(mShareName.getText().toString());
             Folder folder = (Folder) mFoldersSpinner.getSelectedItem();
-            new CopyFilesTask(files, folder).execute();
+            String path = folder.path;
+            if (!TextUtils.isEmpty(mSubDirectoryTextView.getText())) {
+                path += mSubDirectoryTextView.getText();
+            }
```
Nutomic: This seems kind of hacky, would probably be cleaner if you used `getSavedSubDirectory()`.

src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
jmintb: Agreed.

src/main/java/com/nutomic/syncthingandroid/activities/FolderPickerActivity.java
jmintb: I wanted to avoid an empty string "", but I guess that wouldn't actually be a problem.

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
jmintb: In that case, should I use something like saved_sub_directory_(insert folder id here) as the key for the saved sub directories, instead of just the folder id. Just to ensure that the key is unique?

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
jmintb: I just spotted another problem, the folder path ends in a / and the subdirectory starts with a /. One of these should be removed.

src/main/java/com/nutomic/syncthingandroid/activities/FolderPickerActivity.java
Nutomic: Yeah I don't think that's something you should handle really. Alternatively, you could have only one `createIntent()` method, and make the rootDir param nullable. That would also simplify the code a bit (one method less).

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
Nutomic: Yeah that seems easier and less complicated.

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
Nutomic: I think you can just use `File`s instead of strings for them, and it should take care of that automatically.

jmintb: I think I will just use replace, unless you think files would be better?

Nutomic: Yes `File` is definitely better than using strings.

Nutomic:
One thing that I just noticed: if the root of the folder is selected, wouldnt it make more sense to display `/` instead of `browse`? Maybe we could have a field for the path, and a Button with `Select folder`. Also, I think the path should be shown with a trailing slash (right now it has a leading slash).

jmintb:
I think that it looks cleaner and less empty with browse instead of `/`. it would however be more informative to show the `/`, so maybe it does make more sense to show the slash.
I agree that there should be a trailing slash, should the leading slash then be removed?

I will start implementing the separate path field and browse button.

Nutomic:
Hmm, I guess it makes sense to show *something* if no folder is selected. But it could also be confusing, eg if you have a subfolder named `browse`. Yes I would only show a trailing slash for the path (that's how relative paths are usually shown on Linux).

jmintb:
Maybe we could show some sort of hint if no sub folder is selected, like "No sub folder is selected.".

jmintb:
Here is a quick implementation of the separate browse button and sub folder field. I am not sure whether or not the grey folder icon should be kept next to the textview. I havn't fixed the slashes yet.
What do you think?

screenshot_20170829-215019

screenshot_20170829-215007

Nutomic:
Yes that looks better. I would definitely keep the folder icon. For the button, I would put it to the right of the path TextView, and maybe just use the default grey background?

jmintb:
Yes, the default background sounds like a good idea.

jmintb:
There is now a new problem with the UI. When the keyboard is open, then the cancel and save buttons are pushed up, so they are behind the browse button. Any idea of how to deal with this?
I am not sure if we should put them on top of the browse button or if they should be kept below it, but then be hidden behind the keyboard. 
Maybe we could hide the sub directory UI elements while the keyboard is open. 
On phones with a larger screen and higher resolution I do not think this will be an issue, but on my 5 inch 720p display it certainly is. 

Sorry for the long delay, I have just been really busy in the past couple of weeks.

Nutomic:
I think the best option is if you just make the screen scrollable (eg wrap the root layout in a ScrollView). And no worries, just take your time ;)

jmintb:
Making the screen scrollable, introduced other issues relating to the position of the cancel and save button, so I replaced the relative layout with a constraint layout which solved these issues. What do you think?
screenshot_20170920-222231
This below image shows the ShareActivity with the keyboard open, where I have scrolled down to reveal the cancel and share button.
screenshot_20170920-222242

src/main/java/com/nutomic/syncthingandroid/activities/ShareActivity.java
```diff
@@ -272,4 +341,40 @@ protected void onPostExecute(Boolean isError) {
...
+            Folder selectedFolder = (Folder) mFoldersSpinner.getSelectedItem();
+            String folderDirectory = selectedFolder.path;
+            //The folder paths are not entirely consistent in terms of what character they end with,
+            //that is why a / is added if it is missing so it can be used to cut the corresponding part
+            //from the subDirectory string, so only the sub directory is displayed.
+            if (!folderDirectory.endsWith("/")) {
+                folderDirectory += "/";
+            }
+
+            String subDirectory = data.getStringExtra(FolderPickerActivity.EXTRA_RESULT_DIRECTORY) + "/";
```
Nutomic: Instead of messing around with path formatting manually, you can just use [`file.toURI().normalize().getPath()`](https://stackoverflow.com/a/21849556/1837158) to get the properly formatted path. You can put that in a method in `Util` and call it on FolderPickerActivity:200 and ShareActivity:361.

Nutomic:
The layout breaks on smaller screens (tested with Nexus 5, buttons are overlapping). So you will still need a ScrollView in that case. 

screenshot_20170925-135457

jmintb:
The view works fine on small screens now, at least from my testing.

Nutomic:
Thanks :)

PR #948, opened 2017-09-19
Fix permissions of config.xml when Syncthing runs as root

### LINKED ISSUE #812: Root Start Problematic since 0.8.x - syncthing could not create config.XML popup
Do not report security issues in this bug tracker. Instead, contact security@syncthing.net directly - see https://syncthing.net/security.html for more information.

If your issue is a support request ("How do I get my devices to connect?" or similar), please use the support forum at https://forum.syncthing.net/ where a large number of helpful people hang out. This issue tracker is for reporting bugs or feature requests directly to the developers.

If your issue is a bug report, replace this boilerplate with a description of the problem, being sure to include at least:

    what happened,
    what you expected to happen instead, and
    any steps to reproduce the problem.

Also fill out the version information below and add log output or screenshots as appropriate.

If your issue is a feature request, simply replace this template text in its entirety.

### Version Information

    App Version: 0.x.y
    Syncthing Version: v0.x.y
    Android Version: Android x.y / Cyanogenmod x / ...

flipreverse:
Should fix #812


AudriusButkevicius:
So you'll have to fix formatting, and adhere to the style of the rest of the file if you expect this to be merged.

In general, the approach atleast to me looks crude, as it adds sleep 2 and various other hacks.
Isn't there a programmatic approach that does not forking into a subshell?

flipreverse:
@AudriusButkevicius What do you mean by "the style of the rest of the file"?

The sleep might not be necessary. What do you mean by "various other hacks"?
On my opinion, there is no programmatic approach to fix it since you need root privileges to change the se file type and a file's owner. Both steps are necessary for the Android Syncthing Service to read the config file. Otherwise, access will be denied.

AudriusButkevicius:
Style of the rest of the file in terms of indentation, variable names, spaces after commas etc.
Doesn't the superuser library have some API to execute setuid or something like that and then doing a normal chown instead of spawning a subshell and writing out commands to a subprocess...

PR #949, opened 2017-09-19
Provided fix for issue #812

flipreverse:
Check permissions of config.xml before reading or writing from/to it.
If the permissions are wrong, assume root magic is going on, and fix it.

(Please excuse the confusion. Pull request #948 can be deleted.)

### DIFF
--- a/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
+++ b/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
@@ -13,6 +13,7 @@ import android.util.Log;
 import com.nutomic.syncthingandroid.R;
 import com.nutomic.syncthingandroid.service.SyncthingRunnable;
 import com.nutomic.syncthingandroid.service.SyncthingService;
+import com.nutomic.syncthingandroid.util.Util;
 
 import org.mindrot.jbcrypt.BCrypt;
 import org.w3c.dom.Document;
@@ -82,7 +83,7 @@ public class ConfigXml {
     }
 
     private void readConfig() {
-        if (!mConfigFile.canRead() && !checkPermConfigFile()) {
+        if (!mConfigFile.canRead() && !Util.fixAppDataPermissions(mContext)) {
             throw new OpenConfigException();
         }
         try {
@@ -235,7 +236,7 @@ public class ConfigXml {
      * Writes updated mConfig back to file.
      */
     private void saveChanges() {
-        if (!mConfigFile.canWrite() && !checkPermConfigFile()) {
+        if (!mConfigFile.canWrite() && !Util.fixAppDataPermissions(mContext)) {
             Log.w(TAG,"Failed to save updated config. Cannot change the owner of the config file.");
             return;
         }
@@ -249,66 +250,5 @@ public class ConfigXml {
         } catch (TransformerException e) {
             Log.w(TAG, "Failed to save updated config", e);
         }
-    }
-
-    /**
-     * Normally an application's data directory is only accessible by the corresponding application.
-     * Therefore, every file is owned by an application's user and group. When running Syncthing as root,
-     * it writes its config to the application's data directory. This leaves a file owned by root having 0600.
-     * Moreover, writing this file as root changes a file's type in terms of SELinux.
-     * A subsequent start of Syncthing will fail due to insufficient permissions.
-     * Hence, this method fixes the owner, group and the file's type of the config.xml.
-     * 
-     * @return true if the operation was successfully performed. False otherwise.
-     */
-    private boolean checkPermConfigFile() {
-        // We can safely assume that root magic is somehow available, because readConfig and saveChanges check for
-        // read and write access before calling us.
-        // Be paranoid :) and check if root is available.
-        if (useRoot()) {
-            Log.e(TAG,"Root is not available. Cannot fix permssions ");
-            return false;
-        }
-
-        try {
-            ApplicationInfo appInfo = mContext.getPackageManager().getApplicationInfo(mContext.getPackageName(),0);
-            Log.d(TAG,"Uid of '" + mContext.getPackageName() + "' is " + appInfo.uid);
-            Process fixPerm = Runtime.getRuntime().exec("su");
-            DataOutputStream fixPermOut = new DataOutputStream(fixPerm.getOutputStream());
-            String cmd = "chown " + appInfo.uid + ":" + appInfo.uid + " " + mConfigFile + "\n";
-            Log.d(TAG,"Running: '" + cmd);
-            fixPermOut.writeBytes(cmd);
-            // For now, we assume that a file's type should *always* be 'u:object_r:app_data_file:s0:c512,c768'.
-            // At least for those files residing in an application's data folder.
-            // Running syncthing as root, makes it 'u:object_r:app_data_file:s0'.
-            // Simply reverting the type to its default should do the trick.
-            cmd = "restorecon " + mConfigFile + "\n";
-            Log.d(TAG,"Running: '" + cmd);
-            fixPermOut.writeBytes(cmd);
-            fixPermOut.writeBytes("sleep 2\n");
-            fixPermOut.writeBytes("exit\n");
-            fixPermOut.flush();
-            if (fixPerm.waitFor() == 0) {
-                Log.d(TAG,mConfigFile + ": Can read? " + mConfigFile.canRead() + ", Can write? " + mConfigFile.canWrite());
-                Log.i(TAG,"Successfully changed the owner of the config file.");
-                return true;
-            } else {
-                return false;
-            }
-        } catch (IOException | InterruptedException e) {
-            Log.w(TAG,"Cannot chown config file",e);
-        } catch (NameNotFoundException e) {
-            // This should not happen!
-            // One should always be able to retrieve the application info for its own package.
-            Log.w(TAG,"This should not happen",e);
-        }
-        return false;
-    }
-
-    /**
-     * Returns true if root is available and enabled in settings.
-     */
-    private boolean useRoot() {
-        return mPreferences.getBoolean(SyncthingService.PREF_USE_ROOT, false) && Shell.SU.available();
     }
 }
--- a/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
+++ b/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
@@ -1,6 +1,8 @@
 package com.nutomic.syncthingandroid.util;
 
 import android.content.Context;
+import android.content.pm.ApplicationInfo;
+import android.content.pm.PackageManager.NameNotFoundException;
 import android.os.Build;
 import android.os.Environment;
 import android.text.TextUtils;
@@ -18,6 +20,7 @@ import org.xml.sax.SAXException;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.DataOutputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.Locale;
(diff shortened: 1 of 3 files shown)


Nutomic:
There's no need to open a new PR if you change something. Just push new changes to the same branch ;)

src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -63,19 +67,75 @@ public ConfigXml(Context context) throws OpenConfigException {
...
+    private boolean checkPermConfigFile() {
+	// Since either readConfig or saveChanges called this method, the permissions of config.xml got messed up. 
+	// For now, the only way to do so is to run Syncthing as root, because the config file resides in the application's data directory.
+	// That directory is only writeable by this application, and root of course. Syncthing will save it's config as root with 0600.
+	// We can safely assume that root magic is somehow available.
+	// Be paranoid :) and check if su binary is available.
+	if (!Shell.SU.available()) {
+	    Log.e(TAG,"Su is not available. Cannot fix permssions ");
+	    return false;
+	}
```
Nutomic: You should only run this if the user has enabled root in the app settings. If root is disabled, you should use the previous behaviour (do nothing and show an error toast).

src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -63,19 +67,75 @@ public ConfigXml(Context context) throws OpenConfigException {
...
+		    fixPermOut.flush();
+		    if (fixPerm.waitFor() == 0) {
+			Log.d(TAG,mConfigFile + ": Can read? " + mConfigFile.canRead() + ", Can write? " + mConfigFile.canWrite());
+			Log.i(TAG,"Successfully changed the owner of the config file.");
+			return true;
+		    } else {
+			return false;
+		    }
+		} catch (Exception e) {
+		    Log.w(TAG,"Cannot chown config file",e);
```
Nutomic: Please dont catch all exceptions, that will hide things like NullPointerException.

src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -63,19 +67,75 @@ public ConfigXml(Context context) throws OpenConfigException {
...
+            DocumentBuilder db = DocumentBuilderFactory.newInstance().newDocumentBuilder();
+	    Log.d(TAG,"Trying to read '" + mConfigFile + "'");
+            mConfig = db.parse(mConfigFile);
+        } catch (SAXException | ParserConfigurationException | IOException e) {
+	  Log.w(TAG,"Cannot read '" + mConfigFile + "'",e);
+	  throw new OpenConfigException();
+	}
+    }
+
+    private boolean checkPermConfigFile() {
```
Nutomic: Please add some javadoc to describe what the method does, as its relatively complicated. You can probably move some of your inline comments into javadoc.

src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -63,19 +67,75 @@ public ConfigXml(Context context) throws OpenConfigException {
...
+	    return false;
+	}
+
+	try {
+		ApplicationInfo appInfo = mContext.getPackageManager().getApplicationInfo(mContext.getPackageName(),0);
+		Log.d(TAG,"Uid of '" + mContext.getPackageName() + "' is " + appInfo.uid);
+		try {
+		    Process fixPerm = Runtime.getRuntime().exec("su");
+		    DataOutputStream fixPermOut = new DataOutputStream(fixPerm.getOutputStream());
+		    String cmd = "chown " + appInfo.uid + ":" + appInfo.uid + " " + mConfigFile + "\n";
```
Nutomic: I didn't know about appInfo.uid, that's very useful. We should probably also use that for SettingsActivity.ChownFilesRunnable (that seems to be broken at the moment, so the app doesnt start after disabling root). Feel free to add a commit for that.

src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -63,19 +67,75 @@ public ConfigXml(Context context) throws OpenConfigException {
...
+		    }
+		} catch (Exception e) {
+		    Log.w(TAG,"Cannot chown config file",e);
+		}
+	} catch (NameNotFoundException e) {
+		// This should not happen!
+		Log.w(TAG,"This should not happen",e);
+	}
+	return false;
+    }
```
Nutomic: And please use spaces for all indents, not tabs. Also, I'd prefer if you move this method to the end of the class, so the more common methods can be first.

src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
flipreverse: Can you pls give me a short explanation why this method is necessary?
I do not get the point here. :(

FYI, the chmod of the Syncthing binary seems to be broken, too. It is blocked by SELinux: 
09-19 23:36:31.720 18241 18241 W chmod   : type=1400 audit(0.0:1122): avc: denied { setattr } for name="libsyncthing.so" dev="dm-0" ino=426157 scontext=u:r:untrusted_app:s0:c512,c768 tcontext=u:object_r:apk_data_file:s0 tclass=file permissive=0

Nutomic: To make sure database files etc are owned by the app (not root), when the root setting is disabled.

Nutomic: Regarding chmod, I'm not quite sure why we need that. @Zillode

flipreverse: Are you sure?
I disabled root manually by editing the shared_prefs.xml. O:-) This is somehow similar to removing that part of the code. I now cannot start Syncthing, because it cannot read the database.

Nutomic: That's exactly why this method exists. When root is disabled in the settings, it chowns all files in the app folder to make sure they are readable.

flipreverse:
Pls do not merge. I'll have to re-fix something. I thought it might be a good idea to check if 'use_root' is set before chowning the config.xml. However, if root gets disabled, this may leave config.xml as it is. That means root still owns that file.

Nutomic:
That shouldnt be a problem, as SettingsActivity.ChownFilesRunnable is supposed to fix the permissions in that case. But as I said, it seems to be broken.

flipreverse:
I might have found the issue. In SettingsActivity.java:158 a onPreferenceChangeListener for the useRoot checkbox is set, but it should be an onPreferenceClickListener. The described behavior is implemented onPreferenceClick.
Moreover, my chown does not know the parameter '--reference'. I think I'll change the operation to make use of ApplicationInfo.

Nutomic:
You can probably extract the chown stuff into a common Util function.

flipreverse:
Yeah. You're right. Is it ok if I add the code for ChownRunnable in this PR?

Nutomic:
Yes thats totally fine.

flipreverse:
Btw, do have an IRC channel? :)

src/main/java/com/nutomic/syncthingandroid/util/Util.java
```diff
@@ -55,4 +65,59 @@ public static String readableTransferRate(Context context, long bits) {
...
+            fixPermOut.writeBytes(cmd);
+            // For now, we assume that a file's type should *always* be 'u:object_r:app_data_file:s0:c512,c768'.
+            // At least for those files residing in an application's data folder.
+            // Running syncthing as root, makes it 'u:object_r:app_data_file:s0'.
+            // Simply reverting the type to its default should do the trick.
+            cmd = "restorecon -R " + dir + "\n";
+            Log.d(TAG,"Running: '" + cmd);
+            fixPermOut.writeBytes(cmd);
+            fixPermOut.writeBytes("sleep 2\n");
+            fixPermOut.writeBytes("exit\n");
```
AudriusButkevicius: The fact that its called output stream is a bit strange, as smells that we are writing to stdout. Also, is the sleep and exit needed?

flipreverse: Yeah. Thats the Java semantic. Here from the Android Java Doc: "Returns the output stream connected to the normal input of the subprocess. "
Regarding sleep: We might discard that.
Regarding exit: We can discard that as well if we can terminate the root shell using Process.terminate() (--> fixPerm.terminate()). If not, we have to stick to that solution.

AudriusButkevicius: Shouldn't the shell just exit once you close the stdin?

flipreverse: You're right. I've changed it. I removed the sleep and exit call, and closed the outputstream.

Nutomic:
IRC is #syncthing on Freenode. There's also a forum btw.

A minor thing about formatting, can you add a space after every comma? Makes it consistent and easy to read.

src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -54,26 +62,38 @@
 
     private Document mConfig;
 
+    private final SharedPreferences mPreferences;
+
     public ConfigXml(Context context) throws OpenConfigException {
         mContext = context;
         mConfigFile = getConfigFile(context);
+        mPreferences = PreferenceManager.getDefaultSharedPreferences(mContext);
```
Nutomic: Seems to be unused now.

src/main/java/com/nutomic/syncthingandroid/util/Util.java
```diff
@@ -55,4 +65,58 @@ public static String readableTransferRate(Context context, long bits) {
...
+            Log.d(TAG,"Running: '" + cmd);
+            fixPermOut.writeBytes(cmd);
+            fixPermOut.flush();
+            fixPermOut.close();
+            if (fixPerm.waitFor() == 0) {
+                Log.i(TAG,"Successfully changed the owner, the group and the SELinux context of '" + dir + "'.");
+                return true;
+            } else {
+                return false;
+            }
```
Nutomic: You might as well log the exit code and save the if statement.

src/main/java/com/nutomic/syncthingandroid/util/Util.java
```diff
@@ -55,4 +65,58 @@ public static String readableTransferRate(Context context, long bits) {
...
+            Process fixPerm = Runtime.getRuntime().exec("su");
+            DataOutputStream fixPermOut = new DataOutputStream(fixPerm.getOutputStream());
+            String dir = context.getFilesDir().getAbsolutePath();
+            String cmd = "chown -R " + appInfo.uid + ":" + appInfo.uid + " " + dir + "\n";
+            Log.d(TAG,"Running: '" + cmd);
+            fixPermOut.writeBytes(cmd);
+            // For now, we assume that a file's type should *always* be 'u:object_r:app_data_file:s0:c512,c768'.
+            // At least for those files residing in an application's data folder.
+            // Running syncthing as root, makes it 'u:object_r:app_data_file:s0'.
+            // Simply reverting the type to its default should do the trick.
```
Nutomic: Is that comment still relevant, seeing as chown uses the uid directly?

flipreverse: Yes, it is still relevant -- more or less. I'm rephrasing it.

Nutomic:
It seems like there's another problem with disabling the root setting: after disabling the setting, Syncthing can't be stopped, because SyncthingRunnable.killSyncthing() is then run without root. I will try to fix that.

Thanks for implementing this :)

@calmh Please add @flipreverse to the Android team ;)

calmh:
@flipreverse invite sent

PR #950, opened 2017-09-20
Added german translation for string 'environment variables'

flipreverse:


Nutomic:
Translations are handled via [Transifex](https://www.transifex.com/syncthing-android/syncthing-android/). Please submit them there.

PR #951, opened 2017-09-20
Changed misleading title of sttrace option

flipreverse:
The sttrace preference field just takes a list of facilities
as an input (https://docs.syncthing.net/dev/debugging.html).
Calling it 'Debug Options' might be misleading since Syncthing
offers way more debugging options than just sttrace.
Naming it 'STTRACE options' makes it clear for the user what
he or she should enter.


Nutomic:
Makes sense. But please remove the German translation, and submit it via Transifex after it's merged. Otherwise, it would just be overwritten.

flipreverse:
@Nutomic Did so. :)

Issue #953, opened 2017-09-20
Environment variables are ignored when running as root


flipreverse:
Hi folks!

If Syncthing runs as root, the su binary does not pass the enviroment variables to the Syncthing binary.
Therefore, default settings, e.g., STNORESTART=1, and custom settings like STTRACE=all are skipped. This is the default behavior for su. I've observed a similar behavior of sudo on Ubuntu.  
To overcome that issue, su provides the option '--preserve-environment'. Adding this to the ProcessBuilder in SyncthingRunnable.java should fix this problem.
However, I cannot test that fix, because the su binary on my Fairphone 2 goes crazy. It deletes custom variables and resets some of them, e.g., HOME. Even with '--preserve-environment' the custom variables are deleted.
I'm currently working on a patch that suits my device. I do not know if it's suitable as a generic solution to that problem. I'll provide my patch as soon as it is completed -- more or less. :)

Cheers,
Alex

Nutomic:
I can't reproduce this on my device (Nexus 5, Android 6). Maybe you just have a wonky su binary?

flipreverse:
Can you set STTRACE=all and you get a very verbose Syncthing without using --preserve-environment?

Nutomic:
Oh actually you're right, I didn't check properly. I tried by changing SyncthingRunnable:126 to `new ProcessBuilder("su", "-c", "printenv")`, which correctly prints some env vars, but not any of those we set. `--preserve-environment` also doesn't help for me.

PR #954, opened 2017-09-20
A fix for issue #953

flipreverse:


src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
```diff
@@ -123,31 +125,49 @@ public void run() {
...
+
+
+            if (useRoot) {
+                process = pb.start();
+                // The su binary prohibits the inheritance of environment variables.
+                // Even with --preserve-environment the environment gets messed up.
+                // We therefore start a root shell, and set all the environment variables manually.
+                DataOutputStream suOut = new DataOutputStream(process.getOutputStream());
+                for (Map.Entry<String, String> entry : targetEnv.entrySet()) {
+                    suOut.writeBytes("export " + entry.getKey() + "=\"" + entry.getValue() + "\"\n");
```
Nutomic: This would probably be more readable if you use String.format().

Nutomic:
There are some conflicts, can you rebase please?

Also, the run() method is getting very long. Could you extract some of it into seperate functions? Like `buildEnvironment()` which returns the targetEnv, and `setEnvironment()` which applies it depending if root is enabled or not.

flipreverse:
Done. :)

src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
```diff
@@ -126,31 +127,15 @@ public void run() {
...
-            if (sp.getBoolean("use_tor", false)) {
-                env.put("all_proxy", "socks5://localhost:9050");
-                env.put("ALL_PROXY_NO_FALLBACK", "1");
-            }
-            if (sp.getBoolean("use_legacy_hashing", false))
-                env.put("STHASHING", "standard");
-            putCustomEnvironmentVariables(env, sp);
+            HashMap<String, String> targetEnv = buildEnvironment(sp);
+            process = setupAndLaunch(pb, targetEnv);
+            targetEnv = null;
```
Nutomic: There's no reason to set this to null.

src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
```diff
@@ -399,4 +384,49 @@ private void trimLogFile() {
...
+            targetEnv.put("STHASHING", "minio");
+            if (sp.getBoolean("use_tor", false)) {
+                targetEnv.put("all_proxy", "socks5://localhost:9050");
+                targetEnv.put("ALL_PROXY_NO_FALLBACK", "1");
+            }
+            if (sp.getBoolean("use_legacy_hashing", false))
+                targetEnv.put("STHASHING", "standard");
+            putCustomEnvironmentVariables(targetEnv, sp);
+            return targetEnv;
+    }
```
Nutomic: Indentation is wrong (8 spaces instead of 4).

src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
```diff
@@ -399,4 +384,49 @@ private void trimLogFile() {
...
+            // We therefore start a root shell, and set all the environment variables manually.
+            DataOutputStream suOut = new DataOutputStream(process.getOutputStream());
+            for (Map.Entry<String, String> entry : env.entrySet()) {
+                suOut.writeBytes(String.format("export %s=\"%s\"\n", entry.getKey(), entry.getValue()));
+            }
+            suOut.flush();
+            // Exec into Synchting, because we want to wait for Syncthing.
+            // Exec will replace the su process by Syncthing.
+            suOut.writeBytes("exec " + TextUtils.join(" ", mCommand) + "\n");
+            suOut = null;
```
Nutomic: What's the reason for using exec here? I don't really understand from the comment.

Also, there's no reason to set suOut to null. But you should close() it.

src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
```diff
@@ -126,31 +127,15 @@ public void run() {
         try {
             if (wakeLock != null)
                 wakeLock.acquire();
+
             ProcessBuilder pb = (mUseRoot)
-                    ? new ProcessBuilder("su", "-c", TextUtils.join(" ", mCommand))
+                    ? new ProcessBuilder("su")
                     : new ProcessBuilder(mCommand);
```
Nutomic: You should probably move this into `setupAndLaunch()`, as it doesn't make much sense on its own.

src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
flipreverse: Which line are you referring to?

src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
flipreverse: I want Java to free the memory for targetEnv since it is not needed anymore.

src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
flipreverse: The exec function replaces the su process with Syncthing as execlp does in C.
If we do not use exec, we'll drop to the root shell as soon as Syncthing terminates, and we'll wait infinitely for the process to terminate (line 146).

I think when we close the stream, which is the stdin of su, su will terminate. This is the way I terminate the shell in fixAppDataPermissions as pointed out by @AudriusButkevicius .
We therefore should *not* close, but we want to get rid of the stream object.

src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
flipreverse: Yeah. Fixed.

src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
Nutomic: The whole function body.

src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
Nutomic: That's something the compiler can figure out on its own.

src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
```diff
@@ -399,4 +379,56 @@ private void trimLogFile() {
...
+        putCustomEnvironmentVariables(targetEnv, sp);
+        return targetEnv;
+    }
+
+    private Process setupAndLaunch(HashMap<String, String> env) throws IOException {
+        Process process = null;
+
+        ProcessBuilder pb = (mUseRoot)
+                ? new ProcessBuilder("su")
+                : new ProcessBuilder(mCommand);
```
Nutomic: Can you move this into the `if (mUseRoot)` statement? I don't think there's a reason to have it seperately.

flipreverse: No, because the ProcessBuilder has to be started in the else branch as well.

Nutomic: Yes, but you can define seperate ProcessBuilders in the if/else branches. Much simpler than having 2 conditionals that check the exact same condition.

flipreverse: aaah. ok. :) Got it.

Nutomic:
Thanks :)

Issue #873, opened 2017-04-01
Build Syncthing with cgo and GOOS=android


calmh:
This will hopefully get you better networking support, and we can qualify the remaining Android specializations on the architecture at build time instead of checking at runtime. I tested the procedure and it works fine - at least the build does, I can't test the binaries.

https://gist.github.com/calmh/20c24afc283656695b76f7822c8c8997

Nutomic:
@AudriusButkevicius @Zillode Could you try to integrate this with the existing build scripts? I'm really bad with the Go build tools unfortunately.

AudriusButkevicius:
I am still trying understand what this will give us.

calmh:
I'm assuming that the goos=android platform exists for a reason.

AudriusButkevicius:
It mostly exists for gomobile reasons being able to bind with Java.

From what I've looked at:

```
Audrius@Audrius src $ for A in `find -name '*.go' -or -name '*.s' | grep -v 'test' | grep -v '/cmd/' | grep -v '/build/'`; do grep -l android $A; done
./net/cgo_linux.go
./net/cgo_resnew.go
./net/cgo_resold.go
./net/cgo_socknew.go
./net/conf.go
./os/executable_procfs.go
./os/file_unix.go
./os/user/getgrouplist_unix.go
./os/user/listgroups_unix.go
./os/user/lookup_android.go
./os/user/lookup_stubs.go
./os/user/lookup_unix.go
./runtime/cgo/cgo.go
./runtime/internal/sys/gengoos.go
./runtime/internal/sys/zgoos_android.go
./runtime/internal/sys/zgoos_linux.go
./runtime/os_linux_arm.go
./runtime/rt0_android_386.s
./runtime/rt0_android_amd64.s
./runtime/rt0_android_arm.s
./runtime/rt0_android_arm64.s
./runtime/sys_linux_386.s
./runtime/sys_linux_amd64.s
./runtime/tls_arm.s
./runtime/write_err.go
./runtime/write_err_android.go
./time/zoneinfo_unix.go
```

logging to /dev/log/main
some asm tweaks for runtime and TLS
different location for tz defs
different location for tmpdir
disables fpu checks
disables user id/user group lookups
handling of /proc/self/exe (same as linux)
cgo resolver

Nutomic:
Finally gave this a try, but running into errors, using this code:
```
cd ext/syncthing/src/github.com/syncthing/syncthing

[...]

PR #959, opened 2017-09-28
Build with Gomobile

Nutomic:
I have started to try and make our build work with gomobile. If we can get this to work, it would give us a bunch of advantages over the existing build process:

- no more hacks needed to make DNS work
- we can get rid of all the custom bash scripts to build Syncthing
- fixes issues #706, #873, #960
- we can eventually add sdcard support using [gobind](https://godoc.org/golang.org/x/mobile/cmd/gobind) (allows calling Android Java APIs from Go)

There are still some problems with this. First of all, the build isn't working and throwing errors ([output is here](https://gist.github.com/Nutomic/4fd51a707886f1b01ac5ca4eff68de95)). But I have checked it with `golang.org/x/mobile/example/bind/hello` instead of Syncthing and it worked, so it seems to be a problem in Syncthing?

Also, gomobile doesn't seem to support building ARMv5 binaries, and creates only ARMv7 instead. I've created a [question on Stackoverflow](https://stackoverflow.com/q/46461107/1837158) for that.

### Setup instructions

Setup Go and Android build tools first. Then run:
```
go get golang.org/x/mobile/cmd/gomobile
go get golang.org/x/mobile/cmd/gobind
gomobile init -ndk /path/to/ndk
go get github.com/syncthing/syncthing/cmd/syncthing
```

Then just compile with `./gradlew assembleDebug`.

ping @calmh @Zillode @AudriusButkevicius

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,9 @@
 buildscript {
     repositories {
         jcenter()
+        maven {
+            url "https://plugins.gradle.org/m2/"
+        }
 
     }
     dependencies {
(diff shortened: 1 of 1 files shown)


Nutomic:
Gomobile officially requires ARMv7: https://github.com/golang/go/issues/22108#issuecomment-334984137. That means we will have to drop support for those devices after switching to Gomobile.

PR #968, opened 2017-10-13
Build Syncthing with CGO and target Android

Nutomic:
As discussed in #873. Requires https://github.com/syncthing/syncthing/pull/4417. 

This currently works by generating a [standalone ndk toolchain](https://developer.android.com/ndk/guides/standalone_toolchain.html). That isn't great, as it will copy about 3 GB of data on every clean build. It should be possible to avoid that and use the NDK directly, but I didn't get that to work yet.


Nutomic:
@calmh @AudriusButkevicius Do you know why the Travis CI Syncthing build would constantly fail with this error?
```
can't load package: package github.com/syncthing/syncthing/cmd/syncthing: cannot find package "github.com/syncthing/syncthing/cmd/syncthing" in any of:
	/home/travis/.gimme/versions/go1.7.4.linux.amd64/src/github.com/syncthing/syncthing/cmd/syncthing (from $GOROOT)
	/home/travis/gopath/src/github.com/syncthing/syncthing/cmd/syncthing (from $GOPATH)
```

I fixed that by running `go get github.com/syncthing/syncthing/cmd/syncthing` in the travis script, but now I'm always getting this error:
```
../../../../gopath/src/github.com/syncthing/syncthing/cmd/syncthing/gui_statics.go:37: undefined: auto.Assets
../../../../gopath/src/github.com/syncthing/syncthing/cmd/syncthing/gui_statics.go:44: undefined: auto.Assets
```

Travis is running Go 1.9 on Ubuntu 14.04. I can't reproduce either problem locally.

imsodin:
The second is due to missing assets:
https://forum.syncthing.net/t/rework-build-process-to-avoid-repo-bloat/7054/4
You can't go get syncthing, you need to use build.go.

imsodin:
I think your current problem is that you are not building in gopath. 

build.go generates the assets in the working directory:

```cd "${ROOT_DIR}/ext/syncthing/src/github.com/syncthing/syncthing"```
https://github.com/syncthing/syncthing-android/pull/968/files#diff-ac695f831da9cd9755327726ea1b75e6R10

And building happens in $GOPATH which apparently is not the same place:

```/home/travis/gopath/src/github.com/syncthing/syncthing/cmd/syncthing/gui_statics.go:37: undefined: auto.Assets```
https://travis-ci.org/syncthing/syncthing-android/builds/288494822#L2102

This is probably not what you intend to do, however build.go should still pick the problem up and do magic or complain I think, so something in the generation of assets seems broken in relation to gopath.

Nutomic:
Syncthing is already loaded as a submodule, so there is no reason for the first error. And the second error shouldn't happen either, as I am compiling via the build.go script. And like I said, neither of these errors happens on my laptop.

Nutomic:
Sorry didn't see your second comment. Gopath is unset on both travis ci and my laptop, so I don't know what I should do here.

AudriusButkevicius:
I don't see you setting GOPATH, it should be set to ext/syncthing.

imsodin:
To reproduce go to an empty dir and set `export GOPATH=$PWD` then run `go get github.com/syncthing/syncthing/cmd/syncthing`

And GOPATH already needs to be set before the go get line
https://github.com/syncthing/syncthing-android/pull/968/files#diff-354f30a63fb0907d4ad57269548329e3R29
Or get rid of it entirely and do git clone *syncthing* && git checkout *tag*

AudriusButkevicius:
Because Its probably already set or the assets already pregenerated from previous run.

Issue #970, opened 2017-10-17
Changing rotation and changing app cause a Refresh of the Website


Mannshoch:
Please deactivating the refreshing of the Website after every single bit you do on it.
Sometimes I have to prepare something and the Website is refreshing. Then I could again wait for the Pop-up info-box for a new device or a folder what some times needs some minutes.

PR #985, opened 2017-11-10
Fix web UI being reloaded on orientation changes

sferra:
Declare WebGuiActivity to handle orientation changes (not currently necessary) to prevent the redundant re-creation of the Activity.

See #970

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -46,7 +46,8 @@
         <activity
             android:name=".activities.WebGuiActivity"
             android:label="@string/web_gui_title"
-            android:parentActivityName=".activities.MainActivity">
+            android:parentActivityName=".activities.MainActivity"
+            android:configChanges="keyboardHidden|orientation|screenSize">
             <meta-data
                 android:name="android.support.PARENT_ACTIVITY"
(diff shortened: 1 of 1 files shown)


Nutomic:
Yep, this is probably the best option we have. However, it's not working perfectly, as the web GUI is still reloading when you switch between apps. Also, the `keyboardHidden` options seems unnecessary, as toggling the software keyboard does not cause a reload. On the other hand, screen changes (`screenLayout`, `screenSize`, `uiMode`) might be relevant for some users.

https://developer.android.com/guide/topics/manifest/activity-element.html#config

sferra:
@Nutomic, this pull request only attempts to prevent the reload of the web view on device rotation. it does not attempt to fix the reload of the web view on app switch: that is a completely different problem which should be addressed separately.

`keyboardHidden` is not triggered by the software keyboard, but by devices which offer a hardware keyboard. those devices might become increasingly rare, but i thought to handle the case in for now. i can remove it if you wish.

i also wouldn't handle additional configuration changes unless i am certain that they occur when rotating the device and could trigger the issue.

Nutomic:
Fair enough. Thanks for the contribution!

PR #988, opened 2017-11-14
Fix redundant reload of the web UI

sferra:
Don't reload the web UI when resuming the web UI activity if the URL was already loaded.

See #970

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/WebGuiActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/WebGuiActivity.java
@@ -141,7 +141,21 @@ public class WebGuiActivity extends StateDialogActivity
 
     @Override
     public void onWebGuiAvailable() {
-        mWebView.loadUrl(getService().getWebGuiUrl().toString());
+        final String serviceUrl = getService().getWebGuiUrl().toString();
+        if (!isCurrentUrl(serviceUrl)) {
+            mWebView.loadUrl(serviceUrl);
+        }
+    }
+
+    private boolean isCurrentUrl(final String url) {
+        String currentUrl = mWebView.getUrl();
+        if (currentUrl == null || currentUrl.length() == 0) {
+            return false;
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/nutomic/syncthingandroid/activities/WebGuiActivity.java
```diff
@@ -141,7 +141,21 @@ public void onServiceConnected(ComponentName componentName, IBinder iBinder) {
...
+    private boolean isCurrentUrl(final String url) {
+        String currentUrl = mWebView.getUrl();
+        if (currentUrl == null || currentUrl.length() == 0) {
+            return false;
+        }
+        if (currentUrl.endsWith("/")) {
+            currentUrl = currentUrl.substring(0, currentUrl.length() - 1);
+        }
+        return currentUrl.equals(url);
     }
```
Nutomic: Is there any reason for this check? I think it should be enough to check `mWebView.getUrl() == null` in onWebGuiAvailable().

Nutomic: Also thanks for finding this, I wasn't aware that it was causing the reloads.

sferra: well, I was not sure that the URL would always be null (or empty) as I am not familiar with syncthing's web API, so a check against the actual URL made sense.

Nutomic: Ah I see. We actually only need to set the URL the a single time (to open the Syncthing web gui). So it would be enough to check if the current URL is null, and load the web gui in that case.

sferra: In that case the changes in this pull request are a bit of an overkill. Should I create a separate, simplified pull request with a minimal check?

PR #990, opened 2017-11-15
Fix redundant web UI reload

sferra:
Don't reload the web UI if it was already loaded.
Follow up on #988, see #970.


Nutomic:
Thanks! Btw there's no need to make a new PR, you can just (force) push to the existing branch.

Issue #1009, opened 2018-01-26
Recurring Usage Reorting dialog (v.0.10.03)


scienmind:
Steps:

1. Open Syncthing
2. Usage Reporting Dialog appear
3. Accept/Reject it
4. Restart as prompted
5. After a few seconds the prompt appears again

Restarting syncthing doesn't change anything - the UR prompt keep popping up
Setting the UR status via webUI, has no effect - the prompt keep appearing.

Curiously enough, the issue has appeared for me only after updating to v.0.10.03 (which mentions https://github.com/syncthing/syncthing-android/commit/08bf74d7aa523a91ac393299ff19463ea9fc85d7 in the changelog)

### Version Information

App Version: 0.10.03
Syncthing Version: v0.14.43
Android Version: Android 8.0; 8.1

cornclaw:
I can confirm that this is happening as well on Android 7.0

Craeckie:
Same issue on Android 6.0.1 with same version, just that the dialog appears many times on top of each other, so the app is almost unusable. Switching off the internet connection doesn't help.

aamadorj:
Same here: PH-1 running AOSiP 8.1.

vsespb:
Same. Unusable. Xiaomi Redmi 4

imsodin:
As you can see right after the original report, the issue has already a fix proposed in PR #1010. Please refrain from more "same here" comments. They don't help, they just create noise.

Flouss:
As a workaround until this gets fixed, just open the Web GUI. The usage report dialog doesn't appear there.

Catfriend1:
Workaround: open data/data/com.nutomic.sync..../files/config.xml with root explorer and set urseen to 3 and uraccepted to 2. Then the dialog doesnt appear anymore.

narbehaj:
Same issue here, after updating it to latest release Jan 26. Reporting dialog shows up repeatedly. LG-V20 Android 7.0

licaon-kter:
Same here, Play version, Android 5.1.1.

The WebGUI helps, albeit it always has a dialogue on the top of the page with "settings need saving..."

DrSchnagels:
Confirmed, 1+ 5 Oreo 8.0, super annoying lol.
Please fix asap, im in the hospital and i need to sync my pictures.
thanks!

alldigital:
@Catfriend1 your hack did the trick. Thanks.

AudriusButkevicius:
[...]

PR #1010, opened 2018-01-26
MainActivity.java: Don't show Usage Reporting dialog if aready decided

scienmind:
Fix the condition to show usage reporting prompt. (Fixes: https://github.com/syncthing/syncthing-android/issues/1009)

I don't have access to Android Studio atm, so I haven't actually tested it - keep this in mind.


Nutomic:
Damn it, I definitely messed that up. Thanks for the fix!

PR #1024, opened 2018-02-21
Fix issue #1005 Endless loader after enabling "Superuser mode"

Catfriend1:
I already compiled a debug apk and root mode worked fine with the fix. No more endless loading screen after activating experimental "root mode" setting.


Nutomic:
Great! Can you set a proper commit message, eg like commit 4705332a9a658ebd20d0f1e9080c18b76934f591? You can use the commands `git commit --amend` and `git push -f` on the command line.

Catfriend1:
retrying to submit it properly...

PR #1025, opened 2018-02-22
Call suOut.flush on starting the syncthing binary using root - Fixes issue https://github.com/syncthing/syncthing-android/issues/1005

Catfriend1:
Already verified working with a debug apk build on LG H815 rooted phone.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ before_script:
 
 # Install Golang
 before_install:
+  - yes | sdkmanager "platforms;android-27"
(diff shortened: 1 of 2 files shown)


Catfriend1:
Hmm commit --amend worked but how can I delete those old commits from yesterday?

Nutomic:
You would have to use `git rebase -i`, best google around for that. But I can fix that via Github so dont worry. Btw commit messages should be less than 80 characters per line.

And please remove the change to travis.yml, I'm trying to fix that myself.

PR #1027, opened 2018-02-23
Fix root startup of libsyncthing.so binary

Catfriend1:
Call suOut.flush on starting the syncthing binary using root
Fixes issue https://github.com/syncthing/syncthing-android/issues/1005


Nutomic:
Thanks :)

However, there is really no need to make a new PR. Check this post: https://stackoverflow.com/a/15055649/1837158

And [here is the correct syntax to close an issue from a commit](https://stackoverflow.com/a/6742691/1837158).

Catfriend1:
Thanks for the explanation. Helpful for a newbie like me. :)

Nutomic:
Youre welcome. And feel free to ask if anything is unclear ;)

PR #1040, opened 2018-03-14
Prevent to post rooted sub to scanner rest api

### LINKED ISSUE #1032: Syncthing walking the same folder twice
Since upgrading the app to 0.10.5 I encountered a serious showstopper that renders the app unusable. It occurs both using experimental root and not using the experimental root feature.

Reproduction:
- Clear app data and caches (if the app was not fresh installed)
- Setup folder "/storage/emulated/0" as syncthing folder read-write with default settings and scan interval 10 minutes (you can use another scan interval as well - I used this one)
- Setup the ignore list before the first scan to ignore special folder
`(?i)/.android_secure
(?i)/Android/data/**/cache
(?i)/Android/data/com.nutomic.syncthingandroid/files/*
(?i)/Android/data/com.nutomic.syncthingandroid.debug/files/*
(?i)/com.infosysta.mobile.jirapro
(?i)/DCIM/.thumbnails
(?i)/Domoticz
(?i)/LOST.DIR
(?i)/Sygic/Maps
(?i)/Tasker/log/*
(?i)/TWRP/Backups`
- Syncthing walks the folder - displays "Scanning".... 
- Scanning reaches 100% after some seconds (Note - I have only ~ 1.500 files with the above mentioned ignore patterns) and the scan finishes.
- Folder displays state as "current" (translated from german)
- I took a note of how many files are displayed for that folder in the syncthing web GUI = 1.500 files.
- After around 10 seconds, the "scanning" appears again (I did not add, change, delete files).
- The folder is rescanned "by syncthing itself" without any user interaction
- The scan finishes again very quickly after reaching "100%"
- Folder displays state as "current" (translated from german)
- I took a note of how many files are displayed for that folder in the syncthing web GUI = 3.000 files.
- The file count is twice as high as after the first scan.
- Now the folder stays "idle" until the rescan interval passed by.
- I took a full log with STTRACE=scanner (enabled before the test, attached)

Looking at the log "as being no expert", I suspect it is scanning the folder named "." within "/storage/emulated/0" which causes the double amount of files in the web gui report of the folder. At first glance, the first "real to hash" is called for e.g. "somefolder/somefile.txt" and later, during the "second scan" for "/somefolder/somefile.txt".

### Version Information

    App Version: 0.10.5
    Syncthing Version: v0.14.45
    Android Version: Android 7.1.2 @ lg-h815 (LineAgeOS 2017-11-07)

Catfriend1:
Remove the leading slash for syncthing versions <= 0.14.45 fixes #1032

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/http/PostScanRequest.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/http/PostScanRequest.java
@@ -9,12 +9,16 @@ import com.google.common.collect.ImmutableMap;
 import java.net.URL;
 
 public class PostScanRequest extends ApiRequest {
-
     private static final String URI_SCAN = "/rest/db/scan";
 
     public PostScanRequest(Context context, URL url, String apiKey,
                            String folder, String sub) {
         super(context, url, URI_SCAN, apiKey);
+        
+        /* Prevent to post rooted sub to scanner rest api */
+        /* Remove the leading slash for syncthing versions <= 0.14.45 */
+        /* fixes #1032 */
+        sub = sub.replaceAll("^/+", "");
         Uri uri = buildUri(ImmutableMap.of("folder", folder, "sub", sub));
         connect(Request.Method.POST, uri, null, null, null);
     }


Nutomic:
This seems to be the wrong place to fix it. You should adjust FolderObserver and pass the correct path into OnFolderFileChangeListener. This also seems to be wrong because the path should be relative to the folder root (I think), so it's not enough to strip the first slash.

Btw I have no idea why the build is failing.

Catfriend1:
Thanks for the feedback. I am currently having build problems with 14.45+cherry because imsodins commit is "outside" the main syncthing repository.

PR #1045, opened 2018-03-28
Request ACCESS_COARSE_LOCATION on Android 8.1 (fixes #999)

### LINKED ISSUE #999: After upgrade to Android 8.1, Syncthing stays disabled on named wifi networks
See title.  Was working fine.  Upgraded Android, now not working.  "Syncthing is disabled" message.  If allow it to connect to any wifi network, it starts working again.  Only when I try to "Restrict to certain wifi networks" does it disable itself again.  If I flip that back and forth, then Syncthing will continue to renable and disable itself.

I see some other "Syncthing is disabled" bug reports here, but I'm not sure whether this is the same issue.

    App Version: 0.10.1
    Syncthing Version: not listed in app?
    Android Version: Android 8.1.0, patch level Dec. 5

capi:
As outlined in https://issuetracker.google.com/issues/70795529, accessing the WiFi's SSID on Android 8.1 requires `ACCESS_COARSE_LOCATION`.

This PR adds this permission to the manifest and on Android 8.1+ it requests the permission, if the WiFi SSID preference is selected and the permission has not yet been granted.

In case the permission is rejected, a short explanation is shown.

In case the permission granted, the WiFi's SSID is refreshed via `SyncthingService` and configuration of the restricted SSIDs can resume.

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
@@ -1,5 +1,6 @@
 package com.nutomic.syncthingandroid.activities;
 
+import android.Manifest;
 import android.app.AlertDialog;
 import android.content.Intent;
 import android.content.SharedPreferences;
@@ -13,6 +14,7 @@ import android.preference.ListPreference;
 import android.preference.Preference;
 import android.preference.PreferenceFragment;
 import android.preference.PreferenceScreen;
+import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.util.Log;
 import android.widget.Toast;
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -49,6 +51,26 @@ protected void onCreate(Bundle savedInstanceState) {
                 .commit();
     }
 
+    @Override
+    public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults) {
+        // On Android 8.1, ACCESS_COARSE_LOCATION is required, see issue #999
+        if (requestCode ==Constants.PERM_REQ_ACCESS_COARSE_LOCATION) {
```
AudriusButkevicius: Formatting

capi:
Actually, the permission is asked _only_ if the user clicks the "Restrict to certain wifi networks" setting, not before.
Formatting will be fixed shortly.

capi:
Formatting fixed.

AudriusButkevicius:
This won't fix the issue if the feature is already enabled but the permission is not there I assume?

capi:
My assumption for this is, that people have turned off the feature for now, since it won't be syncing, and once they now want to re-enable it, they will be asked for permissions.

Issue #1021, opened 2018-02-08
Annoying notification message even when Wi-Fi is off.


vi:
Syncthing is configured to be only on Wi-FI. Hovewer there is persistent notification "Syncthing is disabled" (or even occasionally "Syncthing is enabled") stays when Wi-Fi is off, especially when mobile network goes on and off.

Syncthing attracts too much of my attention when it should just sit still and wait.

Settings: Run conditions: Always run in background = on (is the accent on "always" or on "background" here?); Run only on wifi = on, run only when charging = off, respect android battery saving = on; Behaviour: Notification = Low Priority.

### Version Information

    App Version: 0.10.4
    Syncthing Version: (empty in Settings->About when Wi-FI is off)
    Android Version: Android 4.4.2

Catfriend1:
This will be discussed on in https://github.com/syncthing/syncthing-android/issues/1048 as there is already a technical approach for a solution found.

Issue #998, opened 2017-12-08
Persistent notification of "Syncthing is Disabled"


Larpon:
For a few versions I've had a *persistent* notification hanging on all my Android devices running syncthing telling me that syncthing is *disabled*. I can't seem to find any settings to hide the notification so I'm reporting this as a bug. To avoid any confusion - I'm not talking about the "Syncthing is *Running*" notification.

I don't need to know if syncthing is off - I need to (optionally) know when it's running :)
Think if every app on your phone notified you it was **not** running :)

Joke aside - I very well could have missed some option - if I haven't - it would be nice to disable the notification or at least make it impersistent?

### Version Information

    App Version: 0.10.0 - from Google Play
    Syncthing Version: blank / nothing happens when clicking in the "About" section?? - but latest version (0.14.40?)
    Android Version: Android 4.1.2 + Android 6.0.1

Ajedi32:
Yeah, ideally a notification should only appear when Syncthing is actively scanning or transferring data, like how every other sync app on Android behaves.

I guess this is probably more difficult than it sounds though, since this app is only a wrapper over Syncthing, and thus doesn't have as much control over and visibility into Syncthing's behavior as perhaps it should.

Ajedi32:
Probably related to #994

Larpon:
I think issue 994 is unrelated though - it's about Oreo and battery usage. My oldest device with this issue is 4.1.2 - I haven't looked at the source so I'm not aware how the Android java code communicate with syncthing - but I'm guessing it's by the http API?

Ajedi32:
The reason I think 994 is related is because I suspect that's the reason why these notifications were added in the first place. They're necessary for apps targeting Oreo (regardless of what other versions of Android they're backwards-compatible with).

PanderMusubi:
Issue is still there on Android 8.0.0

Catfriend1:
@PanderMusubi Is the issue still there in recent version 0.10.8? According to PR https://github.com/syncthing/syncthing-android/pull/1046 the notification should be able to be prioritized and sent to bottom.

akwala:
Works as expected for me.

PanderMusubi:
@Catfriend1 unfortunately, yes. I see the icon at the top of the screen when syncthing is running and is disabled. And in both cases there is a notification whether it is disabled or running when I pull down the notification drawer and cannot get rid of that notification in both cases. When I slide the notification left or right, it simply bounces back.

[...]

PR #1046, opened 2018-03-28
Use separate notification channel while monitoring run conditions

capi:
As discussed in #998, or #1021, since Android 8.0 we are currently forced to have the service running with foreground priority in order to receive broadcasts. This leads to the "Syncthing is disabled" notification, which is annoying.

Splitting the "Syncthing is active" and "Syncthing is disabled" into two separate notification channels at least allows on 8.0+ to hide or at least make the "Syncthing is disabled" notification less intrusive.

Unfortunately this advantage remains only for 8.0+, so this is not a fix for the issues referenced above.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -541,7 +541,8 @@ Please report any problems you encounter via Github.</string>
 
     <string name="notification_crash_text">Click to view logs</string>
 
-    <string name="notifications_persistent_channel">Persistent notification</string>
+    <string name="notifications_persistent_channel">Syncthing active</string>
+    <string name="notification_persistent_waiting_channel">Monitoring run conditions</string>
     <string name="notifications_other_channel">Other notifications</string>
 
     <!-- RestApi -->
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/NotificationHandler.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/NotificationHandler.java
@@ -22,16 +22,19 @@ import javax.inject.Inject;
 public class NotificationHandler {
 
     private static final int ID_PERSISTENT = 1;
+    private static final int ID_PERSISTENT_WAITING = 4;
     private static final int ID_RESTART = 2;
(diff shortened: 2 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -541,7 +541,8 @@ Please report any problems you encounter via Github.</string>
 
     <string name="notification_crash_text">Click to view logs</string>
 
-    <string name="notifications_persistent_channel">Persistent notification</string>
+    <string name="notifications_persistent_channel">Syncthing active</string>
```
imsodin: I think this should use a new `name`, otherwise the existing translations for "Persistent notification" will keep being active.

capi: No, I just wanted to rename the channel. I tested it out, it works. No new channel is created, but the display name of the channel has been updated from "Persistent notification" to "Syncthing active", which now fits better.

capi: Oh, I understand what you mean, I misread. You are right about the translations, of course.

imsodin: Sure in English there's not going to be a problem, I am thinking about the translation system: Is it smart enough to detect the change and remove all existing translations for the `notifications_persistent_channel` string? Otherwise the existing translations of "Persistent notification" will keep being used, while new translations for "Syncthing active" should be used.

capi: Unfortunately I've never cared about the translations yet. I'm not sure how they are done. Are they imported manually? I'll have a look and if it proves to be a problem, I'll either remove the current translations there and/or rename the key, as appropriate.

imsodin: I believe transifex automatically imports/reads this file and the translations then need to be imported/committed manually (e.g. https://github.com/syncthing/syncthing-android/commit/8f324a866dcf93e76abee3439d4d4f30ef007c89). I requested access to German&French translations to check what happened to the string after this merge - maybe anyone else with access could check it too, because I won't be active anymore for a bit over a week past midday today :)

PR #1049, opened 2018-04-07
Docker image, and windows builds

AudriusButkevicius:
With the image available, the build becomes:

```
docker run -v `pwd`/output:/mnt -e BUILD_DIR=/mnt -e GIT_BRANCH=windows -it --rm android-build ./gradlew buildNative assembleDebug
```

The docker image prebuilds NDKs, Go stdlib with CGO enabled and prefetches most gradle packages, so that the image ends up building only syncthing and the APK (unlike previously, the NDK, then Go stdlib, for each arch etc).

This can be disabled by removing SYNCTHING_ANDROID_PREBUILT env var.

This also fixes builds on Windows via cygwin by sprinkling cygwin shit all over the place.


calmh:
Looks nice, let me take it for a spin

AudriusButkevicius:
Proven that it works, hence merging.

Issue #1013, opened 2018-01-31
FS Watcher does not seem to work, logs broken too


vsespb:
I have DCIM folder (internal storage) shared and FS watcher enabled. I have only two devices paired - Android and Linux.

I do the folloging:

1) enabled debug log for and "fs" and "watchaggregator"
2) make some photos
3) go and watch advanced logs

I expect: photos immediately uploaded

But what happens is:

Photos uploaded only after 1+ mins.
Logs are empty (even after photo uploaded). Last line in logs is: "Enabled debug data for watchaggregator".

I also played with fs watcher on linux box, and now I definitely see, that under linux file watcher works, under android - does not.

Attaching log screenshot - this is the log that I saw _already_ when I've got photos on another device. Photos files are IMG_20180131_153249 IMG_20180131_153251 IMG_20180131_153252

So
1) photos were made after I turned on logging (see timestamps)
2) photos were uploaded
3) by that time advanced log was still empty

And I feel like file watcher does not work, but no strong evidences, as logs do not work too.

### Version Information

Xiaomi Redmi 4, Android 6.0.1, MIUI 8.0
SyncThing 0.14.43

![screenshot_2018-01-31-15-41-17-373_com nutomic syncthingandroid](https://user-images.githubusercontent.com/1210230/35623989-990ff1cc-069e-11e8-86e7-303b4aa1d2f0.png)

imsodin:
As far as I know the android wrapper has it's own change detection, so it should work even without Syncthings own fs watching enabled. I seem to remember @Nutomic saying that in Android 6 there is some kind of limitation with the Android watching, so maybe that is the issue here.

As for why Syncthing's watcher isn't working: Do you have one of these log lines:  
```INFO: Started filesystem watcher for folder ...```  
```WARNING: Failed to start filesystem watcher for folder ...```

AudriusButkevicius:
I think at some point Nutomic might have disabled the Java notification watcher.

vsespb:
> Do you have one of these log lines:

Yes, I restarted SyncThing and have line that it's started:

![screenshot_2018-01-31-17-05-07-024_com nutomic syncthingandroid](https://user-images.githubusercontent.com/1210230/35627534-36879634-06aa-11e8-969e-e64ec1b694a7.png)

[...]

PR #1051, opened 2018-04-13
add ConfigXML support for fsWatcher transition

### LINKED ISSUE #368: High Power Consumption
Hello!
Today I noted an excessive power consumption of the app (on the second day of use). Synchronization is turned on only in WLAN. There were seven photos to transfer (14 MB via WLAN) and this required 8% of the total battery power. The sync works beautiful, but the power consumption is to much!
Syncthing Version: v0.10.30
Syncthing-Android Version: 0.5.28
Sony Xperia Z1 Compact
Android 4.4.4

ghost:
Purpose
In regard to syncthing does its job very well by scanning folders periodically plus getting the new file watcher feature in the near future, I'd like to give advanced users the choice to disable the wrapper's FolderObserver. FolderObserver is responsible for real-time file watching on the android side and triggers syncthing scans if files are changed. FolderObserver is not fully implemented as logcat shows it is not handling all filesystem notifications received by the android OS.

Implementation
Default of this new experimental option is "FolderObserver" disabled and the file watcher enabled beginning with config v28. It outputs info log lines to logcat on SyncthingService startup telling if the user has enabled or disabled FolderObserver manually.

Additional tasks
Fix https://github.com/syncthing/syncthing-android/issues/368 by adding syncthing core's "restartOnWakeup" setting to the android wrapper's UI. This feature was enabled by default in the past and stays this way. It is probably planned at a future release to change the default to disabled after additional verification and feedback has been reviewed (not in this PR).

Test
a) fsWatcher transistion
Verified working in a debug build on LineAgeOS 7.1.2 arm64 on LG H815 device.
b) restartOnWakeup
Verified working in a debug build on LineAgeOS 7.1.2 arm64 on LG H815 device.


imsodin:
I didn't look at the code, just some general inputs:  
The watcher isn't that far out in the future, so I wouldn't add any "preliminary" features anymore. Actually there should be a "strategy" in place by the release of 0.14.47, otherwise users will get the notification asking to activate the FS watcher in the web UI  - having both that and the folder observer active is not ideal. If you can confirm, that FS watcher works on android and you think it is better than FolderObserver, I'd just enable it by default and dismiss the user notification. On Android 6 I think the FolderObserver doesn't work at all, so there it is anyway a good idea to use the internal watcher.

Nutomic:
I agree with @imsodin, the watcher in Syncthing probably works much better than the one in the Android wrapper (better tested, and it supports things like batching). So you can just get rid of the watcher in the Android app, and enable the watcher in Syncthing. Might be a good idea to do a beta release though.

Catfriend1:
Okay, so when 14.47 is released, I'll make deep tests with a debug build of the android app integrating this version and fs watcher enabled.

AudriusButkevicius (approved):

AudriusButkevicius:
I approved the change as it looks ok, but I agree with others that we should just kill the android watcher.

Catfriend1:
Would it be of some kind of use to set the android watcher to disabled by default? (at the time of writing the pull it is enabled by the default)
So if an advanced app user has trouble with the new fs watcher feature and got used to the default app behaviour syncing immediately due to the android watcher in place, the user can go to the advanced setting and re-enable it until fs watcher got fixed. Then, if there are no more bug reports regarding fs watcher on android after a definite transition time, we can drop the android watcher completely. Making it an advanced by-default-disabled option would give us a chance to gather fs watcher bug reports and help immediately without going through the rebuild an rerelease process.

AudriusButkevicius:
Sure, but there are other things you'd have to do, like enable the in-built watcher, dismiss the notification that is displayed in the web ui.

Catfriend1:
@AudriusButkevicius: Okay, agreed. I'm trying this when I have more time to work on. First, I've to read through the WebGUI REST API and how Nutomic did those things in the past to follow the existing strategy of altering syncthing-cores config.

Nutomic:
@Catfriend1 check out the file [ConfigXml.java](https://github.com/syncthing/syncthing-android/blob/master/app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java#L116). It should work the same way as the ignorePerms stuff.

Catfriend1:
@imsodin Can you please point me to a commit I can use for testing the fs watcher with web notification as the user will later see it?

imsodin:
https://github.com/syncthing/syncthing/pull/4789

In particular the relevant config setting is set here: https://github.com/syncthing/syncthing/pull/4789/files#diff-319b5feabc77b066118cda94584168a8R382  
To dismiss (never show) the notification just remove that config entry.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -156,17 +164,32 @@ public void updateIfNeeded() {
...
         // Disable weak hash benchmark for faster startup.
         // https://github.com/syncthing/syncthing/issues/4348
         Element options = (Element) mConfig.getDocumentElement()
                 .getElementsByTagName("options").item(0);
         changed = setConfigElement(options, "weakHashSelectionMethod", "never") || changed;
 
+        /* Check if we have to dismiss any specific "unackedNotificationID" */
+        /* Dismiss "fsWatcherNotification" */
+        if (getConfigElement(options, "unackedNotificationID").contains("fsWatcherNotification")) {
+          changed = setConfigElement(options, "unackedNotificationID", getConfigElement(options, "unackedNotificationID").replace("fsWatcherNotification", "")) || changed;
```
AudriusButkevicius: You should just remove the element.

Catfriend1:
Sidenote - I'm getting error in the web gui now using the file watcher
2018-04-16 22:25:30: Failed to start filesystem watcher for folder "rw_lgg4_F36E-1114" (srh6g-wqh74): lstat /storage/F36E-1114/.android_secure: permission denied

Catfriend1:
Testing the newest app-debug build with commit 3d02fcd I could not find any more errors. App is running fine on Android 7.1.2@lg-h815 device. LGTM?

Catfriend1:
This is also intended to fix https://github.com/syncthing/syncthing-android/issues/1013

Catfriend1:
@AudriusButkevicius: I saw you recently added "rollout to beta channel". What is your opinion on best strategy for field-testing this? Do we have enough android users within the maintainer target user group so we could invite "ourselves" to the play store beta channel and place a build there?

Catfriend1:
@imsodin @AudriusButkevicius Please let me know if this can be merged or when and what needs to be done before. This part of settings migration code done, I would like to investigate more issues and come up with future PR if I find a fix.

imsodin:
I don't know the code/java, so someone else needs to check it. What I understand looks reasonable. The restart on wakeup stuff isn't related to the watcher though, right? Then I'd move that into a separate commit/PR.

Also note you must not release an RC as a regular release (unless there is a beta type release for the app too), otherwise usage reporting gets enabled unconditionally. So I'd propose not updating the submodule until 0.14.47 in the first place.

Catfriend1:
@imsodin: Can you please give me a hint how to separate commits afterwards or make a new pr not conflicting with this one? I am unsure if merging works if the same code parts were worked on. Thats why I had the idea to consolidate this minor config stuff and then start working on the webgui not loading problem in a new pr. Rc not as Release is clear to me - I'd suggest it only for the beta release Audrius prepared the master for already.:)

imsodin:
Beta release is fine, I don't know the usual release process on Android. It just once happened that an RC got released, so I am sensitised to that.

Don't worry about the PR separation, if it is too annoying to do. I usually just "copy" (`checkout -p` or `cherry-pick --no-commit -p`) the desired parts of the diff to a new branch. The once that branch is in master, you can merge master into the original branch. Unless I do major changes in the same parts of the code I just don't care about conflicts and work away in separate branches - if any arise, they are usually simple to resolve.

AudriusButkevicius:
I need to look at this on a bigger screen than my 5" phone. Hopefully this evening.

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -303,6 +308,7 @@ public boolean onSyncthingPreferenceChange(Preference preference, Object o) {
                     mOptions.globalAnnounceServers = Iterables.toArray(splitter.split((String) o), String.class);
                     break;
                 case "address":               mGui.address = (String) o;  break;
+                case "restartOnWakeup":       mOptions.restartOnWakeup = (boolean) o;              break;
```
AudriusButkevicius: Some sketchy formatting, either move the two options that span the same line above next to others (and then align the breaks), or align don't align the breaks at all, or align the breaks within the subsection.

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -186,9 +188,10 @@ public void onActivityCreated(Bundle savedInstanceState) {
             Preference environmentVariables = findPreference("environment_variables");
             Preference stReset              = findPreference("streset");
 
-            mUseRoot                     = (CheckBoxPreference) findPreference(Constants.PREF_USE_ROOT);
-            Preference useWakelock       = findPreference(Constants.PREF_USE_WAKE_LOCK);
-            Preference useTor            = findPreference("use_tor");
+            mUseRoot                        = (CheckBoxPreference) findPreference(Constants.PREF_USE_ROOT);
+            Preference useWakelock          = findPreference(Constants.PREF_USE_WAKE_LOCK);
+            Preference useTor               = findPreference("use_tor");
```
AudriusButkevicius: I know this is not your change, but should we have a constant for this?

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
```diff
@@ -70,6 +70,11 @@ public SyncthingRunnable(Context context, Command command) {
         mContext = context;
         mSyncthingBinary = Constants.getSyncthingBinary(mContext);
         mLogFile = Constants.getLogFile(mContext);
+
+        // Get preferences relevant for meaningful logs in bug reports.
+        Log.v(TAG, "SyncthingRunnable: syncthing-core option restartOnWakeup = " + Boolean.toString(mPreferences.getBoolean("restartOnWakeup", false)));
```
AudriusButkevicius: This seems like a fairly random place to wedge this in. Isn't there a better place?

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -62,8 +66,32 @@ public ConfigXml(Context context) throws OpenConfigException {
...
+              * and https://forum.syncthing.net/t/question-about-restartonwakeup-setting/2222/11
+              */
+            Element options = (Element) mConfig.getDocumentElement()
+                    .getElementsByTagName("options").item(0);
+            changed = setConfigElement(options, "restartOnWakeup", "true") || changed;
+
+            // Save changes if we made any.
+            if (changed) {
+                saveChanges();
+            }
```
AudriusButkevicius: I don't think any change is needed here, as this sets it to the default value anyway.
Also, if we ever decide that the default should be the other way around, doing it under isFirstStart is not a good idea.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -110,9 +138,69 @@ public String getUserName() {
      */
     @SuppressWarnings("SdCardPath")
     public void updateIfNeeded() {
-        Log.i(TAG, "Checking for needed config updates");
         boolean changed = false;
+
+        /* Get preference - PREF_USE_FOLDER_OBSERVER */
+        boolean prefUseFolderObserver = false;
```
AudriusButkevicius: Don't need this, just

```
boolean prefUseFolderObserver = mPreferences.getBoolean(Constants.PREF_USE_FOLDER_OBSERVER, false);
```

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -110,9 +138,69 @@ public String getUserName() {
...
+            Element r = (Element) folders.item(i);
+
+            // Enable "fsWatcherEnabled" attribute.
+            if (!r.hasAttribute("fsWatcherEnabled") ||
+                    !Boolean.parseBoolean(r.getAttribute("fsWatcherEnabled"))) {
+              Log.i(TAG, "Set 'fsWatcherEnabled' on folder " + r.getAttribute("id"));
+              r.setAttribute("fsWatcherEnabled", Boolean.toString(true));
+              changed = true;
+            }
+          }
```
AudriusButkevicius: This section that tweaks the fsWatcherEnabled can go away, and just be handled below in the if (prefUseFolderObserver) branch, removing the if and always doing
```
r.setAttribute("fsWatcherEnabled", Boolean.toString(!prefUseFolderObserver));
```

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -156,17 +245,37 @@ public void updateIfNeeded() {
...
 
+        /* Check if we have to dismiss any specific "unackedNotificationID" */
+        /* Dismiss "fsWatcherNotification" according to https://github.com/syncthing/syncthing-android/pull/1051 */
+        if (getConfigElement(options, "unackedNotificationID").contains("fsWatcherNotification")) {
+          Log.i(TAG, "Remove option 'unackedNotificationID' to dismiss 'fsWatcherNotification'.");
+          // According to review, it is sufficient to remove all "unackedNotificationID" contents.
+          // changed = setConfigElement(options, "unackedNotificationID", getConfigElement(options, "unackedNotificationID").replace("fsWatcherNotification", "")) || changed;
+          options.removeChild(options.getElementsByTagName("unackedNotificationID").item(0));
+          changed = true;
+        }
```
AudriusButkevicius: This is not handled correctly, there can be multiple `unackedNotificationID` elements, each with a different value.
You check it seems only the first one, and remove only the first one.

app/src/main/res/values/strings.xml
```diff
@@ -344,6 +344,14 @@ Please report any problems you encounter via Github.</string>
 
     <string name="use_legacy_hashing_summary">Force Syncthing to use legacy hashing package for compatibility purposes</string>
 
+    <string name="use_folder_observer_title">Use FolderObserver</string>
+
+    <string name="use_folder_observer_summary">Use wrapper\'s FolderObserver for realtime change detection. Syncthing core\'s fsWatcher will be disabled on all folders.</string>
```
AudriusButkevicius: These terms (FolderObserver, fsWatcher) make a lot of sense if you read the code, yet the user has no fucking clue what these are and what these do, so the description should be more human.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingService.java
```diff
@@ -221,22 +221,32 @@ protected void onPostExecute(Void aVoid) {
...
+        
+        boolean prefUseFolderObserver = mPreferences.getBoolean(Constants.PREF_USE_FOLDER_OBSERVER, false);
+        if (prefUseFolderObserver) {
+          Log.i(TAG, "onSyncthingStarted(): Enumerating folders to create FolderObserver instances.");
+          
+          Handler handler = new Handler();
+          new Thread(() -> {
+              for (Folder r : mApi.getFolders()) {
+                  try {
+                      mObservers.add(new FolderObserver(mApi, r, handler));
```
AudriusButkevicius: Just out of curiosity, in case you know, how do these threads get cleaned up?

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -261,6 +265,7 @@ public void onApiChange(SyncthingService.State currentState) {
             mRelaysEnabled.setChecked(mOptions.relaysEnabled);
             mGlobalAnnounceServers.setText(joiner.join(mOptions.globalAnnounceServers));
             mAddress.setText(mGui.address);
+            mRestartOnWakeup.setChecked(mOptions.restartOnWakeup);
```
AudriusButkevicius: I don't see this being user anywhere where it would actually set the setting?

AudriusButkevicius (changes requested):
Also, why are we bumping version of syncthing?

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
Nutomic: Agreed, it doesnt make sense to log app preferences, we can just ask users for them.

Or we could decide to log all app preferences, but that would have to be a seperate PR. But not even Syncthing does that.

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
Catfriend1: Right, this would apply to many prefs. Looking through the code, only some are implemented with a constant. I would prefer doing the individual string to constant improvement in a new PR.

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
Catfriend1: useFolderObserver is toggled by the user and does only affect the wrapper. Therefore we have the ChangeListener.
mRestartOnWakeup is derived from syncthing core, see Options.java (if I remember it correct), so the app just reflects what is set within syncthing core. If changed in the wrapper's UI, it is written back to syncthing core and the core is restarted.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingService.java
Catfriend1: Its more down in the "shutdown" function:
Stream.of(mObservers).forEach(FolderObserver::stopWatching);
mObservers.clear();

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: Ok, took it out (committing soon)

app/src/main/res/values/strings.xml
Catfriend1: ... so I submit a new suggestion.:)

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: Ok, switched back to my first version only removing the string "fsWatcherNotification" from the unackedNotificationID node.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: Agreed. Good point.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: Much more easier! I was little blinded by Nutomics strategy only writing the config back if it was changed and closely tracking changes to the config - but you're right - assuming we do a migration from version 27>28 once, we are sure we change the config during this step and we must write it back. So indeed the extra code for "only logging and saving if the config changed" can be removed here.

Catfriend1:
@AudriusButkevicius The manual bump of the version was advised by calmh in https://github.com/syncthing/syncthing/issues/4882#issuecomment-381749550

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: Setting it the opposite of "prefUseFolderObserver" has a disadvantage from my point of view as the users of the web UI are forced to live with or without the file watcher. On my point of view, there should be more liberty by:
a) Forcing file watcher on during one-time migration for all folders and then leaving the decision of the filewatcher on/off up to the user
b) If FolderObserver is enabled in the app settings UI, the user is "helped" in preventing a manual misconfiguration by force-turning off the fsWatcher for all folders. This is also stated in the explanation text under the experimental option checkbox.

Catfriend1:
@AudriusButkevicius Thanks for your feedback on this. Can you have a look again, please?

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -110,9 +124,65 @@ public String getUserName() {
...
+          for (int i = 0; i < folders.getLength(); i++) {
+            Element r = (Element) folders.item(i);
+
+            // Disable "fsWatcherEnabled" attribute.
+            if (r.hasAttribute("fsWatcherEnabled")) {
+              r.setAttribute("fsWatcherEnabled", Boolean.toString(false));
+              changed = true;
+            }
+          }
+        }
```
Nutomic: Your formatting is off compared to existing code (blocks should be indented by 4 spaces not 2). Same thing in other places in this PR. There should be a setting for this in your editor.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -189,22 +277,38 @@ private Element getGuiElement() {
...
+
+                /**
+                  * Only alter the first occurency of the device tag in assumption it is the device running this instance.
+                  * Reason: Sometimes encountered a bug when manually restored syncthing db+config with root explorer
+                  *         where app started and all devices got their name altered to the model name. Config was messed up then.
+                  *         This should work around the problem at least until a better fix will be implemented.
+                  * ToDo:   Implement a better fix by reading the device ID first, e.g. from REST API and only alter the correct
+                  *         node's device name attribute.
+                  */
+                return true;
```
Nutomic: Maybe put this in a seperate PR? And I would put the long explanation into the javadoc for the method, not inside the function. Or just describe the problem in a github issue and add the issue link.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: Thanks for taking the time. I corrected intendations and moved the migration code to a separate function.

Catfriend1:
@AudriusButkevicius Ok, I reverted the fix in changeLocalDeviceName back to the original code. Ok now to merge?

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -156,17 +200,76 @@ public void updateIfNeeded() {
...
         Element options = (Element) mConfig.getDocumentElement()
                 .getElementsByTagName("options").item(0);
         changed = setConfigElement(options, "weakHashSelectionMethod", "never") || changed;
 
+        /* Check if we have to dismiss any specific "unackedNotificationID" */
+        /* Dismiss "fsWatcherNotification" according to https://github.com/syncthing/syncthing-android/pull/1051 */
+        if (getConfigElement(options, "unackedNotificationID").contains("fsWatcherNotification")) {
+            Log.i(TAG, "Remove 'fsWatcherNotification' from 'unackedNotificationID' node.");
+            changed = setConfigElement(options, "unackedNotificationID", getConfigElement(options, "unackedNotificationID").replace("fsWatcherNotification", "")) || changed;
+        }
```
AudriusButkevicius: I expect this still doesn't do what you expect it to.
The resulting XML is like this:

```
<options>
   <foo>11</foo>
   <baz>Nuff</baz>
   <unackedNotificationID>notification1</unackedNotificationID>
   <unackedNotificationID>notification2</unackedNotificationID>
</options>
```

So I am not sure what it would be removing in this case.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -110,9 +124,38 @@ public String getUserName() {
...
+        mPreferences = PreferenceManager.getDefaultSharedPreferences(mContext);
+        boolean prefUseFolderObserver = mPreferences.getBoolean(Constants.PREF_USE_FOLDER_OBSERVER, false);
+        if (prefUseFolderObserver) {
+            Log.i(TAG, "Disabling fsWatcher on all folders because experimental option to use FolderObserver was manually set.");
+            for (int i = 0; i < folders.getLength(); i++) {
+                Element r = (Element) folders.item(i);
+
+                // Disable "fsWatcherEnabled" attribute.
+                if (r.hasAttribute("fsWatcherEnabled")) {
+                    r.setAttribute("fsWatcherEnabled", Boolean.toString(false));
```
AudriusButkevicius: Don't we want to enable it if the internal watcher is disabled?

Because it seems that toggling the `prefUseFolderObserver` post config migration will lead to a state where the syncthings own watcher cannot be enabled under any circumstances via the app.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -156,17 +200,76 @@ public void updateIfNeeded() {
...
+            }
+
+            /**
+            * Set config version to 28 after manual config migration
+            * This prevents "unackedNotificationID" getting populated
+            * with the fsWatcher GUI notification.
+            */
+            iConfigVersion = 28;
+            mConfig.getDocumentElement().setAttribute("version", Integer.toString(iConfigVersion));
+            Log.i(TAG, "Config version " + Integer.toString(iConfigVersion) + " reached.");
```
AudriusButkevicius: Probably want to move the attribute setting and logging juist before the return, as there might be more than one migration, so no point setting the version attribute in each migration, as you can just adjust the stack variable in each individual migration.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -189,6 +292,8 @@ private Element getGuiElement() {
      *
      * We need to iterate through XML nodes manually, as mConfig.getDocumentElement() will also
      * return nested elements inside folder element.
+     *
+     * Returns if changes to the config have been made.
```
AudriusButkevicius: This is now incorrect.

AudriusButkevicius (changes requested):

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: Removed.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: Good thought :)

Catfriend1: Also found a problem and fixed it, too. "changed = true" must not live in the for. Assume no folder would have changed because a user enabled fsWatcher manually before the migration, the config would then not have been written back because changed would still have been "false" beside we bumped the version number.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: I verified the setConfigElement....replace working correctly just removing the one desired fsWatcher word from the unackedNotificationID string in an earlier debug build containing this code snippet. That can be explained by looking at setConfigElement. First, the "if" clause is true, because the word was found - that is when unackedNotificationID exists and contains a string - so setConfigElement will not add a new node instead it will update the existing unackedNotificationID node string. All fine :)

imsodin: To me it looks like you just remove the `fsWatcherNotification` string within `unackedNotificationID` node. What you should do is remove the entire `unackedNotificationID` node that contains `fsWatcherNotification`.

AudriusButkevicius: But this only applies for the first element which might not always be the case and also leaves a tag with an empty string which is nit what we want.

Catfriend1: @imsodin I had a code suggestion to remove the whole node as advised, but @AudriusButkevicius did not agree this is the best way to do it. Removing just the one word can leave the empty string. If this is not okay, we can make a consent on deleting the whole node if it would end up empty.

Catfriend1: Improved that part detecting if removing or cleaning a word fist best and doing it.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Nutomic: If it's that complicated, I would just remove FolderObserver.java, always set fsWatcherEnabled and release it as a beta. If there are any problems, people will report them. The Go version is definitely better tested, so I dont think there will be any problems.

AudriusButkevicius: Yeah perhaps just remove the built in watcher.

Catfriend1: Ok, I will do some more testing on my phone with fsWatcherEnabled as its running fine since I installed the rc4 debug release. If there arent any problems will remove it. Would you suggest disabling the code (dont load it and make it hidden) or should I completely remove FolderObserver.java and relevant code out of the project?

AudriusButkevicius: Just remove the code, we have version control for history.

AudriusButkevicius: Just remove the code, we have version control for history.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -156,17 +200,84 @@ public void updateIfNeeded() {
...
+        /* Dismiss "fsWatcherNotification" according to https://github.com/syncthing/syncthing-android/pull/1051 */
+        /* If "unackedNotificationID" would end up with an empty string or whitespace, remove the whole node */
+        if (getConfigElement(options, "unackedNotificationID").contains("fsWatcherNotification")) {
+            Log.i(TAG, "Remove 'fsWatcherNotification' from 'unackedNotificationID' node.");
+            String unackedNotificationID = getConfigElement(options, "unackedNotificationID").replace("fsWatcherNotification", "");
+            if (TextUtils.isEmpty(unackedNotificationID.replaceAll("\\s+",""))) {
+                options.removeChild(options.getElementsByTagName("unackedNotificationID").item(0));
+            } else {
+                setConfigElement(options, "unackedNotificationID", unackedNotificationID);
+            }
```
imsodin: I think there is still a misconception here. There is no need to do manipulate any strings: One unackedNotification node contains only one notification ID. So in the beginning it could look like this:
```xml
   <unackedNotificationID>otherNotification1</unackedNotificationID>
   <unackedNotificationID>fsWatcherNotification</unackedNotificationID>
   <unackedNotificationID>otherNotification2</unackedNotificationID>
   <unackedNotificationID>otherNotification3</unackedNotificationID>
```
and the result should be 
```xml
   <unackedNotificationID>otherNotification1</unackedNotificationID>
   <unackedNotificationID>otherNotification2</unackedNotificationID>
   <unackedNotificationID>otherNotification3</unackedNotificationID>
```
So you need to get the node that contains "fsWatcherNotification", which is not necessarily `.item(0)`, and remove the node, not the "fsWatcherNotification" string.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: Done with commit https://github.com/syncthing/syncthing-android/pull/1051/commits/8d0a1dc0e36a3a99b95969f12de2d193b0f57377

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: I am on it. Incoming commit ...

Catfriend1:
Can anyone please give me a hint why the most recent changes break the Travis build vut TeamCity build still passes?
:app:lint
Ran lint on variant release: 112 issues found
Ran lint on variant debug: 112 issues found
Wrote HTML report to file:///home/travis/build/syncthing/syncthing-android/app/build/reports/lint-results.html
Wrote XML report to file:///home/travis/build/syncthing/syncthing-android/app/build/reports/lint-results.xml
:app:lint FAILED
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':app:lint'.
> Lint found errors in the project; aborting build.
  
  Fix the issues identified by lint, or add the following to your build script to proceed with errors:
  ...
  android {
      lintOptions {
          abortOnError false
      }
  }

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
Catfriend1: I replaced this one with a constant on the way.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: Corrected by commit https://github.com/syncthing/syncthing-android/pull/1051/commits/896829c21907d65e508a815d2953c3627b84c9b7

Catfriend1:
@Nutomic The logging shows the trigger of editFolder() and sendConfig() as I expected. But the attribute "type" (readonly, readwrite) is immediately saved to config.xml when leaving the FolderActivity UI and the "fsWatcherEnabled" is not. I have no clue why the http request does not save the change properly, looking at more folder options behind the UI, it could be a problem with parsing the boolean type - not sure if I am on the right track here. (html log file helped, thanks!)

Catfriend1:
@imsodin I analyzed the problem where thr user can can toggle fswatcher on off from the android apps ui. As Nutomic suggested I looked at the post config request. I see the folder type is correctly posted and written to config.xml by syncthing core. Problem is the fswatcherenabled attribite which is correctly posted to the api and does not get applied by syncthing core. Can this be a st core issue?
Here is what the app posts :


04-27 09:27:44.408 V/RestApi (7995): sendConfig: {"devices":[{"addresses":["dynamic"],"certName":"","compression":"metadata","deviceID":"QOVRVZX-RH4IIQS-VLSNM4D-YFQQPTJ-CLDJJO7-HRC4SSZ-TY3VENS-UT3WRA2","introducer":false,"name":"LG-H815"}],"folders":[{"autoNormalize":true,"copiers":0,"devices":[{"deviceID":"QOVRVZX-RH4IIQS-VLSNM4D-YFQQPTJ-CLDJJO7-HRC4SSZ-TY3VENS-UT3WRA2","introducer":false}],"disableSparseFiles":false,"disableTempIndexes":false,"fsWatcherEnabled":true,"fsync":true,"hashers":1,"id":"lg-h815-Bilder","ignoreDelete":false,"ignorePerms":true,"label":"Kamera","maxConflicts":-1,"minDiskFree":{"unit":"%","value":1.0},"order":"random","path":"/storage/emulated/0/DCIM","pullerPauseS":0,"pullerSleepS":0,"pullers":0,"rescanIntervalS":3600,"scanProgressIntervalS":0,"type":"readonly","versioning":{"params":{},"type":""}},{"autoNormalize":true,"copiers":0,"devices":[],"disableSparseFiles":false,"disableTempIndexes":false,"fsWatcherEnabled":true,"fsync":true,"hashers":0,"id":"jy87n-6jae8","ignoreDelete":false,"ignorePerms":true,"label":"D","maxConflicts":10,"path":"/storage/emulated/0/Download/","pullerPauseS":0,"pullerSleepS":0,"pullers":0,"rescanIntervalS":3600,"scanProgressIntervalS":0,"type":"readwrite","versioning":{"params":{}}}],"gui":{"address":"127.0.0.1:8384","apiKey":"FeDoWTZidRrkxTym4e32rUrFrs6w6JaL","enabled":true,"insecureAdminAccess":false,"password":"$2a$04$NPcssm2yFbvVnZWG2NuOs.wS1pTdOq2sNSYyd/sH3fu9iV.vpOuBu","theme":"default","useTLS":false,"user":"syncthing"},"ignoredDevices":[],"options":{"alwaysLocalNets":[],"autoUpgradeIntervalH":12,"cacheIgnoredFiles":false,"globalAnnounceEnabled":true,"globalAnnounceServers":["default"],"keepTemporariesH":24,"limitBandwidthInLan":false,"listenAddresses":["default"],"localAnnounceEnabled":true,"localAnnounceMCAddr":"[ff12::8384]:21027","localAnnouncePort":21027,"maxRecvKbps":0,"maxSendKbps":0,"minHomeDiskFreePct":0,"natEnabled":true,"natLeaseMinutes":60,"natRenewalMinutes":30,"natTimeoutSeconds":10,"overwriteRemoteDeviceNamesOnConnect":false,"progressUpdateIntervalS":5,"reconnectionIntervalS":60,"relayReconnectIntervalM":10,"relaysEnabled":true,"releasesURL":"https://upgrades.syncthing.net/meta.json","restartOnWakeup":true,"startBrowser":true,"symlinksEnabled":false,"tempIndexMinBlocks":10,"urAccepted":3,"urInitialDelayS":1800,"urPostInsecurely":false,"urURL":"https://data.syncthing.net/newdata","urUniqueId":"SWLfZZfy"},"version":28}

AudriusButkevicius:
Use chrome debugger tools to check what the web UI sends.

Catfriend1:
@AudriusButkevicius It results in a successful return from the REST call. Looking at /lib/config/folderconfiguration.go I noticed that fsWatcherEnabled gets "hard-set" to false if no fsWatcherDelay is given in the POST. Hey, that was the case. So I'll have to update the Folder model in the android app > seems not to be an issue with syncthing-core.

Catfriend1:
@AudriusButkevicius I tested this again on my phone and confirmed all parts of this PR working properly. In case you wonder, I've removed the string this time using Android Studio - the left-over comments in other languages will be fixed when tx pull is done later after merge as we've discussed before. LGTM?

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -156,17 +177,85 @@ public void updateIfNeeded() {
...
+            * Set config version to 28 after manual config migration
+            * This prevents "unackedNotificationID" getting populated
+            * with the fsWatcher GUI notification.
+            */
+            iConfigVersion = 28;
+            changed = true;
+        }
+
+        mConfig.getDocumentElement().setAttribute("version", Integer.toString(iConfigVersion));
+        Log.i(TAG, "New config version is " + Integer.toString(iConfigVersion));
```
AudriusButkevicius: This could be under a `if (changed)` as otherwise it will always print new config version, which might not change in most cases.
The other (potentially better) option would be to have `iOldConfigVersion` and `iConfigVersion` and only update and print stuff if old does not match new.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -156,17 +177,85 @@ public void updateIfNeeded() {
...
+
+        return changed;
+    }
+
+    private String getConfigElement(Element parent, String tagName) {
```
AudriusButkevicius: Don't see this being used.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -156,17 +177,85 @@ public void updateIfNeeded() {
...
+
+            /* Enable fsWatcher for all folders */
+            NodeList folders = mConfig.getDocumentElement().getElementsByTagName("folder");
+            for (int i = 0; i < folders.getLength(); i++) {
+                Element r = (Element) folders.item(i);
+
+                // Enable "fsWatcherEnabled" attribute and set default delay.
+                Log.i(TAG, "Set 'fsWatcherEnabled', 'fsWatcherDelayS' on folder " + r.getAttribute("id"));
+                r.setAttribute("fsWatcherEnabled", Boolean.toString(true));
+                r.setAttribute("fsWatcherDelayS", Integer.toString(10));
```
AudriusButkevicius: Either use static strings here and in changeDefaultFolder, or use converted types in both places please.

app/src/main/res/values-bg/strings.xml
```diff
@@ -256,8 +256,7 @@
   <string name="syncthing_active">Syncthing работи</string>
   <string name="syncthing_disabled">Syncthing не работи</string>
   <!--Toast shown if folder observer fails to traverse a folder-->
-  <string name="toast_folder_observer_stack_overflow">Твърде много папки. Проверяване за cyclic symlinks</string>
-  <!--Toast shown if syncthing failed to create a config-->
+    <!--Toast shown if syncthing failed to create a config-->
```
AudriusButkevicius: This is incorrectly indented, in all files

I still don't understand why you keep messing around with auto-generated. Just change the english strings and leave the others around which will get updated automatically.
PLease

AudriusButkevicius (changes requested):
Also, revert your syncthing update, as we are currently on the latest released version, and you bumping syncthing version to some unreleased version will cause a disaster.

app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
```diff
@@ -423,13 +432,13 @@ private void initFolder() {
...
-            // Scan every 3 days (in case inotify dropped some changes)
-            mFolder.rescanIntervalS = 259200;
+            // Scan every hour (in case real time change detection failed)
+            mFolder.rescanIntervalS = 3600;
         }
         else {
             // FileObserver is broken on Marshmallow.
             // https://github.com/syncthing/syncthing-android/issues/787
-            mFolder.rescanIntervalS = 60;
+            mFolder.rescanIntervalS = 300;
```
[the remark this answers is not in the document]
Catfriend1: I found this consumed too much battery during testing because if you have very large folders syncing to a mobile device the scan will take half of the scan interval time and the rescan will soon occur. I would advise, as our app users cannot easily tweak the rescan interval, giving them a more "lightweight" interval. With the file watcher coming up soon, users that need instant-syncing of files can (and will get) the watcher turned on to fit their needs. This is more a strategy, consider a user turning off instant-sync in the UI it means "the user has time to wait for syncing".

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: ok, I chose the "better" option.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: removed.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: ok

app/src/main/res/values-bg/strings.xml
Catfriend1: oh sh*** I forgot to set the indent for this project in android studio. So automatically done yes, but incorrectly configured tool.

Catfriend1:
@AudriusButkevicius I guess you will take care of the syncthing submodule update before releasing the master branch, so I can just hold back those syncthing-update commits and test locally. Is the PR okay now?

AudriusButkevicius:
Yeah, I'll update to .47 once it's released.

app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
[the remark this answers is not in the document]
imsodin: Isn't that marshmallow problem FolderObserver specific? Such that now with the FolderObserver removed, this condition is no longer needed and the rescan interval can be set the same for all android versions.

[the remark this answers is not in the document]
AudriusButkevicius: This is fallback scans isn't it?
I thought the watcher adjusts the interval automatically?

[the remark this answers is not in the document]
imsodin: No fallback, that idea was discarded early. Full scan interval is completely independent from fs watcher. Just in the edit folder dialog we propose default values depending on whether fs watcher is enabled or not.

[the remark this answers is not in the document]
Catfriend1: I made a new PR where we can decide how we want the interval. So far, I followed syncthing core where it is 3600s. https://github.com/syncthing/syncthing-android/pull/1069

app/src/main/res/values/strings.xml
```diff
@@ -141,6 +141,9 @@ Please report any problems you encounter via Github.</string>
     <!-- Setting title -->
     <string name="folder_master">Send Only</string>
 
+    <!-- Setting title -->
+    <string name="folder_fileWatcher">Realtime Sync</string>
```
AudriusButkevicius: This is such a poor name, we should change it. It's nowhere close to realtime, as there is a 10 second delay.
Should probably be "Watch for changes" with a text explaining that otherwise it takes an hour to detect a change (or pressing rescan button in Web UI).

Catfriend1:
Ok, should I make a PR for that? Need to look if we can add a summary for that UI item...

AudriusButkevicius:
Yeah.
I've also noticed that if you change settings in the web ui, they are not propagated back to the app. I thought the app handles configSaved events.

Catfriend1:
You observed it correctly. Thats why i am currently working on eventprocessor and extending the handler by folder stats and configsaved.

AudriusButkevicius:
Also, I think the release without tagging screwed up play store beta channel as even though the release is there, I am not getting it as a beta tester, had to install manually. It seems some of the publish steps or something picks up the version from the tag, so the apk is tagged with the old version.


Also had threeinteresting events, tweaking the folder manually, it seems I was prompted to restart syncthing in the web ui (which means something is screwed in syncthing and causes unnecessery restarts), and then post restart had permission denied on watches from syncthing, and then suddenly the app went into perma-loading mode.

Also it seems the logs are a bit of a hit and miss and are not always there.
I suspect it might be due to buffering.

Catfriend1:
I'm sorry to hear about the problems. I am currently testing a self-built debug apk from commit 0.10.9 without any problems on two phones. perma-loading sounds very bad as most likely the config got corrupted. I also changed the rescan interval on folders manually in the webgui and then quit the app and restarted it as I remember restarting syncthing core from the web UI did not work fine in previous versions of the app.

AudriusButkevicius:
It's not really problems, it's more like observations. I think I had to restart the phone to get it to work again.
I suspect it could be something like me abruptly killing the app, syncthing binary stays running in the background, and then next launch of the app fails as it can't start syncthing (nor can it connect to the existing one potentially idk?)

I don't think the config got corrupted as it worked ok after a reboot.

app/src/main/res/values/strings.xml
Catfriend1: Made https://github.com/syncthing/syncthing-android/pull/1077

Nutomic:
You always need to update versionCode and versionName if you make a new version, and Google Play should also show a warning if you don't change that. The git tag only matters if you use the prepare-release.sh script that I put in the repo.

Nutomic:
Yes that's what the script does. If you want to set the version name manually, you have to comment out the line in the script that changes `versionName`. And either run the script to increase `versionCode`, or comment that out too and set it manually in build.gradle.

PR #1054, opened 2018-04-17
Updated bg (Bulgarian) translation

ImAnnoying2:
Commit made via [Stringlate](https://lonamiwebs.github.io/stringlate/)


imsodin:
Please use https://www.transifex.com/syncthing/syncthing-android for translations.

AudriusButkevicius:
Transalations are handled using Transifex.

Issue #1060, opened 2018-04-24
Add possibility to reset delta indexes


imsodin:
For debugging purposes it would be nice if there was an option like the existing "Reset Database" in the debug menu also for resetting the delta indexes. As there is no rest API endpoint for that, I am unable to do it on android (or admittedly just unwilling to find/try out how I could do it manually).

Nutomic:
Isnt there an environment variable for this? You should be able to pass that from the app settings.

imsodin:
Not that I am aware of. There is the command line option `-reset-deltas` similar to `-reset-database`. I assume the mentioned "Reset Database" just restarts the syncthing binary with `-reset-database`, so a new option "Reset Delta Indexes" could likely reuse that by replacing the command line parameter. to `-reset-deltas`.

Catfriend1:
I am currently doing this in a new PR ...

PR #1061, opened 2018-04-25
reset delta indexes from UI

ghost:
This should address the enhancement described in https://github.com/syncthing/syncthing-android/issues/1060

Quote
For debugging purposes it would be nice if there was an  option like the existing "Reset Database" in the debug menu also for  resetting the delta indexes. As there is no rest API endpoint for that, I  am unable to do it on android (or admittedly just unwilling to find/try  out how I could do it manually).

Question / ToDo
How can I remove those strings from all translation files (and not just from the english one)? Creating a transifex branch with "git push -s -b [branchname]" says Forbidden. I have Team1 level access at the moment. @Nutomic any advise for me?

.../res/values/strings.xml
`
<string name="streset_title">to be removed</string>
<string name="streset_question">to be removed</string>
<string name="streset_done">to be removed</string>`

Testing
Verified working on Android 7.1.2 @lg-h815.

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
@@ -80,7 +80,8 @@ public class SettingsActivity extends SyncthingActivity {
         private static final String KEY_STTRACE = "sttrace";
         private static final String KEY_EXPORT_CONFIG = "export_config";
         private static final String KEY_IMPORT_CONFIG = "import_config";
-        private static final String KEY_STRESET = "streset";
+        private static final String KEY_STRESETDATABASE = "stresetdatabase";
+        private static final String KEY_STRESETDELTAS = "stresetdeltas";
 
         @Inject NotificationHandler mNotificationHandler;
         @Inject SharedPreferences mPreferences;
@@ -184,7 +185,8 @@ public class SettingsActivity extends SyncthingActivity {
 
             Preference stTrace              = findPreference("sttrace");
             Preference environmentVariables = findPreference("environment_variables");
-            Preference stReset              = findPreference("streset");
+            Preference stResetDatabase      = findPreference("stresetdatabase");
+            Preference stResetDeltas        = findPreference("stresetdeltas");
 
             mUseRoot                     = (CheckBoxPreference) findPreference(Constants.PREF_USE_ROOT);
             Preference useWakelock       = findPreference(Constants.PREF_USE_WAKE_LOCK);
@@ -204,7 +206,8 @@ public class SettingsActivity extends SyncthingActivity {
 
             stTrace.setOnPreferenceChangeListener(this);
             environmentVariables.setOnPreferenceChangeListener(this);
-            stReset.setOnPreferenceClickListener(this);
+            stResetDatabase.setOnPreferenceClickListener(this);
+            stResetDeltas.setOnPreferenceClickListener(this);
 
             mUseRoot.setOnPreferenceClickListener(this);
             useWakelock.setOnPreferenceChangeListener((p, o) -> requireRestart());
@@ -376,6 +379,7 @@ public class SettingsActivity extends SyncthingActivity {
 
         @Override
         public boolean onPreferenceClick(Preference preference) {
+            final Intent intent;
             switch (preference.getKey()) {
                 case Constants.PREF_USE_ROOT:
                     if (mUseRoot.isChecked()) {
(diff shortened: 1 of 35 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -372,15 +372,31 @@ Please report any problems you encounter via Github.</string>
...
-    <string name="streset_question">This action should only be performed based on a recommendation from our support team.
-        \nAre you sure you want to clear Syncthing\'s index database?
+    <string name="stresetdatabase_question">This action should only be performed based on a recommendation from our support team.
+        \nAre you sure you want to clear Syncthing\'s database?
+    </string>
+
+    <!-- to be removed after merge into the master -->
+    <string name="streset_title">to be removed</string>
+    <string name="streset_question">to be removed</string>
+    <string name="streset_done">to be removed</string>
```
AudriusButkevicius: Why not remove now?

AudriusButkevicius:
Just remove the translations, we'll update them as we release.

AudriusButkevicius:
lgtm, but I'll leave this to timeout until the weekend for @Nutomic or @capi to review, if none have time, I guess we'll just merge.

app/src/main/res/values/strings.xml
Catfriend1: Hi,
thanks for taking time.
Removing the strings in the en-lang causes the build to fail on travis and teamcity.
I would have to remove them in all lang files manually one by one for the build to be okay.
Which way is correct - leaving the build failing or going through all 36 lang files?
I tried to create a transifex branch (that can merge later synced with the github merge) but it says my access level is not sufficient (Error forbidden) - is this a better way to do it?
I wonder how @Nutomic's procedure was when strings had to be removed.

AudriusButkevicius: There might be a command to regenerate transifex stuff. What's the failure when you remove strings?

imsodin: I believe the right thing to do is not rename the existing strings. The proposed keys might be better, but it means all translations must be done again. I.e. Just add the `stresetdelta...` strings and leave the existing `streset...` alone.

Catfriend1: I'll post the error tomorrow and then make a commit with resetdb renamed back to streset.

Nutomic: Android Studio has a "translation editor" which allows you to remove all translations for a single string. Just open any strings.xml file in Android Studio, and it should show a notification at the top of the editor. Then select the string you want to remove, and press the red minus at the top.

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -418,16 +422,31 @@ public boolean onPreferenceClick(Preference preference) {
                             .setNegativeButton(android.R.string.no, null)
                             .show();
                     return true;
-                case KEY_STRESET:
-                    final Intent intent = new Intent(getActivity(), SyncthingService.class)
-                            .setAction(SyncthingService.ACTION_RESET);
+                case KEY_STRESETDATABASE:
```
Nutomic: You should put an underscore, like `KEY_ST_RESET_DATABASE` (also for the action string below).

app/src/main/res/values/strings.xml
Catfriend1: Thanks for the hint @Nutomic. I chose to remove the original streset strings because they referred to "index" in lang-en. This can be confusing, thus just a little bit, if talking about index and delta index on two UI buttons. I think its more straightline if we stick to syncthing doc's and call it database / delta indexes reset. Commit https://github.com/syncthing/syncthing-android/pull/1061/commits/8a33d007fc371ce4f849d4dff7445ba0682b8943

capi:
Also looks good for me in the diff viewer. Need to try when in front of my workstation again. But won't be before next week, so don't necessarily wait for me on that.

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
Catfriend1: Ok, commit https://github.com/syncthing/syncthing-android/pull/1061/commits/21fd135c264bcdb0a7ee5cc9d6bba5daca1ef560

Catfriend1:
@AudriusButkevicius  I read the transifex help pages. If I understand correctly, the person that has full access to the transifex push/pull/branch functions can skip the changes made to the other lang's strings.xml and overwrite it. Transifex only imports the primary lang-en and will - hopefully - trigger a new translation for the two reset strings as I find this behaviour appropriate (see reason above)

AudriusButkevicius:
Sure, but then why faff around with other language files in this PR?

Catfriend1:
@audriusbutkevicius I intended to fix the build as we currently dont have a CI with transifex including a proper setup branch for that.

AudriusButkevicius:
What does the build fail with if you have extra strings in some non-main translations?

Catfriend1:
It tells exactly that you are missing a string in the base languages strings.xml

AudriusButkevicius:
Did the IDE remove the strings or you manually?

Catfriend1:
I removed the strings in lang-en/Strings.xml manually, so they were still in the files of other languages until I renamed them.

AudriusButkevicius:
Why don't you just use the IDE to do the right thing. I am not sure if faffing around manually by hand will do the right thing.

Catfriend1:
Sorry, I didn't mean to mess with it. I'll compare results now to the result I'm getting using the IDE. If necessary I'll commit.

Catfriend1:
@AudriusButkevicius The windows IDE is quite huge to setup beside my linux machine. I cannot see a difference producing the files in the github editor instead of IDE. Please reconsider letting us focus on how we send valid source to transifex without creating conflicts between branches. As this is a minor change, I'd even go as far as "let's test it". If something breaks unexpectedly when merging this, we can go back (as transifex still knows the old string) or recommit with a transifex in place (would require getting priviledges from @calmh on transifex repo)

Nutomic:
Just make sure `gradle lint` passes (it will show any translation errors), and then run `tx push` in the repo with the Transifex client. Renaming strings isn't a problem really.

Catfriend1:
@Nutomic:  Thanks for clarifying the translation process. gladle lint passes without translation errors, see attached log.
[lint-results.zip](https://github.com/syncthing/syncthing-android/files/1954084/lint-results.zip)
@AudriusButkevicius : Do we already have a maintainer responsible for the translation push and pull that needs to be done after merge?

AudriusButkevicius:
So I think I was granted access, and I'll do that before the next release.

AudriusButkevicius:
Other than that one nitpick, looks good to me.

Catfriend1:
Thanks, fixed it.

AudriusButkevicius:
Thanks for the PR 🎉

Issue #838, opened 2017-01-25
Most File Managers don't support Open Folder Intent


Nutomic:
https://github.com/syncthing/syncthing-android/issues/831#issuecomment-271219787

https://forum.syncthing.net/t/poll-open-folder-button-and-file-managers/9067

Nutomic:
[This Stackoverflow thread](https://stackoverflow.com/questions/17165972/android-how-to-open-a-specific-folder-via-intent-and-show-its-content-in-a-file) has some more options, like different mime types. Might be worth a try

ping @PaulAnnekov

PaulAnnekov:
I googled all the web and checked many SO questions. Immediately remove answers with `/ACTION_GET_CONTENT/`, because it's wrong action. Of course it can show you the contents, but when you will click on any file - activity will close and return you a link to selected file, because that's the purpose of [ACTION_**GET**_CONTENT](https://developer.android.com/reference/android/content/Intent.html#ACTION_GET_CONTENT).

And `ACTION_VIEW` is not supported by most of the file managers + it's crashing. So I still think we should better integrate own file manager, because Android will tighten the screws more and more. Most sync software (Google Drive/Dropbox) have their own files view, so it won't damage UX.

Nutomic:
Did you try the mime type `*/*` or `text/csv` with `ACTION_VIEW` already? Integrated file manager would be okay if its not too much effort, but it would be much better if we could also support the user's chosen file manager.

PaulAnnekov:
I tried `*/*` and `text/csv` with `ACTION_VIEW`. 

`*/*` + `ACTION_VIEW` means - "show me intents that can view at least one mime type", so you get **EVERYTHING** (Contact viewers, Google Play Store, Music players, file explorers, emails apps...).

`text/csv` + `ACTION_VIEW` means "show me intents that can view files with csv mime type", so you get listed all text editors (ES Note Editor, FE Text Editor, FX Text Viewer, Totalcmd-Editor...).

pchampin:
This is a hard choice here:
* either guarantee a smoother user experience with a built-in file manager, but locking at the same time the users in that view,
* or leave the user free to use their own file manager, but with a rougher interface...
There's no surprise that Google or Dropbox would chose the first option; locking the user in is part of their business model. But for an open-source project, it is trickier.

[...]

PR #1066, opened 2018-04-27
fix opening file explorer apps

ghost:
Purpose
Fix issue https://github.com/syncthing/syncthing-android/issues/838
Most file explorers do not implement the "resource/folder" mime type. If the user has such a file explorer installed, e.g. Root Explorer, OI file manager, a second try to find a file manager app that can handle opening the folder is attempted by using the chooserIntent. It will popup a menu like "choose which app you want to use to explore the folder".

Testing
Verified working with root explorer and OI file manager on Android 7.1.2 @lg-h815

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
@@ -7,6 +7,7 @@ import android.net.Uri;
 import android.support.annotation.NonNull;
 import android.support.v4.content.ContextCompat;
 import android.text.TextUtils;
+import android.util.Log;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
This doesn't explain why/how this fixes it.

Catfriend1:
@AudriusButkevicius I put a brief summary of the problem from the linked issue I intended to fix with this PR on top.

AudriusButkevicius:
You said that you've implemented a fallback mechanism, but I don't understand why using `application/*` mime type suddenly fixes this.

Catfriend1:
@AudriusButkevicius I checked using the debug build (https://build.syncthing.net/repository/download/SyncthingAndroid_Build/19963:id/debug/app-debug.apk). The code detects in resolveActivity that the result activitiy is null (no file manager app found installed that is compatible to folder/resource mime type). Before the change, the code showed a toast error and stopped. Now, it retries with the application/* mime type and the chooser Intent. That can be interpreted as an alternative way with a broader-matching filter. Now it finds a bunch of apps compatible to open (Root Explorer, QuickPic, OI, ...) and shows a selection UI to the user (this is done by ChooserIntent which is from android itself).

AudriusButkevicius:
I guess this doesn't make it any worse.

PR #1069, opened 2018-04-27
Adjust folder rescan interval

ghost:
Purpose
The folder rescan interval in seconds should be unified and adjusted so it's the same on all android versions the app runs on.

imsodin
I mean I am not sure that our fs watcher actually works on Android 6/Marshmallow, but I assume so, as Nutomic mentioned it as an advantage of the fs watcher.
I'd go for the same defaults as Syncthing: 1h/3600. I see no reason to scan more often, especially on generally slower devices like phones. And yes, sure, PR would be nice ;)

AudriusButkevicius
This is fallback scans isn't it?
I thought the watcher adjusts the interval automatically?

imsodin
No fallback, that idea was discarded early. Full scan interval is completely independent from fs watcher. Just in the edit folder dialog we propose default values depending on whether fs watcher is enabled or not.

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
@@ -431,6 +431,7 @@ public class FolderActivity extends SyncthingActivity
                 ? getIntent().getStringExtra(EXTRA_FOLDER_ID)
                 : generateRandomFolderId();
         mFolder.label = getIntent().getStringExtra(EXTRA_FOLDER_LABEL);
+        // Folder rescan interval defaults to 3600s as it is the default in syncthing.
         mFolder.rescanIntervalS = 3600;
         mFolder.versioning = new Folder.Versioning();
     }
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
```diff
@@ -431,15 +431,7 @@ private void initFolder() {
...
-        if (Build.VERSION.SDK_INT != Build.VERSION_CODES.M) {
-            // Scan every hour (in case real time change detection failed)
-            mFolder.rescanIntervalS = 3600;
-        }
-        else {
-            // FileObserver is broken on Marshmallow.
-            // https://github.com/syncthing/syncthing-android/issues/787
-            mFolder.rescanIntervalS = 300;
-        }
+        mFolder.rescanIntervalS = 3600;
```
AudriusButkevicius: Add a comment explaining why we are setting this. Also, shouldn't we set this to 60s if the watcher is disabled?

AudriusButkevicius: Well the comment is still a lie, as default is 60 and suggested when watcher is enabled is 3600

imsodin: This sets the default values for folders created in the app, right? Where is the fs watcher enabled? I think the interval should be set at the same place.

AudriusButkevicius: It's always enabled and this sets it for all folders unconditionally

Catfriend1: @imsodin: Yes, only folders created in the app get this rescanInterval as default. see FolderActivity.java line 155.

Issue #766, opened 2016-10-19
Set proxy from within app


benshep:
I would like the ability to set a proxy within the app. On Windows (and presumably Linux) it's straightforward to set proxies using environment variables. On Android, not so much. Can we have this as a setting in the Settings menu please?

ProactiveServices:
At the risk of hijacking this issue, if we could set any of the relevant environment variables it'd be handy. I currently need to use the STHASHING envvar to diagnose a possible hashing problem but can't.

Nutomic:
@ProactiveServices That makes sense. Can you open a seperate issue please?

Catfriend1:
@benshep This can only be achieved with the correct proxy exception as this would else break the web UI display in the app. I am not sure, how Syncthing core would get the proxy information being set on android app's side.
Would a linux env help here to tell st core to use a proxy @AudriusButkevicius?

Catfriend1:
https://docs.syncthing.net/users/proxying.html says the following env vars are respected by syncthing.
export all_proxy=socks5://10.20.30.40:5060
export http_proxy=...
export https_proxy=...

PR #1071, opened 2018-04-28
Add proxy settings to settings UI

ghost:
Purpose
This PR should add the enhancement from https://github.com/syncthing/syncthing-android/issues/766 to the experimental settings in the UI.  It generates two new EditText's to set HTTP(S) and SOCKS-Proxy according to syncthing doc syntax (https://docs.syncthing.net/users/proxying.html). If the existing experimental option use_tor is checked, it overrides the http/socks proxy and greyes out the EditText's.

Testing
Verified working on my Android 7.1.2 phone (lg-h815). Syncthing core shows the appropriate log entry ("proxy settings detected") when proxy is set in app's UI settings.

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
@@ -110,6 +110,7 @@ public class SettingsActivity extends SyncthingActivity {
 
         /* Experimental options */
         private CheckBoxPreference mUseRoot;
+        private CheckBoxPreference mUseTor;
         private EditTextPreference mSocksProxyAddress;
         private EditTextPreference mHttpProxyAddress;
 
@@ -195,8 +196,8 @@ public class SettingsActivity extends SyncthingActivity {
             Preference stResetDeltas        = findPreference("st_reset_deltas");
 
             mUseRoot                        = (CheckBoxPreference) findPreference(Constants.PREF_USE_ROOT);
-            Preference useWakelock          = findPreference(Constants.PREF_USE_WAKE_LOCK);
-            Preference useTor               = findPreference(Constants.PREF_USE_TOR);
+            Preference useWakelock          = (CheckBoxPreference) findPreference(Constants.PREF_USE_WAKE_LOCK);
+            mUseTor                         = (CheckBoxPreference) findPreference(Constants.PREF_USE_TOR);
             mSocksProxyAddress              = (EditTextPreference) findPreference(Constants.PREF_SOCKS_PROXY_ADDRESS);
             mHttpProxyAddress               = (EditTextPreference) findPreference(Constants.PREF_HTTP_PROXY_ADDRESS);
 
@@ -220,8 +221,11 @@ public class SettingsActivity extends SyncthingActivity {
             /* Experimental options */
             mUseRoot.setOnPreferenceClickListener(this);
             useWakelock.setOnPreferenceChangeListener((p, o) -> requireRestart());
-            useTor.setOnPreferenceChangeListener((p, o) -> requireRestart());
+            mUseTor.setOnPreferenceChangeListener(this);
+
+            mSocksProxyAddress.setEnabled(!(Boolean) mUseTor.isChecked());
             mSocksProxyAddress.setOnPreferenceChangeListener(this);
+            mHttpProxyAddress.setEnabled(!(Boolean) mUseTor.isChecked());
             mHttpProxyAddress.setOnPreferenceChangeListener(this);
 
             /* Initialize summaries */
@@ -405,6 +409,11 @@ public class SettingsActivity extends SyncthingActivity {
                                 .show();
                         return false;
                     }
+                    break;
+                case Constants.PREF_USE_TOR:
+                    mSocksProxyAddress.setEnabled(!(Boolean) o);
+                    mHttpProxyAddress.setEnabled(!(Boolean) o);
+                    requireRestart();
                     break;
                 case Constants.PREF_SOCKS_PROXY_ADDRESS:
                     if (handleSocksProxyPreferenceChange(preference, o.toString().trim())) {
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
@@ -230,8 +230,8 @@ public class SettingsActivity extends SyncthingActivity {
 
             /* Initialize summaries */
             mPreferences = PreferenceManager.getDefaultSharedPreferences(getActivity());
-            handleSocksProxyPreferenceChange (screen.findPreference(Constants.PREF_SOCKS_PROXY_ADDRESS),  mPreferences.getString(Constants.PREF_SOCKS_PROXY_ADDRESS, ""));
-            handleHttpProxyPreferenceChange (screen.findPreference(Constants.PREF_HTTP_PROXY_ADDRESS), mPreferences.getString(Constants.PREF_HTTP_PROXY_ADDRESS, ""));
+            handleSocksProxyPreferenceChange(screen.findPreference(Constants.PREF_SOCKS_PROXY_ADDRESS),  mPreferences.getString(Constants.PREF_SOCKS_PROXY_ADDRESS, ""));
+            handleHttpProxyPreferenceChange(screen.findPreference(Constants.PREF_HTTP_PROXY_ADDRESS), mPreferences.getString(Constants.PREF_HTTP_PROXY_ADDRESS, ""));
 
             try {
                 appVersion.setSummary(getActivity().getPackageManager()
@@ -559,7 +559,7 @@ public class SettingsActivity extends SyncthingActivity {
             if (newValue.equals("")) {
                 preference.setSummary(getString(R.string.do_not_use_proxy) + " " + getString(R.string.generic_example) + ": " + getString(R.string.socks_proxy_address_example));
                 return true;
-            } else if (newValue.matches("^socks.://(\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}):(\\d{1,5})$")) {
+            } else if (newValue.matches("^socks5://(\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}):(\\d{1,5})$")) {
                 preference.setSummary(getString(R.string.use_proxy) + " " + newValue);
                 return true;
             } else {
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
@@ -559,10 +559,7 @@ public class SettingsActivity extends SyncthingActivity {
             if (newValue.equals("")) {
                 preference.setSummary(getString(R.string.do_not_use_proxy) + " " + getString(R.string.generic_example) + ": " + getString(R.string.socks_proxy_address_example));
                 return true;
-            } else if (newValue.matches("^socks5://" +
-                                        "(((([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))|" +
-                                        "((([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])))" +
-                                        ":(\\d{1,5})$")) {
+            } else if (newValue.matches("^socks5://.*:\\d{1,5}$")) {
                 preference.setSummary(getString(R.string.use_proxy) + " " + newValue);
                 return true;
             } else {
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -211,9 +218,20 @@ public void onActivityCreated(Bundle savedInstanceState) {
...
+
+            mSocksProxyAddress.setEnabled(!(Boolean) mUseTor.isChecked());
+            mSocksProxyAddress.setOnPreferenceChangeListener(this);
+            mHttpProxyAddress.setEnabled(!(Boolean) mUseTor.isChecked());
+            mHttpProxyAddress.setOnPreferenceChangeListener(this);
+
+            /* Initialize summaries */
+            mPreferences = PreferenceManager.getDefaultSharedPreferences(getActivity());
+            handleSocksProxyPreferenceChange (screen.findPreference(Constants.PREF_SOCKS_PROXY_ADDRESS),  mPreferences.getString(Constants.PREF_SOCKS_PROXY_ADDRESS, ""));
+            handleHttpProxyPreferenceChange (screen.findPreference(Constants.PREF_HTTP_PROXY_ADDRESS), mPreferences.getString(Constants.PREF_HTTP_PROXY_ADDRESS, ""));
```
AudriusButkevicius: Whats up with the space before the brace?

app/src/main/res/values/strings.xml
```diff
@@ -337,13 +340,29 @@ Please report any problems you encounter via Github.</string>
 
     <string name="export_config">Export Configuration</string>
 
+    <!-- Experimental options -->
+    <string name="use_proxy">Connect through proxy</string>
+    <string name="do_not_use_proxy">Do not connect through proxy</string>
```
AudriusButkevicius: "a proxy"

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -512,5 +549,43 @@ protected void onPostExecute(Boolean haveRoot) {
...
+        /**
+         * Handles a new user input for the SOCKS proxy preference.
+         * Returns if the changed setting requires a reboot.
+         */
+        private boolean handleSocksProxyPreferenceChange(Preference preference, String newValue) {
+            // Valid input is either a proxy adress or an empty field to disable the proxy.
+            if (newValue.equals("")) {
+                preference.setSummary(getString(R.string.do_not_use_proxy) + " " + getString(R.string.generic_example) + ": " + getString(R.string.socks_proxy_address_example));
+                return true;
+            } else if (newValue.matches("^socks.://(\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}):(\\d{1,5})$")) {
```
AudriusButkevicius: This can be socks5://host.com:8080

app/src/main/res/values/strings.xml
```diff
@@ -337,13 +340,29 @@ Please report any problems you encounter via Github.</string>
...
     <string name="keep_wakelock_while_binary_running">Keep the CPU awake while Syncthing is running</string>
 
     <string name="keep_wakelock_while_binary_running_summary">Use this setting if you experience unexpected disconnects while operating on battery. This will result in increased battery consumption.</string>
 
     <string name="use_tor_title">Use Tor</string>
 
-    <string name="use_tor_summary">Force all traffic through Tor for increased privacy. Requires Orbot</string>
+    <string name="use_tor_summary">Force all traffic through Tor for increased privacy. Requires Orbot. Disables proxy options.</string>
+
+    <string name="socks_proxy_address_title">SOCKS-Proxy</string>
```
AudriusButkevicius: Why the dash? Also, maybe we should say SOCKS5 proxy, as we don't support SOCKS4 etc.

Catfriend1: hmm that came on my intuition cause "germans like dash connected words". Removed it. Maybe later on the german translation ;)

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -512,5 +549,43 @@ protected void onPostExecute(Boolean haveRoot) {
...
+        /**
+         * Handles a new user input for the SOCKS proxy preference.
+         * Returns if the changed setting requires a reboot.
+         */
+        private boolean handleSocksProxyPreferenceChange(Preference preference, String newValue) {
+            // Valid input is either a proxy adress or an empty field to disable the proxy.
+            if (newValue.equals("")) {
+                preference.setSummary(getString(R.string.do_not_use_proxy) + " " + getString(R.string.generic_example) + ": " + getString(R.string.socks_proxy_address_example));
+                return true;
+            } else if (newValue.matches("^socks5://(\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}):(\\d{1,5})$")) {
```
AudriusButkevicius: The ip address pattern is still invalid, this can easily be a hostname rather than IP

Catfriend1: Ok, i took a RFC 1123 compliant pattern from https://stackoverflow.com/questions/106179/regular-expression-to-match-dns-hostname-or-ip-address and adjusted it to our needs. Verified working on phone and online reg ex tester using different DNS, FQDN, NetBIOS and IP samples (http://www.regexplanet.com/advanced/java/index.html)

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -559,7 +559,10 @@ private boolean handleSocksProxyPreferenceChange(Preference preference, String n
             if (newValue.equals("")) {
                 preference.setSummary(getString(R.string.do_not_use_proxy) + " " + getString(R.string.generic_example) + ": " + getString(R.string.socks_proxy_address_example));
                 return true;
-            } else if (newValue.matches("^socks5://(\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}):(\\d{1,5})$")) {
+            } else if (newValue.matches("^socks5://" +
+                                        "(((([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))|" +
+                                        "((([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])))" +
+                                        ":(\\d{1,5})$")) {
```
AudriusButkevicius: Can you just not check anything here?

Catfriend1: Whats wrong with a RFC compliant check? Don't be scared...

AudriusButkevicius: or just socks5://.*:\d{1,}

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -578,7 +581,10 @@ private boolean handleHttpProxyPreferenceChange(Preference preference, String ne
             if (newValue.equals("")) {
                 preference.setSummary(getString(R.string.do_not_use_proxy) + " " + getString(R.string.generic_example) + ": " + getString(R.string.http_proxy_address_example));
                 return true;
-            } else if (newValue.matches("^http://(\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}):(\\d{1,5})$")) {
+            } else if (newValue.matches("^http://" +
```
AudriusButkevicius: Also, I think https proxy should use a https scheme, so you might need a separate field as I suspect this might not work.

Catfriend1: According to https://wiki.ubuntuusers.de/Proxyserver/ https proxy url is the same as http proxy url:
[quote]
export http_proxy=http://SERVER-IP_ODER_NAME:PORT
export https_proxy=http://SERVER-IP_ODER_NAME:PORT

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
Catfriend1: yeah that was my first guess... mandatory or optional for merge ;)?

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
AudriusButkevicius: It seems that there are varying opinions about this on the internet, but I guess it's ok.

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -512,5 +549,49 @@ protected void onPostExecute(Boolean haveRoot) {
...
+         */
+        private boolean handleSocksProxyPreferenceChange(Preference preference, String newValue) {
+            // Valid input is either a proxy adress or an empty field to disable the proxy.
+            if (newValue.equals("")) {
+                preference.setSummary(getString(R.string.do_not_use_proxy) + " " + getString(R.string.generic_example) + ": " + getString(R.string.socks_proxy_address_example));
+                return true;
+            } else if (newValue.matches("^socks5://" +
+                                        "(((([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))|" +
+                                        "((([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])))" +
+                                        ":(\\d{1,5})$")) {
```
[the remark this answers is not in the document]
Catfriend1: Ok, just for reference in case we'll need it in the future - this regex is fucking scaring me but it can do ipv4, ipv6 and hostname check in one piece. I am convinced now using "^socks5//.*:\\d{1,5}$"
´(^\s*((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?))\s*$)|(^\s*((?=.{1,255}$)[0-9A-Za-z](?:(?:[0-9A-Za-z]|\b-){0,61}[0-9A-Za-z])?(?:\.[0-9A-Za-z](?:(?:[0-9A-Za-z]|\b-){0,61}[0-9A-Za-z])?)*\.?)\s*$)|(^\s*((([0-9A-Fa-f]{1,4}:){7}([0-9A-Fa-f]{1,4}|:))|(([0-9A-Fa-f]{1,4}:){6}(:[0-9A-Fa-f]{1,4}|((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){5}(((:[0-9A-Fa-f]{1,4}){1,2})|:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){4}(((:[0-9A-Fa-f]{1,4}){1,3})|((:[0-9A-Fa-f]{1,4})?:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){3}(((:[0-9A-Fa-f]{1,4}){1,4})|((:[0-9A-Fa-f]{1,4}){0,2}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){2}(((:[0-9A-Fa-f]{1,4}){1,5})|((:[0-9A-Fa-f]{1,4}){0,3}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){1}(((:[0-9A-Fa-f]{1,4}){1,6})|((:[0-9A-Fa-f]{1,4}){0,4}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(:(((:[0-9A-Fa-f]{1,4}){1,7})|((:[0-9A-Fa-f]{1,4}){0,5}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:)))(%.+)?\s*$)´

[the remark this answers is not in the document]
AudriusButkevicius: That's exactly the reason why we don't need something this complicated.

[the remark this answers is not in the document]
Catfriend1: @Nutomic : For the hint, I've already socks_proxy_address_example and http_proxy_address_example strings in place that show in the EditText field's summary. Do we need more?

app/src/main/res/values/strings.xml
```diff
@@ -337,13 +340,29 @@ Please report any problems you encounter via Github.</string>
...
+    <string name="do_not_use_proxy">Do not connect through a proxy</string>
+
     <string name="keep_wakelock_while_binary_running">Keep the CPU awake while Syncthing is running</string>
 
     <string name="keep_wakelock_while_binary_running_summary">Use this setting if you experience unexpected disconnects while operating on battery. This will result in increased battery consumption.</string>
 
     <string name="use_tor_title">Use Tor</string>
 
-    <string name="use_tor_summary">Force all traffic through Tor for increased privacy. Requires Orbot</string>
+    <string name="use_tor_summary">Force all traffic through Tor for increased privacy. Requires Orbot. Disables proxy options.</string>
```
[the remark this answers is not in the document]
Catfriend1: I tried this already but have to disappoint you. It's not supported on android xml to reference a concatenation of two strings. That applies both for strings.xml and app-settings.xml. So the only way implementing this would be to write concatenation code which I think is the worse alternative.

Catfriend1:
@AudriusButkevicius The unneccessary restart got fixed by checking the new value vs. the old value and only responding to changes.

AudriusButkevicius:
There is a ton of lint errors, but I think none of them seem like yours, they must come from some earlier code.

Catfriend1:
@AudriusButkevicius I'll look into those lint errors and then make a new PR.

Issue #856, opened 2017-03-12
Pausing individual folders or devices


jebeld17:
While the official version of Syncthing on PC, Mac OS, and Linux get the advanced feature to pause synchronization of individual folders or devices, the Android variant has been left in the dust and currently has no way to pause individual things unless the user knows how to get to the Web UI and familiar with it in the first place.
This isn't fair to the modern day Android user, and this needs to be addressed.

I would love to see the Syncthing Android app natively support pausing of individual computers and folders just like the rest of the variants for all other OSes.

wweich:
You are right, that this should be implemented, especially as it is needed for #511, but:

> unless the user knows how to get to the Web UI

The Web UI has an entry inside the main menu, it can't get any easier.

> and familiar with it in the first place.

The Web UI is the UI used on any other platform, why shouldn't a user know how to use it?

jebeld17:
If there is such an "easy" Web UI, why isn't the Android app UI based off a lightweight browser - such as Chrome - instead?

Catfriend1:
I made a PR to improve the UI.

PR #1072, opened 2018-04-28
Add pause folder and device UI

ghost:
Purpose
This PR should implement the suggested enhancement from https://github.com/syncthing/syncthing-android/issues/856 Syncthing Core supports pausing and resuming folders and devices from the Web UI since several versions. Android app is currently missing implementation of paused devices and folders in the folder and device model code and a user-friendly UI is missing.

Implementation
Users can click on folders and see their properties. There's a switch to pause/resume folder synchronization.
Users can click on devices and see their properties. There's a switch to pause/resume device synchronization.
UI previously showed paused folders and devices as unknown/disconnected. This is fixed now clearly stating "Paused".

Testing
Verified working on Android 7.1.2 on lg-h815 phone.


Nutomic:
Wow you're really on a roll now! The code looks fine to me, I think you have it figured out by now.

When this is working, you could think about changing the behaviour for disabling Syncthing, so instead of stopping and starting, you just pause Syncthing instead. That might be good if the disabled state changes often, because Syncthing won't do the file rescan on startup all the time. And it might be simpler in the code. But of course it could also increase CPU usage, and cause a little data usage from the discovery.

app/src/main/java/com/nutomic/syncthingandroid/activities/DeviceActivity.java
```diff
@@ -149,6 +151,17 @@ public void onCheckedChanged(CompoundButton buttonView, boolean isChecked) {
...
     };
 
+    private final CompoundButton.OnCheckedChangeListener mDevicePausedCheckedChangeListener = new CompoundButton.OnCheckedChangeListener() {
+        @Override
+        public void onCheckedChanged(CompoundButton buttonView, boolean isChecked) {
+            if (mDevice.paused != isChecked) {
+                mDeviceNeedsToUpdate = true;
+
+                mDevice.paused = isChecked;
+            }
```
AudriusButkevicius: Noop blank line.
Also can't this just be:
```
public void onCheckedChanged(CompoundButton buttonView, boolean isChecked) {
  mDeviceNeedsToUpdate = true;
  mDevice.paused = isChecked;
}
```

app/src/main/java/com/nutomic/syncthingandroid/activities/DeviceActivity.java
```diff
@@ -167,6 +180,7 @@ public void onCreate(Bundle savedInstanceState) {
         mCompressionContainer = findViewById(R.id.compressionContainer);
         mCompressionValueView = findViewById(R.id.compressionValue);
         mIntroducerView = findViewById(R.id.introducer);
+        mDevicePaused = findViewById(R.id.devicePauseSync);
```
AudriusButkevicius: Just devicePause, there is only one type of pause, so adding the word sync just makes it confusing.

app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
```diff
@@ -114,6 +115,10 @@ public void onCheckedChanged(CompoundButton view, boolean isChecked) {
                     mFolder.fsWatcherEnabled = isChecked;
                     mFolderNeedsToUpdate = true;
                     break;
+                case R.id.folderPauseSync:
```
AudriusButkevicius: same thing, the sync is pointless.

app/src/main/java/com/nutomic/syncthingandroid/model/Folder.java
```diff
@@ -13,6 +13,7 @@
 
     public String id;
     public String label;
+    public String filesystemType = "basic";
```
AudriusButkevicius: Why do you need this?
Does not having this (and other attributes you are adding below) this break something?

app/src/main/java/com/nutomic/syncthingandroid/views/DevicesAdapter.java
```diff
@@ -45,25 +45,39 @@ public View getView(int position, View convertView, @NonNull ViewGroup parent) {
...
+            if (mConnections.connections.containsKey(deviceId)) {
+                conn = mConnections.connections.get(deviceId);
             }
-            download.setText(Util.readableTransferRate(getContext(), conn.inBits));
-            upload.setText(Util.readableTransferRate(getContext(), conn.outBits));
         }
-        else {
+
+        if (conn != null)
+        {
```
AudriusButkevicius: This is not C#, brace goes on the same line.

app/src/main/java/com/nutomic/syncthingandroid/views/DevicesAdapter.java
```diff
@@ -45,25 +45,39 @@ public View getView(int position, View convertView, @NonNull ViewGroup parent) {
...
+
+            if (conn.paused) {
+                status.setText(r.getString(R.string.device_paused));
+                status.setTextColor(ContextCompat.getColor(getContext(), R.color.text_black));
+            } else if (conn.connected) {
+                if (conn.completion == 100) {
+                    status.setText(r.getString(R.string.device_up_to_date));
+                    status.setTextColor(ContextCompat.getColor(getContext(), R.color.text_green));
+                }
+                else {
```
AudriusButkevicius: `} else {`

app/src/main/java/com/nutomic/syncthingandroid/views/DevicesAdapter.java
```diff
@@ -45,25 +45,39 @@ public View getView(int position, View convertView, @NonNull ViewGroup parent) {
...
             }
-            download.setText(Util.readableTransferRate(getContext(), conn.inBits));
-            upload.setText(Util.readableTransferRate(getContext(), conn.outBits));
         }
-        else {
+
+        if (conn != null)
+        {
+            download.setText(Util.readableTransferRate(getContext(), (conn.paused || !conn.connected) ? 0 : conn.inBits));
+            upload.setText(Util.readableTransferRate(getContext(), (conn.paused || !conn.connected) ? 0 : conn.outBits));
```
AudriusButkevicius: Just move these into the relevant branches, the double inline if over two lines is too much.
Also, I assume inBits and outBits might be 0 if paused or disconnected?

app/src/main/java/com/nutomic/syncthingandroid/views/DevicesAdapter.java
```diff
@@ -45,25 +45,39 @@ public View getView(int position, View convertView, @NonNull ViewGroup parent) {
...
-                status.setText(r.getString(R.string.device_up_to_date));
-                status.setTextColor(ContextCompat.getColor(getContext(), R.color.text_green));
-            }
-            else {
-                status.setText(r.getString(R.string.device_syncing, conn.completion));
-                status.setTextColor(ContextCompat.getColor(getContext(), R.color.text_blue));
+
+        Connections.Connection conn = null;
+        if (mConnections != null) {
+            if (mConnections.connections.containsKey(deviceId)) {
```
AudriusButkevicius: You can move this into the if above, as in
```
if (mConnections != null && mConnections.connections.containsKey(deviceId)) {
```

app/src/main/java/com/nutomic/syncthingandroid/views/DevicesAdapter.java
AudriusButkevicius: Also you could do
```
if (conn == null) {
   set some things
   return early
}
```

so there is less mental exercise following the ifs.

app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
```diff
@@ -118,17 +122,15 @@ public View getView(int position, View convertView, @NonNull ViewGroup parent) {
...
-            case "syncing":  return c.getString(R.string.state_syncing, percentage);
-            case "error":    return c.getString(R.string.state_error);
-            case "unknown":  // Fallthrough
-            case "":         return c.getString(R.string.state_unknown);
-        }
-        if (BuildConfig.DEBUG) {
-            throw new AssertionError("Unexpected folder state " + state);
+            case "idle":        return c.getString(R.string.state_idle);
+            case "scanning":    return c.getString(R.string.state_scanning);
+            case "cleaning":    return c.getString(R.string.state_cleaning);
```
AudriusButkevicius: this state does not exist...

app/src/main/res/values/strings.xml
```diff
@@ -146,6 +152,9 @@ Please report any problems you encounter via Github.</string>
     <!-- Setting title -->
     <string name="folder_fileWatcher">Realtime Sync</string>
 
+    <!-- Setting title -->
+    <string name="folder_pause_sync">Pause Sync</string>
```
AudriusButkevicius: Pause folder

app/src/main/res/values/strings.xml
```diff
@@ -209,6 +218,9 @@ Please report any problems you encounter via Github.</string>
     <!-- Setting title -->
     <string name="introducer">Introducer</string>
 
+    <!-- Setting title -->
+    <string name="device_pause_sync">Pause Sync</string>
```
AudriusButkevicius: Pause device

app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
```diff
@@ -118,17 +122,15 @@ public View getView(int position, View convertView, @NonNull ViewGroup parent) {
...
+            case "syncing":     return c.getString(R.string.state_syncing, percentage);
+            case "error":       return c.getString(R.string.state_error);
+            case "unknown":     return c.getString(R.string.state_unknown);
+            case "":            return c.getString(R.string.state_paused);
```
AudriusButkevicius: You should not do it based on an empty string. You should check that the Folder object has the paused attribute set to true.

app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
```diff
@@ -92,6 +92,10 @@ public View getView(int position, View convertView, @NonNull ViewGroup parent) {
                     case "syncing":
                         binding.state.setTextColor(ContextCompat.getColor(getContext(), R.color.text_blue));
                         break;
+                    case "":
```
AudriusButkevicius: You should not do it based on an empty string. You should check that the Folder object has the paused attribute set to true.

Perhaps model.paused?

app/src/main/java/com/nutomic/syncthingandroid/views/DevicesAdapter.java
```diff
@@ -45,25 +45,39 @@ public View getView(int position, View convertView, @NonNull ViewGroup parent) {
...
+                status.setText(r.getString(R.string.device_disconnected));
+                status.setTextColor(ContextCompat.getColor(getContext(), R.color.text_red));
+            }
+        } else {
             download.setText(Util.readableTransferRate(getContext(), 0));
             upload.setText(Util.readableTransferRate(getContext(), 0));
-            status.setText(r.getString(R.string.device_disconnected));
+            status.setText(r.getString(R.string.device_state_unknown));
             status.setTextColor(ContextCompat.getColor(getContext(), R.color.text_red));
         }
```
AudriusButkevicius: This whole block could be rewritten as:
```
if (conn == null) {
   do stuff
   return early
}

if (conn.paused) {
  do stuff
  return early
}

if (conn.connected) {
  do stuff
  return early
}

set disconnected
return
```

This way there is less nesting and easier to follow.

app/src/main/java/com/nutomic/syncthingandroid/activities/DeviceActivity.java
Catfriend1: I simplified the code by merging the button change part in mCheckedListener (as its already the case in FolderActivity.java)

app/src/main/java/com/nutomic/syncthingandroid/activities/DeviceActivity.java
Catfriend1: Ok.

app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
Catfriend1: Done.

app/src/main/java/com/nutomic/syncthingandroid/model/Folder.java
Catfriend1: I made a comparison with syncthing's v0.14.46 config.xml to the Folder model and found keys missing in there. I thought it is good practice to have them set to a default. This can't break anything because the model is normally retrieved by rest get "/rest/system/config" before any change is done to the attributes.  Reason why I put this is I stumbled over the problem with the fsWatcherEnabled in my previous PR where this wasn't saved correctly by syncthing core on the POST request because fsWatcherDelayS was missing in the model. Imagine a user updating syncthing app from a very old version before the fsWatcher (or filesystemType attribute). It could cause the config to break if the app assumes the REST API sent information on it but it didn't on the upgrade in place. Then a write-back the app's post could mess it up.

AudriusButkevicius: Sure.

app/src/main/java/com/nutomic/syncthingandroid/views/DevicesAdapter.java
Catfriend1: Agreed.

app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
AudriusButkevicius: No. There is a paused attribute on the object returned by the config, that is what should be used.

Catfriend1:
@AudriusButkevicius Ok, thanks again for your feedback. I used folder.paused to show the correct UI text.

app/src/main/res/values/strings.xml
```diff
@@ -209,6 +218,9 @@ Please report any problems you encounter via Github.</string>
     <!-- Setting title -->
     <string name="introducer">Introducer</string>
 
+    <!-- Setting title -->
+    <string name="device_pause">Pause Device</string>
```
AudriusButkevicius: Can you call this pause_device, as the string below is delete_device, so this matches up.

app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
```diff
@@ -118,17 +123,15 @@ public View getView(int position, View convertView, @NonNull ViewGroup parent) {
...
-        }
-        if (BuildConfig.DEBUG) {
-            throw new AssertionError("Unexpected folder state " + state);
+            case "idle":        return c.getString(R.string.state_idle);
+            case "scanning":    return c.getString(R.string.state_scanning);
+            case "cleaning":    return c.getString(R.string.state_cleaning);
+            case "syncing":     return c.getString(R.string.state_syncing, percentage);
+            case "error":       return c.getString(R.string.state_error);
+            case "unknown":     // Fallthrough
+            case "":            return c.getString(R.string.state_unknown);
```
AudriusButkevicius: When is this an empty string?

app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
AudriusButkevicius: Just remove it.

app/src/main/java/com/nutomic/syncthingandroid/activities/DeviceActivity.java
```diff
@@ -293,18 +299,26 @@ private void onApiChange(SyncthingService.State currentState) {
     }
 
     private void updateViewsAndSetListeners() {
+        mIdView.removeTextChangedListener(mIdTextWatcher);
+        mNameView.removeTextChangedListener(mNameTextWatcher);
+        mAddressesView.removeTextChangedListener(mAddressesTextWatcher);
+        mIntroducerView.setOnCheckedChangeListener(null);
+        mDevicePaused.setOnCheckedChangeListener(null);
```
AudriusButkevicius: Why are we touching mIdView and others here?

app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
```diff
@@ -82,18 +82,23 @@ public View getView(int position, View convertView, @NonNull ViewGroup parent) {
...
+                    binding.state.setText(getContext().getString(R.string.state_paused));
+                    binding.state.setTextColor(ContextCompat.getColor(getContext(), R.color.text_black));
+                } else {
+                    binding.state.setText(getLocalizedState(getContext(), model.state, percentage));
+                    switch(model.state) {
+                        case "idle":
+                            binding.state.setTextColor(ContextCompat.getColor(getContext(), R.color.text_green));
+                            break;
+                        case "scanning":
+                        case "cleaning":
```
AudriusButkevicius: This state does not exist, just remove it.

app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
AudriusButkevicius: Perhaps just replace it with `default: return state`, and return unknown explicitly?

Catfriend1: According to my debugging, the string model.state is empty when a folder is in paused state. The REST API does not return "paused" here.

app/src/main/res/values/strings.xml
Catfriend1: ok

app/src/main/java/com/nutomic/syncthingandroid/activities/DeviceActivity.java
Catfriend1: Removing those listeners temporarily prevents the "user changed property" function to trigger if we fill or update the UI. Thats why the function is called updateViewsAndSetListeners. Look a litte bit down, they are readded after the fields have been filled.

app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
Catfriend1: I think you meant the "cleaning" state above as mentioned here? I removed it. Never saw it in syncthing before.

Catfriend1:
@AudriusButkevicius ok, I implemented your suggestions :)

app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
AudriusButkevicius: Sure, but that's just a coincidence, I am pretty sure that's not intended, so deliberately checking for that seems wrong.

app/src/main/java/com/nutomic/syncthingandroid/activities/DeviceActivity.java
AudriusButkevicius: Right, but what in this PR prompted the need for this change?

app/src/main/res/values-bg/strings.xml
```diff
@@ -291,7 +291,7 @@
   <!--Possible folder states-->
   <string name="state_idle">Бездействие</string>
   <string name="state_scanning">Сканиране</string>
-  <string name="state_cleaning">Почистване</string>
+
```
AudriusButkevicius: Blank new lines everywhere.

app/src/main/java/com/nutomic/syncthingandroid/activities/DeviceActivity.java
Catfriend1: The new 'consolidated' checkedlistener does not compare previous and new values to make the code simpler. This does not work if the no change event will trigger from the above ui init function. I just did it the same way as it was aleady in FolderActivity before this pr.

app/src/main/res/values-bg/strings.xml
Catfriend1: They'll get deleted (overwritten) as tx pull will take place fetching up new translations from transifex. Btw can you please push out new strings to transifex with tx push after merge, please?

AudriusButkevicius: I don't want to have a PR that adds a bunch of meaningless lines in the history, so let's remove these in this PR. I'll try to update the translations post merge.

Catfriend1: Ok, makes sense keeping the history clean.  Build may fail because of the "beetling" translation strings, but we are fine.

Catfriend1:
@AudriusButkevicius I have a commit prepared locally where the translated strings were removed properly according to our coding guidelines. In case we need it, just tell me and I push it here to fix the build.

AudriusButkevicius:
Yes push it as I can't merge while the build is broken.

Catfriend1:
Ok, done (https://github.com/syncthing/syncthing-android/pull/1072/commits/f3882e6000d25fb1b6910403ded468cbe3af5685)

Issue #1059, opened 2018-04-24
ConfigXml.java:changeLocalDeviceName() changes all device names after app restore


Catfriend1:
The loop in function is intended to change the local device name on first app start to build.MODEL. When manually recovering syncthings app data (TitaniumBackup, Google Backup, adb backup or Root Explorer) the app thinks it starts for the first time but the config is already there and valid. It does not generate new keys but changes all device names accidentially to build.MODEL. 

`for (int i = 0; i < childNodes.getLength(); i++) {
            Node node = childNodes.item(i);
            if (node.getNodeName().equals("device")) {
                ((Element) node).setAttribute("name", Build.MODEL);
            }
        }`

To fix this we can say the first device in the config always proves to be ourself and we can stop after adjusting the first occurence device name. I've verified the fix working in PR https://github.com/syncthing/syncthing-android/pull/1051 and added a todo comment about this. In a subsequent new PR, I'd like to dig deeper into this issue and also detect to correct device xml node to change by reading the "Self device ID" first and directly looking for that in the xml.

    App Version: 0.10.8 (and before)
    Syncthing Version: all recent and previois versions
    Android Version: Android7.1.2, LineAgeOS 14.1 @ lg-h815
    Root: yes/no does not matter

AudriusButkevicius:
I don't think the fix is right, probably a better option would be to derive device ID from the certificates, and rename that.

AudriusButkevicius:
So I am fine with doing in a separate PR, but I'd rather not add any half-assed fix now.

Catfriend1:
Ok, I reverted back to the original code so we can fix it later properly.

Catfriend1:
I am sure this user had the same problem as I described here, link https://forum.syncthing.net/t/all-settings-are-gone-after-updating-app/11514/6. I'll reopen that as a reminder if have still to find a proper fix for "changeLocalDeviceName" which should not go into nested elements inside the folder element as the comment above the function states.

PR #1073, opened 2018-04-30
fix config corruption on app upgrade or restore

Catfriend1:
Purpose
This PR should address the bug described in issue https://github.com/syncthing/syncthing-android/issues/1059 . ConfigXml/changeLocalDeviceName overrides all device names if the app has been upgraded or restored. Instead, it should only change the local device's name to build.model (phone model) as originally intended.

Implementation
I implemented a new function that reads the syncthing native binary log on first app launch after the keys got generated. Then, it extracts the local device ID and hands it over to changeLocalDeviceName which then ensures to only change the device name for the node having the correct device ID. If somehow the device ID cannot be retrieved, it won't do the first-app-launch rename of the local device for safety.

Testing
Verified working on my lg-h815 phone running Android 7.1.2.
image


Catfriend1:
damn wrong base

PR #1074, opened 2018-04-30
 Correct device renaming, save config via temp file

ghost:
Purpose
This PR should address the bug described in issue https://github.com/syncthing/syncthing-android/issues/1059 . ConfigXml/changeLocalDeviceName overrides all device names if the app has been upgraded or restored. Instead, it should only change the local device's name to build.model (phone model) as originally intended.
Also fixes this problem. https://forum.syncthing.net/t/all-settings-are-gone-after-updating-app/11514/6

Implementation
I implemented a new function that reads the syncthing native binary log on first app launch after the keys got generated. Then, it extracts the local device ID and hands it over to changeLocalDeviceName which then ensures to only change the device name for the node having the correct device ID. If somehow the device ID cannot be retrieved, it won't do the first-app-launch rename of the local device for safety.

Testing
Verified working on my lg-h815 phone running Android 7.1.2.
image

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
@@ -133,13 +133,13 @@ public class SyncthingRunnable implements Runnable {
             Thread lWarn = null;
             if (returnStdOut) {
                 try {
-                    InputStreamReader isr = new InputStreamReader(process.getInputStream(), Charsets.UTF_8);
-                    BufferedReader br = new BufferedReader(isr);
+                    BufferedReader br = new BufferedReader(new InputStreamReader(process.getInputStream(), Charsets.UTF_8));
                     String line;
                     while ((line = br.readLine()) != null) {
                         Log.println(Log.INFO, TAG_NATIVE, line);
                         capturedStdOut = capturedStdOut + line + "\n";
                     }
+                    br.close();
                 } catch (IOException e) {
                     Log.w(TAG, "Failed to read Syncthing's command line output", e);
                 }
@@ -346,6 +346,7 @@ public class SyncthingRunnable implements Runnable {
                         Files.append(line + "\n", mLogFile, Charsets.UTF_8);
                     }
                 }
+                br.close();
             } catch (IOException e) {
                 Log.w(TAG, "Failed to read Syncthing's command line output", e);
             }
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
@@ -132,16 +132,19 @@ public class SyncthingRunnable implements Runnable {
             Thread lInfo = null;
             Thread lWarn = null;
             if (returnStdOut) {
+                BufferedReader br = null;
                 try {
-                    BufferedReader br = new BufferedReader(new InputStreamReader(process.getInputStream(), Charsets.UTF_8));
+                    br = new BufferedReader(new InputStreamReader(process.getInputStream(), Charsets.UTF_8));
                     String line;
                     while ((line = br.readLine()) != null) {
                         Log.println(Log.INFO, TAG_NATIVE, line);
                         capturedStdOut = capturedStdOut + line + "\n";
                     }
-                    br.close();
                 } catch (IOException e) {
                     Log.w(TAG, "Failed to read Syncthing's command line output", e);
+                } finally {
+                    if (br != null)
+                        br.close();
                 }
             } else {
                 lInfo = log(process.getInputStream(), Log.INFO, true);
@@ -335,9 +338,9 @@ public class SyncthingRunnable implements Runnable {
      */
     private Thread log(final InputStream is, final int priority, final boolean saveLog) {
         Thread t = new Thread(() -> {
+            BufferedReader br = null;
             try {
-                InputStreamReader isr = new InputStreamReader(is, Charsets.UTF_8);
-                BufferedReader br = new BufferedReader(isr);
+                br = new BufferedReader(new InputStreamReader(is, Charsets.UTF_8));
                 String line;
                 while ((line = br.readLine()) != null) {
                     Log.println(priority, TAG_NATIVE, line);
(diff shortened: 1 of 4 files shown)


AudriusButkevicius:
No, this is the wrong approach, you should generate the ID from the certifocate file like syncthing does. If we change the log format or localize logs, this house of cards falls apart.

AudriusButkevicius:
Also, given syncthing is up at this point, you can use the API to work out the device IDby checking /system or headers.

AudriusButkevicius:
Also, when we write the config back we should be careful, write to a tmp file and then rename to make the write atomic.

Catfriend1:
Ok, so this naming niceness in first startup will need a lot of more work to fix to be even compatible with future releases. First I thought it was as easy as @calmh pointed it out in  https://github.com/syncthing/syncthing/issues/4217Dd where he suggested to parse the log. I'm ging to look a those X509Certificate classes and Android crypto classes when I have more time and hopefully find an acceptable solution.

calmh:
If it's just the device ID you're after, there is a supported way from the command line:

```
$ syncthing --device-id
SYNO4VL-TKVFF25-QKC2T3V-TTK4BGU-D2ZA5OT-PDKSI3R-3P4JQVS-BL4VXAC
```

Catfriend1:
@AudriusButkevicius: Would you agree if I try implementation via the command line first and later improve it by directly loading the id from the certificates?

AudriusButkevicius:
I guess. Also, why can't you update the name after startup when the API is available and all the info you need is there.

Catfriend1:
@AudriusButkevicius The problem is after the command "generatekeys" that is issued via SyncthingRunnable syncthing terminates itself. So we would need a new SyncthingRunnable.Run() function with all that asynchronous multi-thread stuff for polling the WebGUI after its available and starting syncthing in "production" mode after generating the keys. On my opinion this makes it too complicated and new bugs could rise easily for just gaining a proper a proper device rename. That's why I would stick to the same way "generatekeys" is already implemented and extend it to capture the command line output of "device-id" issued afterwards. I'm committing the code that worked on my phone shortly.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
```diff
@@ -120,16 +129,34 @@ public void run() {
...
             mSyncthing.set(process);
 
-            Thread lInfo = log(process.getInputStream(), Log.INFO, true);
-            Thread lWarn = log(process.getErrorStream(), Log.WARN, true);
+            Thread lInfo = null;
+            Thread lWarn = null;
+            if (returnStdOut) {
+                try {
+                    InputStreamReader isr = new InputStreamReader(process.getInputStream(), Charsets.UTF_8);
+                    BufferedReader br = new BufferedReader(isr);
```
AudriusButkevicius: Don't you need to close these?

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -313,15 +322,25 @@ private void saveChanges() {
...
             transformer.transform(domSource, streamResult);
         } catch (TransformerException e) {
-            Log.w(TAG, "Failed to save updated config", e);
+            Log.w(TAG, "Failed to save temporary config file", e);
+            return;
+        }
+        try {
+            // Delete outdated config file and put temporary config file instead.
+            mConfigFile.delete();
+            mConfigTempFile.renameTo(mConfigFile);
```
AudriusButkevicius: RenameTo should be enough, no need to delete

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
Catfriend1: Yes. I also added the bufferedReader.close() to the existing code below I didn't touch.
The close also closes the InputStreamReader associated with it (https://docs.oracle.com/javase/8/docs/api/java/io/BufferedReader.html#close--)

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: hmm.. I'll better test this before as JavaDoc states it can be wrong. Link https://docs.oracle.com/javase/6/docs/api/java/io/File.html#renameTo%28java.io.File%29

Catfriend1: Verified with a debug build and its working correctly without the delete before renameTo. Good point!

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
```diff
@@ -133,13 +133,13 @@ public String run(boolean returnStdOut) {
...
                 try {
-                    InputStreamReader isr = new InputStreamReader(process.getInputStream(), Charsets.UTF_8);
-                    BufferedReader br = new BufferedReader(isr);
+                    BufferedReader br = new BufferedReader(new InputStreamReader(process.getInputStream(), Charsets.UTF_8));
                     String line;
                     while ((line = br.readLine()) != null) {
                         Log.println(Log.INFO, TAG_NATIVE, line);
                         capturedStdOut = capturedStdOut + line + "\n";
                     }
+                    br.close();
```
AudriusButkevicius: Probably want to do this in a finally, in case something throws.

Catfriend1: Yeah, got a logcat and indeed, after the "generatekeys" execution, the original code before this PR throws a lot of bufferedReader exceptions before the thread is terminated.

Catfriend1: Huh I've got a strange compiler warning when I want to add the br.close into the finally-block of function thread "log" (line 354). It complains an IOException has to be caught, so please don't wonder why there a two tries now to clean this up.

Catfriend1: Verified the lastest commit https://github.com/syncthing/syncthing-android/pull/1074/commits/f436d4ee58c40124f80a763449a0dcca730d70e6 working correctly. And hey, you're idea kicke ass - the warning lines because of bufferedReader no longer occur.

AudriusButkevicius (approved):

Issue #1062, opened 2018-04-26
Incorrect native UI status of a remote device during synchronization


scienmind:
Following up on the [forum discussion](https://forum.syncthing.net/t/incorrect-devices-status-on-android/11510):

   1. Add a big file to a folder shared with a remote device
   2. Open Syncthing app (native UI), and review the remote device's status
   * Result: remote device is shown as "Up to Date"
   * Expected result: remote device is shown as "Syncing"

WebUI shows "Syncing" status correctly

### Version Information

    App Version: 0.10.8
    Syncthing Version: v0.14.46
    Android Version: Android 8.0

Catfriend1:
Working on this in https://github.com/syncthing/syncthing-android/compare/master...Catfriend1:deviceSyncStatus

Catfriend1:
Hi @scienmind ,
can you please test if this debug build fixes the problem with the device UI sync status?
https://build.syncthing.net/repository/download/SyncthingAndroid_Build/20341:id/debug/app-debug.apk
Please be sure to backup your data and config first before switching to the debug build. It contains a bunch of more changes that are currently unreleased like syncthing 0.14.47

scienmind:
Thanks, @Catfriend1 !

Just tested the linked apk on one of the devices, looks fixed indeed 👍

Catfriend1:
@scienmind Thanks for testing.

Catfriend1:
@scienmind The last apk linked above had a serious error in its code as I discovered today and fixed it (endless loop problem if a device got deleted). Would you mind trying this one? https://build.syncthing.net/repository/download/SyncthingAndroid_Build/20364:id/debug/app-debug.apk

scienmind:
I tried the newer apk, shows "Syncing" status when appropriate as well.

FYI - don't consider it "well tested" by me in any way. Just a sanity check if it basically does what it should.

PR #1079, opened 2018-05-06
Fix incorrect remote device syncing status UI

ghost:
Purpose
This PR fixes the problem described in issue https://github.com/syncthing/syncthing-android/issues/1062 where remote devices are constantly shown as "Up-to-date" and idle in the app's device UI. With this fix, the app listens to the events REST API and caches remote device and folder completion information. This information is display on a per-device basis in the device UI. The user can now see, "how much percent a remote device" is already synced, e.g. when the remote downloads data from ourself.

I've also applied some logical renames to class and variable names because the code called the "FolderStatus" "model" before which was confusing. The classes are now clearly called out "FolderStatus" (existed before, unchanged) and "Completion".

Testing
Verified working on my phone lg h-815 under Android 7.1.2 at commits:
- https://github.com/syncthing/syncthing-android/pull/1079/commits/1d6d8df1b92f33ddcda64faea37915049f11c062


app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
```diff
@@ -102,12 +102,12 @@ public View getView(int position, View convertView, @NonNull ViewGroup parent) {
...
-                    .getQuantityString(R.plurals.files, (int) model.inSyncFiles, model.inSyncFiles, model.globalFiles));
+                    .getQuantityString(R.plurals.files, (int) FolderStatusInfo.inSyncFiles, FolderStatusInfo.inSyncFiles, FolderStatusInfo.globalFiles));
             binding.size.setVisibility(VISIBLE);
             binding.size.setText(getContext().getString(R.string.folder_size_format,
-                    Util.readableFileSize(getContext(), model.inSyncBytes),
-                    Util.readableFileSize(getContext(), model.globalBytes)));
-            setTextOrHide(binding.invalid, model.invalid);
+                    Util.readableFileSize(getContext(), FolderStatusInfo.inSyncBytes),
+                    Util.readableFileSize(getContext(), FolderStatusInfo.globalBytes)));
+            setTextOrHide(binding.invalid, FolderStatusInfo.invalid);
```
Nutomic: Just realized that this method is really long. It would be a good idea to split it up somehow (but not necessarily in this PR).

app/src/main/java/com/nutomic/syncthingandroid/http/GetRequest.java
```diff
@@ -16,14 +16,14 @@
...
-    public static final String URI_REPORT      = "/rest/svc/report";
-    public static final String URI_EVENTS      = "/rest/events";
+    public static final String URI_CONFIG       = "/rest/system/config";
+    public static final String URI_VERSION      = "/rest/system/version";
+    public static final String URI_SYSTEM       = "/rest/system/status";
+    public static final String URI_CONNECTIONS  = "/rest/system/connections";
+    public static final String URI_STATUS       = "/rest/db/status";
+    public static final String URI_DEVICEID     = "/rest/svc/deviceid";
+    public static final String URI_REPORT       = "/rest/svc/report";
+    public static final String URI_EVENTS       = "/rest/events";
```
Nutomic: Why the whitespace change?

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,44 @@
...
+        addFolder(deviceId, folderId);
+        return (deviceFolderMap.get(deviceId)).get(folderId);
+    }
+
+    public void setCompletionInfo(String deviceId, String folderId,
+                                    CompletionInfo completionInfo) {
+        addFolder(deviceId, folderId);
+        (deviceFolderMap.get(deviceId)).put(folderId, completionInfo);
+    }
+}
```
Nutomic: You should add some comments to this class and methods.

app/src/main/java/com/nutomic/syncthingandroid/http/GetRequest.java
Catfriend1: I did some tabbing due to URI_CONNECTIONS. Revert?

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: Ok, done.

app/src/main/java/com/nutomic/syncthingandroid/http/GetRequest.java
Nutomic: I just wondered why you changed the other lines too, because there is no need to indent it any more. Now it looks fine.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Nutomic: You should also use /** ... */ for method documentation, as it is used by various tools for documenation (including Android Studio). Those tools ignore comments with //.

Catfriend1: Thanks for the hint. Corrected it.

Nutomic (approved):

Catfriend1:
lgtm

Catfriend1:
scienmind tested this and confirmed it working, see linked issue above.

AudriusButkevicius:
I need to go through this in more detail. Perhaps this evening.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * This class caches remote folder and device synchronization
+ * completion indicators defined in {@link CompletionInfo#CompletionInfo}
+ * according to syncthing's REST "/completion" JSON result schema.
+ * Completion model of syncthing's web UI is completion[deviceId][folderId]
+ */
+public class Completion {
+
+    HashMap<String, HashMap<String, CompletionInfo>> deviceFolderMap =
+        new HashMap<String, HashMap<String, CompletionInfo>>();
```
AudriusButkevicius: Perhaps something like this can reduce boilerplate code https://stackoverflow.com/a/7519422

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * Adds a device to the cache model if it does not exist.
+     */
+    private void addDevice(String deviceId) {
+        if (!deviceFolderMap.containsKey(deviceId))
+            deviceFolderMap.put(deviceId, new HashMap<String, CompletionInfo>());
```
Nutomic: You should be consistent about braces, like `if (...) {}`. I'd say it's better to always have braces, as that prevents errors and makes clear what's happening.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Nutomic: Didn't notice that before, it actually looks pretty complicated to have nested maps like this. Maybe you can add an inner class that holds the second map and simplifies things.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: Ok, good idea.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: Whoa... I'm quite out of ideas for today. I get your point but currently I haven't found an easier, shorter implementation yet. I also added two update functions with commit https://github.com/syncthing/syncthing-android/pull/1079/commits/a5b2c57cc725c4435c8bb38464fc9ef30fca0eeb that are called after a config reload took place to clean out folders and devices that were cached before in the Completion class and do no longer exist after the config reload. I verified this working "fastly" on a small setup. But the code is not optimized for a high count of folder and device pairs because searching through the hashmaps has nearly an O(n^2) complexity. I thought of replacing it with Collection.html#retainAll 
( http://download.oracle.com/javase/6/docs/api/java/util/Collection.html#retainAll%28java.util.Collection%29 ) but this has the worst complexity in doing single comparations between old and new config hierarchy. I would appreciate any help on this.

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
```diff
@@ -117,9 +121,16 @@ public void onEvent(Event event) {
 
                 notify(title, pi);
                 break;
+            case "FolderCompletion":
+                deviceId = (String) event.data.get("device");
+                folderId = (String) event.data.get("folder");
+                CompletionInfo completionInfo = new CompletionInfo();
+                completionInfo.completion = (Double) event.data.get("completion");
```
AudriusButkevicius: Why can't we decode json straight into the struct, and then just updateCompletion(completion) and internally access folder and device id fields to do whatever is needed.

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
```diff
@@ -170,6 +181,17 @@ public void onEvent(Event event) {
...
+            case "StateChanged":
+            case "StartupComplete":
+            case "DeviceConnected":
+            case "FolderSummary":
+            case "RemoteIndexUpdated":
+            case "LocalIndexUpdated":
+            case "ItemStarted":
+            case "FolderPaused":
+            case "FolderScanProgress":
+            case "LoginAttempt":
```
AudriusButkevicius: I would break after this. These states are handled, we know about them but choose to do nothing about them.
You also might want to update completion on IndexUpdated events.

AudriusButkevicius: And also folder summary events for local completion

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+
+    /**
+     * Calculates remote device sync completion percentage across all folders
+     * shared with the device.
+     */
+    public int getDeviceCompletion(String deviceId) {
+        addDevice(deviceId);
+        int folderCount = 0;
```
AudriusButkevicius: Instead of calling add and what not, you could do getOrDefault and just potentoally return zeroes early.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+        return (int) Math.floor(sumCompletion / folderCount);
+    }
+
+    /**
+     * Returns completionInfo from the completion[deviceId][folderId] model.
+     */
+    public CompletionInfo getCompletionInfo(String deviceId, String folderId) {
+        addFolder(deviceId, folderId);
+        return (deviceFolderMap.get(deviceId)).get(folderId);
```
AudriusButkevicius: Same here. getOrDefault

app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
```diff
@@ -95,10 +97,14 @@
     private long mPreviousConnectionTime = 0;
 
     /**
-     * Stores the latest result of {@link #getModel} for each folder, for calculating device
-     * percentage in {@link #getConnections}.
+     * Stores the latest result of {@link #getFolderStatus} for each folder
      */
-    private final HashMap<String, Model> mCachedModelInfo = new HashMap<>();
+    private HashMap<String, FolderStatus> mCachedFolderStatusInfo = new HashMap<>();
```
AudriusButkevicius: Perhaps folderStatuses

app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
```diff
@@ -157,13 +163,19 @@ public void onWebGuiAvailable() {
...
+    private void onReloadConfigComplete(String result) {
+        Log.v(TAG, "onReloadConfigComplete: " + result);
+        mConfig = new Gson().fromJson(result, Config.class);
+        if (mConfig == null) {
+            throw new RuntimeException("config is null: " + result);
+        }
+
+        // Update cached device and folder information stored in the mCompletion model.
+        mCompletion.updateDevices(getDevices(true));
+        mCompletion.updateFolders(getFolders());
```
AudriusButkevicius: Why would config being saved affect completion?

AudriusButkevicius: Actually, this just removes stuff that doesn't exist etc, so it's ok.

app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
```diff
@@ -241,6 +253,7 @@ public void editFolder(Folder newFolder) {
 
     public void removeFolder(String id) {
         removeFolderInternal(id);
+        mCompletion.removeFolder(id);
```
AudriusButkevicius: We don't need this, the ConfigSaved would handle this already.

app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
```diff
@@ -300,6 +313,7 @@ public void editDevice(Device newDevice) {
 
     public void removeDevice(String deviceId) {
         removeDeviceInternal(deviceId);
+        mCompletion.removeDevice(deviceId);
```
AudriusButkevicius: Ditto

app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
```diff
@@ -494,6 +476,14 @@ private void normalizeDeviceId(String id, OnResultListener1<String> listener,
         });
     }
 
+
+    /**
+     * Updates cached folder and device completion info according to event data.
+     */
+    public void updateCompletionInfo(String deviceId, String folderId, CompletionInfo completionInfo) {
+        mCompletion.setCompletionInfo(deviceId, folderId, completionInfo);
+    }
```
AudriusButkevicius: This function literally does nothing.

app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
```diff
@@ -36,7 +36,7 @@
 
     private static final String TAG = "FoldersAdapter";
 
-    private final HashMap<String, Model> mModels = new HashMap<>();
+    private final HashMap<String, FolderStatus> mFolderStatus = new HashMap<>();
```
AudriusButkevicius: Statuses

app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
```diff
@@ -132,16 +136,16 @@ private static String getLocalizedState(Context c, String state, int percentage)
...
+    public void updateFolderStatus(RestApi api) {
         for (int i = 0; i < getCount(); i++) {
-            api.getModel(getItem(i).id, this::onReceiveModel);
+            api.getFolderStatus(getItem(i).id, this::onReceiveFolderStatus);
         }
     }
 
-    private void onReceiveModel(String folderId, Model model) {
-        mModels.put(folderId, model);
+    private void onReceiveFolderStatus(String folderId, FolderStatus FolderStatusInfo) {
```
AudriusButkevicius: Copy pasta variable naming

app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
```diff
@@ -390,46 +404,14 @@ public void getConnections(final OnResultListener1<Connections> listener) {
...
      */
-    public void getModel(final String folderId, final OnResultListener2<String, Model> listener) {
-        new GetRequest(mContext, mUrl, GetRequest.URI_MODEL, mApiKey,
+    public void getFolderStatus(final String folderId, final OnResultListener2<String, FolderStatus> listener) {
+        new GetRequest(mContext, mUrl, GetRequest.URI_STATUS, mApiKey,
                     ImmutableMap.of("folder", folderId), result -> {
-            Model m = new Gson().fromJson(result, Model.class);
-            mCachedModelInfo.put(folderId, m);
+            FolderStatus m = new Gson().fromJson(result, FolderStatus.class);
+            mCachedFolderStatusInfo.put(folderId, m);
```
AudriusButkevicius: Why can't you just pass m? I assume folderId is an attribute on that struct.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,146 @@
...
+        if (!deviceFolderMap.get(deviceId).containsKey(folderId)) {
+            deviceFolderMap.get(deviceId).put(folderId, new CompletionInfo());
+        }
+    }
+
+    /**
+     * Removes a folder from the cache model.
+     */
+    public void removeFolder(String folderId) {
+        for (String deviceId : deviceFolderMap.keySet()) {
```
AudriusButkevicius: I don't think you need half of these methods, of you grt or default and always update from config.

app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
```diff
@@ -157,13 +163,19 @@ public void onWebGuiAvailable() {
...
 
     public void reloadConfig() {
-        new GetRequest(mContext, mUrl, GetRequest.URI_CONFIG, mApiKey, null, result -> {
-            Log.v(TAG, "reloadConfig: " + result);
-            mConfig = new Gson().fromJson(result, Config.class);
-            if (mConfig == null) {
-                throw new RuntimeException("config is null: " + result);
-            }
-        });
+        new GetRequest(mContext, mUrl, GetRequest.URI_CONFIG, mApiKey, null, result -> onReloadConfigComplete(result));
```
AudriusButkevicius: Can't the lamba just be rewritten as this::onReloadConfigComplete

app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
AudriusButkevicius: I assume this is local stuff and that double map stuff is remote? Its not obvious from variable names

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
Catfriend1: Thats because of the structure difference describen in https://docs.syncthing.net/rest/db-completion-get.html and in https://docs.syncthing.net/events/foldercompletion.html . I created the CompletionInfo class according to the first source, so we can later improve the code and initially poll for it. To save resources, I decided to leave this part out as syncthing automatically emits this information when a device connects for the folder associated with it. If there are problems in the future that the event does not suffice, we have the full structure including globalBytes, needBytes, needDeletes available.
Btw, the event has just three fields in "event.data". An alternative would be to define an update structure separately (do we need this complexity at first integration?).

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
Catfriend1: Added to my todo list.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: I tried implementing this, but gradle complains about the API level must be raised to >= 24 to use getOrDefault.  ../../src/main/java/com/nutomic/syncthingandroid/model/Completion.java:122: Call requires API level 24 (current min is 16): java.util.HashMap#getOrDefault

Catfriend1: Trying implementation via extending HashMap to DefaultHashMap ...

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: I removed "getCompletionInfo" as it is not called externally atm. This removes redundant checks in "getDeviceCompletion" where is moved the getCompletion code part with getOrDefault.

app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
Catfriend1: ok

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: ... done.

app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
Catfriend1: removed.

app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
Catfriend1: removed.

app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
Catfriend1: It's a setter to avoid exposing the whole mCompletion instance to all mRestApi parents.

app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
Catfriend1: Renamed to mLocalFolderStatuses

app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
Catfriend1: Indeed, sorry. I'll update that. Thanks for your patience.

app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
Catfriend1: No, it isn't. See FolderStatus.java. Anyways, I'd like the idea to put that attribute in, I think that wouldn't help us simplifying things because that "old" code before the PR needs folderId as a key on the HashMap.

app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
Catfriend1: ok

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: I'll put this on my todo list. For now, I've combined the two function Completion.updateDevices and Completion.updateFolders into one, made the code less redundant checking for existing keys on the hash maps. Added detection of newly created devices on configReload event. The detection of newly created folders (with applying them to the mCompletion model) would require two also implement the two nested hashmap approach in all code parts referring to "Nutomics original" Folder.java and Device.java model classes. I'd suggest doing this in a later PR.

Catfriend1:
Verified working after commit https://github.com/syncthing/syncthing-android/commit/51a436f2dd1402cc5580b5d0c5fa40ff0a0c0ef4

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: So, for now updateFromConfig is complete adopting the existing Folder.java model containing the devices in a list under every folder. It checks for added folders since commit https://github.com/syncthing/syncthing-android/pull/1079/commits/f652b265c39fd7eaa5d3fbdbb3d36d82a1db76c5

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,175 @@
...
+                break;
+            }
+        }
+    }
+
+    /**
+     * Updates device and folder information in the cache model
+     * after a config update.
+     */
+    public void updateFromConfig(List<Device> newDeviceSet, List<Folder> newFolderSet) {
```
AudriusButkevicius: It's not a set like the variable name claims, perhaps just newDevices newFolders

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,175 @@
...
+        if (deviceFolderMap.containsKey(deviceId)) {
+            deviceFolderMap.remove(deviceId);
+        }
+    }
+    */
+
+    /**
+     * Adds a folder to the cache model if it does not exist.
+     */
+    private void addFolder(String deviceId, String folderId) {
```
AudriusButkevicius: Don't see this used.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,175 @@
...
+                }
+            }
+            if (!deviceFound) {
+                removedDevices.add(deviceId);
+            }
+        }
+        for (String deviceId : removedDevices) {
+            Log.v(TAG, "updateFromConfig: Remove device '" + deviceId + "' from cache model");
+            deviceFolderMap.remove(deviceId);
+        }
```
AudriusButkevicius: This block could live inside the `if (!deviceFound)` and remove the removedDevices

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,175 @@
...
+                }
+                if (!folderFound) {
+                    removedFolders.add(folderId);
+                }
+            }
+        }
+        for (String folderId : removedFolders) {
+            Log.v(TAG, "updateFromConfig: Remove folder '" + folderId + "' from cache model");
+            removeFolder(folderId);
+        }
```
AudriusButkevicius: Could call removeFolder straight in the `if (!folderFound)` block, and potentially log inside removeFolder rather than here.

app/src/main/java/com/nutomic/syncthingandroid/util/DefaultHashMap.java
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Class implementing a HashMap with a getOrDefault interface.
+ * HashMap requires at least API level >= 24 to support getOrDefault natively.
+ */
+public class DefaultHashMap<K,V> extends HashMap<K,V> {
+    public V getOrDefault(Object k, V defaultValue) {
+        return containsKey(k) ? this.get(k) : defaultValue;
+    }
+}
```
AudriusButkevicius: So I am pretty sure Java 8 has this out of the box. Where is the reference stating that this requires API level 24?

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,175 @@
...
+     * Adds a folder to the cache model if it does not exist.
+     */
+    private void addFolder(String deviceId, String folderId) {
+        // Add device parent node if it does not exist.
+        if (!deviceFolderMap.containsKey(deviceId)) {
+            deviceFolderMap.put(deviceId, defaultDevice);
+        }
+        // Add folder.
+        if (!deviceFolderMap.get(deviceId).containsKey(folderId)) {
+            deviceFolderMap.get(deviceId).put(folderId, defaultFolder);
```
AudriusButkevicius: All devices all folders will refer to the same object instance which I think is not what you want?

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,175 @@
...
+        for (String deviceId : removedDevices) {
+            Log.v(TAG, "updateFromConfig: Remove device '" + deviceId + "' from cache model");
+            deviceFolderMap.remove(deviceId);
+        }
+
+        // Handle devices that were added to the config.
+        for (Device device : newDeviceSet) {
+            if (!deviceFolderMap.containsKey(device.deviceID)) {
+                Log.v(TAG, "updateFromConfig: Add device '" + device.deviceID + "' to cache model");
+                deviceFolderMap.put(device.deviceID, defaultDevice);
```
AudriusButkevicius: Ditto about everyone referring to the same map. defaultDevice is essentially a pointer.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    /**
+     * Calculates remote device sync completion percentage across all folders
+     * shared with the device.
+     */
+    public int getDeviceCompletion(String deviceId) {
+        int folderCount = 0;
+        double sumCompletion = 0;
+        for (String folderId : deviceFolderMap.getOrDefault(deviceId, defaultDevice).keySet()) {
```
AudriusButkevicius: You are not using this correctly.

I suggest you go back to:
```
Map<String, CompletionInfo> completion = deviceFolderMap.get(deviceId);
if (completion == null) {
    return 0
}
// Don't know exact Java API but there must be a way to iterate key value pairs.
for (Pair<String, CompletionInfo> pair : completion.iterator()) {
  calculate compeltion
}
```

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,175 @@
...
+     */
+    public void setCompletionInfo(String deviceId, String folderId,
+                                    CompletionInfo completionInfo) {
+        // Add device parent node if it does not exist.
+        if (!deviceFolderMap.containsKey(deviceId)) {
+            deviceFolderMap.put(deviceId, defaultDevice);
+        }
+        // Add folder or update existing folder entry.
+        deviceFolderMap.get(deviceId).put(folderId, completionInfo);
+    }
```
AudriusButkevicius: Again, the default value is misused, perhaps do:

```
if deviceMap.contains(deviceId) {
  deviceMap.get(deviceId).put(folderId, completionInfo)
} else {
  Map<String, Completion> deviceCompletion = new HashMap<String, Completion>()
  deviceCompletion.put(folder, completion)
  deviceMap.put(device, deviceCompeltion)
}
```

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,175 @@
...
+
+        // Handle folders that were added to the config.
+        for (Folder folder : newFolderSet) {
+            for (Device device : newDeviceSet) {
+                if (folder.getDevice(device.deviceID) != null) {
+                    // folder is shared with device.
+                    if (!deviceFolderMap.get(device.deviceID).containsKey(folder.id)) {
+                        Log.v(TAG, "updateFromConfig: Add folder '" + folder.id +
+                                    "' shared with device '" + device.deviceID + "' to cache model.");
+                        deviceFolderMap.get(device.deviceID).put(folder.id, defaultFolder);
```
AudriusButkevicius: All devices all folders will refer to the same object instance which I think is not what you want?

app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
AudriusButkevicius: Why not call it setCompletionInfo then?

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
AudriusButkevicius: I don't even see us calling /completion, so I don't see why should we follow the completion endpoint if it's never called.

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
```diff
@@ -167,11 +188,29 @@ public void onEvent(Event event) {
...
+            case "RemoteIndexUpdated":
+                // ToDo
                 Log.v(TAG, "Unhandled event " + event.type);
+                break;
+            case "Starting":
+                break;
+            case "StartupComplete":
+                break;
+            case "StateChanged":
+                break;
```
AudriusButkevicius: You don't need to explicitly break on every event.
Just let them fall through and break before default.

Also, no point logging stuff about indexUpdate et all events if they are knowingly not handled. What value does logging stuff you know you don't handle provide?

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
```diff
@@ -93,16 +94,21 @@ public void onDone(long lastId) {
...
+
         switch (event.type) {
             case "ConfigSaved":
                 if (mApi != null) {
                     Log.v(TAG, "Forwarding ConfigSaved event to RestApi to get the updated config.");
                     mApi.reloadConfig();
                 }
                 break;
+            case "DeviceConnected":
+                break;
```
AudriusButkevicius: You are probably trying to sort it alphabetically, but I'd rather just wedge all the unhandled events at the bottom with a fallthrough down to the bottom and then break.

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
```diff
@@ -167,11 +188,29 @@ public void onEvent(Event event) {
...
+                Log.v(TAG, "Unhandled event " + event.type);
+                break;
+            case "LoginAttempt":
+                break;
             case "Ping":
                 // Ignored.
                 break;
-            default:
+            case "RemoteIndexUpdated":
+                // ToDo
```
AudriusButkevicius: Actually, having access to a computer and looking at the web UI code, it seems only LocalIndexUpdated triggers stuff, and none of the stuff it checks is visible in android (last scan, last item etc), so there is nothing todo under neither IndexUpdated events.

My bad, sorry.

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
```diff
@@ -145,6 +160,12 @@ public void onEvent(Event event) {
 
                 notify(title, pi);
                 break;
+            case "FolderScanProgress":
+                break;
+            case "FolderSummary":
+                Log.v(TAG, "Unhandled event " + event.type);
+                // ToDo
```
AudriusButkevicius: This should trigger `updateFolderStatus(RestApi)`, or rip the status out of the event and use that.
I understand that this won't necessarily happen in this PR.

AudriusButkevicius (changes requested):

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
Catfriend1: We'll need it in the future to display more info than just percentage.

app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
Catfriend1: ok

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: ok

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: Commented out.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: No, that was the case in an unsubmitted debug version I compiled and I got a java exception complaining about changing the map while it is accessed in a for iterator.

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
AudriusButkevicius: GUI calls it only after config save (for god knows what reason, probably legacy and can be removed), so I am not sure you will need it.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
AudriusButkevicius: Just remove, no point in keeping dead commented out code around

app/src/main/java/com/nutomic/syncthingandroid/util/DefaultHashMap.java
Catfriend1: Just remove DefaultFolderMap.java and try to build the project with native HashMap in place. Lint will tell you that you need to raise the API level. (I posted the message above in my comment)

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
Catfriend1: You're right. We should do this when I got more event handling implemented as we currently poll the folder db stats every second in FoldersAdapter.java. When this permanent banging of the db REST API got replaced, we can save resources by listening to the FolderSummary event. I'll add a comment in place.

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
Catfriend1: For now, I added a comment in place and commented the unused values out.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: I'll backup to my local machine so I can quickly get it back if needed.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: Yes, I reverted it to the new <object> code I had before.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: Well, you wanted it short and simple by getOrDefault, now it is: Let me explain it please. When a default device is returned because of a deviceId not in the model cache, it has no keyset of folders in it, so the for won't do anything. This will lead to assume the default return value of 100% - as "nutomic" had it in the connections UI code before. It was tested working. If not okay, I would go back to the check with containsKey and "auto-learning" mode to add missing deviceId to the HashMap.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: fixed, see above

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: fixed, see above

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: oh wait, in the mentioned lines is no getOrDefault nor is there a defaultFolder or deviceDevice ?

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
Catfriend1: No problem, your info is still valuable for me working on future PR :)

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,138 @@
...
+    DefaultHashMap<String, DefaultHashMap<String, CompletionInfo>> deviceFolderMap =
+        new DefaultHashMap<String, DefaultHashMap<String, CompletionInfo>>();
+
+    DefaultHashMap<String, CompletionInfo> defaultDevice = new DefaultHashMap<String, CompletionInfo>();
+
+    /**
+     * Removes a folder from the cache model.
+     */
+    private void removeFolder(String folderId) {
+        for (String deviceId : deviceFolderMap.keySet()) {
```
AudriusButkevicius: You can definately use an interator here as you are not modifying the map you are iterating over.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,138 @@
...
+            if (!deviceFolderMap.containsKey(device.deviceID)) {
+                Log.v(TAG, "updateFromConfig: Add device '" + device.deviceID + "' to cache model");
+                deviceFolderMap.put(device.deviceID, new DefaultHashMap<String, CompletionInfo>());
+            }
+        }
+
+        // Handle folders that were removed from the config.
+        List<String> removedFolders = new ArrayList<>();;
+        Boolean folderFound;
+        for (String deviceId : deviceFolderMap.keySet()) {
```
AudriusButkevicius: You can just iterate over values here, as deviceId is not used anywhere other than getting the values.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,138 @@
...
+
+        // Handle folders that were added to the config.
+        for (Folder folder : newFolders) {
+            for (Device device : newDevices) {
+                if (folder.getDevice(device.deviceID) != null) {
+                    // folder is shared with device.
+                    if (!deviceFolderMap.get(device.deviceID).containsKey(folder.id)) {
+                        Log.v(TAG, "updateFromConfig: Add folder '" + folder.id +
+                                    "' shared with device '" + device.deviceID + "' to cache model.");
+                        deviceFolderMap.get(device.deviceID).put(folder.id, new CompletionInfo());
```
AudriusButkevicius: Doing the get twice just seems wrong. Assign it to some variable somewhere and then do a put.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,138 @@
...
+
+    DefaultHashMap<String, CompletionInfo> defaultDevice = new DefaultHashMap<String, CompletionInfo>();
+
+    /**
+     * Removes a folder from the cache model.
+     */
+    private void removeFolder(String folderId) {
+        for (String deviceId : deviceFolderMap.keySet()) {
+            if (deviceFolderMap.get(deviceId).containsKey(folderId)) {
+                deviceFolderMap.get(deviceId).remove(folderId);
```
AudriusButkevicius: Get twice :(

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
AudriusButkevicius: I envisioned a different DefaultHashMap, which added the entry to the map and then returned if it was not present, avoiding these puts etc, but to be honest, it's probably better we don't rely on that magical insert if not exists behavior.

In this case it's easier just to do a get and do a null check rather than having a magic class that has one function that returns an empty object that has to be defined on the module scope.

Also the getOrDefault+get seems wrong, you could do a get once, null check, and then get the folder.

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
```diff
@@ -170,8 +181,23 @@ public void onEvent(Event event) {
...
+                // ToDo in a separate PR
+                // Call updateFolderStatus(RestApi api) to update the folder status UI
+            case "ItemStarted":
+            case "LocalIndexUpdated":
+            case "LoginAttempt":
+            case "RemoteIndexUpdated":
+            case "Starting":
+            case "StartupComplete":
+            case "StateChanged":
                 Log.v(TAG, "Unhandled event " + event.type);
```
AudriusButkevicius: These are handled, and ignored, there is a whole explicit switch statement handling these, no point logging it.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: Yes, but why do the work in second valid way? Java 5 allows us to direct usage of the iterator, with a for-each loop. What is the advantage? I've also in mind that an iterator does not solve the problem we are not allowed to immediately remove an element while iterating through.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: Yeah, and make it longer and import a new util.?!

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
Catfriend1: As I explained before, I like the log. It is just verbose level. ApiRequest is going wild on logging btw as it logs every second it does a query. We should something do about this limiting log spam.

Catfriend1:
Internal todo:
Improve getcompletion by null check
Improve double get's

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
Catfriend1: Or maybe we should call the message ignored event? Mostly devs care about verbose log and I haven't time to begin every coding session on a new topic with "oh my log points and indicators what st does when and why are missing and have to be inserted and later deleted". Plus, I can ask an experienced user to filter for eventprocessor to see if events were read, handled, ignored, unknown and the main plot of st usage was. Strategically this app needs a lot of bug fixes, enhancements that can easier be done with extensive logging. If the app reached a certain level of biggest bugs resolved for months, we can drop much logging or make it optional.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
AudriusButkevicius: Well it gets rid of the double call to get for starters. You are not removing elements from deviceFolderMap, you are removing it from the submap for the given device.

AudriusButkevicius: You can also simply use valueSet

Catfriend1:
About that pair iterator suggestion: that is another model type. I chose the hashmap instead of the pair class definition because a map has a unique key.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
AudriusButkevicius: What util import? Just iterate over map.entrySet()

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
AudriusButkevicius: So every time I had to look at someones provide logs from android, the log was full of lines telling me about events it didn't handle. I can read the code, I don't need the log to tell me that.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: Oh good point :)

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
AudriusButkevicius: Or even map.values()

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
Catfriend1: User says: after my folder download completed syncthing crashed. Log is full of other things but event would have told us. If you want to filter something out, use excel grep and so on.

AudriusButkevicius: What would a log line telling me RemoteIndexUpdated was dropped would have told me? The app does not use that, behavior of the app on wether it got logged or not does not change, so what value does it provide?

If I care about the events, I enable logging in synching, so I can see the payloads of the events not just the mere fact that it was emitted, god knows for what device what folder combination.

Sure, I can filter logs and what not, but it would be nice to have the log consumable out of the box eithout having to run it through grep.

Logs should be useful, not just verbose.

AudriusButkevicius:
When I say pair iterator, I mean entrySet, which returns key value pairs.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
AudriusButkevicius: Entryset is a method on the map, I am suprised that needs an iterator, but if it does, let it be, makes the code more concise and potentially avoids the repeated gets.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -112,9 +108,12 @@ public void updateFromConfig(List<Device> newDevices, List<Folder> newFolders) {
     public int getDeviceCompletion(String deviceId) {
         int folderCount = 0;
         double sumCompletion = 0;
-        for (String folderId : deviceFolderMap.getOrDefault(deviceId, defaultDevice).keySet()) {
-            sumCompletion += (deviceFolderMap.get(deviceId)).get(folderId).completion;
-            folderCount++;
+        HashMap<String, CompletionInfo> folderMap = deviceFolderMap.get(deviceId);
+        if (folderMap != null) {
+            for (String folderId : folderMap.keySet()) {
```
AudriusButkevicius: Iterate over values and just access completiom straight away without the need of double get...

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: Hmm.. doesn't compile:
/home/builduser/syncthing-android/app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java:26: error: incompatible types: Entry<String,HashMap<String,CompletionInfo>> cannot be converted to HashMap<String,HashMap<String,CompletionInfo>>
        for (HashMap<String, HashMap<String, CompletionInfo>> folderMap : deviceFolderMap.entrySet()) {
                                                                                                  ^

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: ok

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: fixed

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: Ok figured it out, import java.util.Map was required for entrySet.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: I have this already, uploading shortly ..

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: As I understand getValue() will copy the object, so I would leave the get as it is in the removeFolder function. We need a reference there. To improve it, I store the reference in folderMap after the one and only get.

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
Catfriend1: So lets find an agreement in the middle and only log that details in a debug build. That would be enough for me having the logs when working on the code.
https://github.com/syncthing/syncthing-android/pull/1079/commits/6e5293d79e797703f434732852abb5487525addc

Catfriend1:
Verified workin @https://github.com/syncthing/syncthing-android/pull/1079/commits/1d6d8df1b92f33ddcda64faea37915049f11c062

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
```diff
@@ -0,0 +1,143 @@
...
+
+    HashMap<String, HashMap<String, CompletionInfo>> deviceFolderMap =
+        new HashMap<String, HashMap<String, CompletionInfo>>();
+
+    /**
+     * Removes a folder from the cache model.
+     */
+    private void removeFolder(String folderId) {
+        HashMap<String, CompletionInfo> folderMap;
+        for (String deviceId : deviceFolderMap.keySet()) {
```
AudriusButkevicius: You can still iterate over values to avoid the get.

AudriusButkevicius (approved):
Appart from the last nitpick, looks good to me.

app/src/main/java/com/nutomic/syncthingandroid/model/Completion.java
Catfriend1: Ok, did it even though it's not 100% clear to me why we had too much get() in here and what the negative effects were. removeFolder() is expected to be called in rare circumstances when the user edits folders and the config reloads. CPU time isn't that rare than it was in the 90's. Well, I'd also like to have a good polished code before merge, yes, but if you can bear it please don't let us drive too many rounds looking for 3...4...5 alternative implementations that work unless it's really of crucial importance. Just needed to say that, please don't be disappointed. I really like to thank you for all honest ideas and feedback and - most valuable - the time you took on your side to stand by me. We'll get syncthing better together part by part for sure.

Catfriend1:
lgtm?

AudriusButkevicius:
Uff. You don't need map.entrySet, you can just do map.values

Catfriend1:
Ok, done.

Catfriend1:
Are we ready now?
Reminder: Please run tx push when you're near your computer

Issue #1078, opened 2018-05-04
WebUI changes not propagated back to the app


Catfriend1:
According to bug report https://github.com/syncthing/syncthing-android/pull/1051#issuecomment-386451058
AudriusButkevicius: "I've also noticed that if you change settings in the web ui, they are not propagated back to the app. I thought the app handles configSaved events."

Suggestion to fix it:
- EventProcessor should react to "ConfigSaved" events.

Catfriend1:
@AudriusButkevicius I did some tests and could easily reproduce the issue of configSaved not being handled and the app getting stuck if you press the "Restart Syncthing" button on the web UI.
Possible ways to solve this:
a) If the config is saved but syncthing core needs a restart, let the android app automatically do this. This would be my preferred way as it is easy and straightforward to implement.
b) Popping up the "Syncthing app needs a restart" notification in android's notification bar - as it's already done during the settings dialog for a postponed restart of syncthing. But then, we need to do an ugly workaround to also allow the user to restart syncthing core binary from the web UI (parse log for the process exit reason and restart the binary automatically from "SyncthingRunnable").
I'd like to have some feedback before I start to code please.

AudriusButkevicius:
Sure, but that has nothing todo with stale ui state

AudriusButkevicius:
To be honest, we should just restart syncthing when it needs it. Yet that could be weird, if yousave one folder and the UI goes down...

Nutomic:
I think it would be possible to restart only the app in that case, and leave Syncthing running. Then data in the app would be up to date, and the user can still restart Syncthing when they are done. So basically you would only restart the RestApi class, but not SyncthingRunnable. Not sure how much effort that would be though.

Catfriend1:
I made some draft code changes...
a) ConfigSaved triggers a ACTION_RESTART intent (like changing app's preferences) > found this too much as it's also happening if we change a folders option the app's UI (beside the intended restart when we use the webUI to change folder settings)
b) ConfigSaved tries to reinstantiate restApi and EventProcessor... we are not sure if we miss a sendConfig command from the app itself or an event will be fired after a regular sendConfig and ACTION_RESTART intent... hmm... not sure.
[...]

PR #1081, opened 2018-05-07
Reread the config on configSaved event

ghost:
Purpose
This PR fixes the issue described in https://github.com/syncthing/syncthing-android/issues/1078
Every time syncthing core notifies us about changed config, the config will be re-read.

Testing
Verified working on my phone lg-815 on Android 7.1.2

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
@@ -95,10 +95,10 @@ public class EventProcessor implements SyncthingService.OnWebGuiAvailableListene
     public void onEvent(Event event) {
         switch (event.type) {
             case "ConfigSaved":
-                Log.d(TAG, "Restarting syncthing due to ConfigSaved event.");
-                Intent iSyncthingService = new Intent(mContext, SyncthingService.class)
-                        .setAction(SyncthingService.ACTION_RESTART);
-                mContext.startService(iSyncthingService);
+                if (mApi != null) {
+                    Log.v(TAG, "Forwarding ConfigSaved event to RestApi to get the updated config.");
+                    mApi.onWebGuiAvailable();
+                }
                 break;
             case "DeviceRejected":
                 String deviceId = (String) event.data.get("device");
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
@@ -97,7 +97,7 @@ public class EventProcessor implements SyncthingService.OnWebGuiAvailableListene
             case "ConfigSaved":
                 if (mApi != null) {
                     Log.v(TAG, "Forwarding ConfigSaved event to RestApi to get the updated config.");
-                    mApi.onWebGuiAvailable();
+                    mApi.reloadConfig();
                 }
                 break;
             case "DeviceRejected":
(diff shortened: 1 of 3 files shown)


Nutomic (approved):

Nutomic:
Cool, this is much easier than I thought it might be :)

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
```diff
@@ -94,6 +94,12 @@ public void onDone(long lastId) {
     @Override
     public void onEvent(Event event) {
         switch (event.type) {
+            case "ConfigSaved":
+                if (mApi != null) {
+                    Log.v(TAG, "Forwarding ConfigSaved event to RestApi to get the updated config.");
+                    mApi.onWebGuiAvailable();
```
AudriusButkevicius: We should probably just have a mApi.reloadConfig and call that. It seems that on web ui available calls a bunch of other endpoints that have not changed and does some magic atomic increments.

```
        new GetRequest(mContext, mUrl, GetRequest.URI_CONFIG, mApiKey, null, result -> {
            Log.v(TAG, "reloadConfig: " + result);
            mConfig = new Gson().fromJson(result, Config.class);
            if (mConfig == null) {
                throw new RuntimeException("config is null: " + result);
            }
        });
```

Catfriend1: @Nutomic Could you please give me some advice if I need the full code triggering all classes to listen to the new, reloaded config as if the syncthing app restart option had been fired? Reading through the classes, I thought it would be the best way to fully perform a reinit of all associated classes (like when onWebGuiAvailable() regularly fires) to get the right increment counter for reading events, folderstatus model and so on. If we change the webgui port and address, it looks wrong to me just pushing a new config in and missing some steps. I hope you can help me clarify that.
@AudriusButkevicius Strategically I like your approach of minimalistic reloading the config and associated classes. Let's wait what Nutomic says and then decide how to generate the mApi.reloadConfig() function.

AudriusButkevicius: I can tell you from reading the code that you don't need that, as it fetches system info and device id, which is not changed by saving the config.

AudriusButkevicius: Port and what not is not handled by the existing code anyway

Catfriend1: Ok, agreed. I'll make a new suggestion.

app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
```diff
@@ -150,10 +158,6 @@ public void onWebGuiAvailable() {
             }
             tryIsAvailable();
         });
-        getSystemInfo(info -> {
```
AudriusButkevicius: I think the tryIsAvailable should not be there in case of configSaved?

Catfriend1: I tried on my phone, you're right. If tryIsAvailable() is there, the app just crashes.

AudriusButkevicius (approved):

Catfriend1:
I am currently verifying if it works crashless... hold on a moment. I'll give feedback very soon.

Catfriend1:
Ok, I verified this working on my phone including the latest commit https://github.com/syncthing/syncthing-android/pull/1081/commits/37f7f38c5650cbf31f0dbb116c68d244abc135f2 . One minor glitch that I think can be accepted was discovered testing it: When I change a folder and disable watch for changes in the web UI and go "as fast as I can" after hitting apply to the app's native UI, I no longer see the switch flip from on to off with a little delay waiting and staring at it. Should be okay, as I expect a normal user won't work in the Web UI and as fast after that on the native app UI - just a failing stress test. If I find a way to further improve this without firing the whole onWebGuiAvailable magic, I'll submit a separate PR :-).

PR #1093, opened 2018-05-18
Make current proxy text clearer

licaon-kter:
The current one sounds strange `Do not connect through proxy Example <proxy>`, so I setup a proxy to NOT connect? (Yes it got changed when edited...)

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -355,7 +355,7 @@ Please report any problems you encounter via Github.</string>
 
     <!-- Experimental options -->
     <string name="use_proxy">Connect through a proxy</string>
-    <string name="do_not_use_proxy">Do not connect through a proxy</string>
+    <string name="do_not_use_proxy">No proxy configured.</string>
 
     <string name="keep_wakelock_while_binary_running">Keep the CPU awake while Syncthing is running</string>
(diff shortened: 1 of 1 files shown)


Catfriend1:
I agree with the optimization suggested. Strings would have to be retranslated.

AudriusButkevicius:
I guess this needs to remove all existing strings from all files?

Catfriend1:
@AudriusButkevicius It is okay just changing "app/src/main/res/values/strings.xml". It will need a tx push, people retranslate the string (transifex will make a suggestion based on the old string's meaning and mark it as untranslated) and then tx pull.

AudriusButkevicius:
We'll do a tx push before the release.

Catfriend1:
Can we do one now or let's say about a week for the next release so people have the ability to translate the new UI stuff in the meantime? Last time only english language was available as tx push and the release had only minutes in-between.

AudriusButkevicius:
Whats the big deal with rushing the translations in? It can wait.

Catfriend1:
People would have a translated app on release date. Especially ones that have difficulties with understanding it in english. For me its fine, personally.

AudriusButkevicius:
Right, lets stop talking about this then.

PR #1097, opened 2018-05-24
fix parseInt exception - fixes https://github.com/syncthing/syncthing…

ghost:
Purpose
If the user enters an invalid integer in syncthing options for allowed bandwidth, a message is shown telling the accepted value range.

Testing
Verified working on Android 7.1.2 @ lg-h815.

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
@@ -322,7 +322,7 @@ public class SettingsActivity extends SyncthingActivity {
                     try {
                         maxRecvKbps = Integer.parseInt((String) o);
                     } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


AudriusButkevicius (approved):

AudriusButkevicius:
Isn't min value negative?

Catfriend1:
@AudriusButkevicius Right, fixed it as negative values don't make sense here.

AudriusButkevicius:
Isn't there an unsigined integer in java? People can still put in -1... Also, isn't there a way to validate this in the view/layout?

Catfriend1:
Hmm yes -1 may be a point but I would just suggest drop this case unhandled as syncthing core ui also does allow setting -1. People need to understand the options. Else we could make an if to check if its negative... Really necessary?

Catfriend1:
@AudriusButkevicius I looked again at the implementation and found it is correct. Java has no unsigned integers plus the layout disallows entering negative values. So, lgtm please.

AudriusButkevicius:
https://developer.android.com/reference/android/widget/TextView.html#attr_android:numeric

It seems that you van force only positive numbers in the view.

Issue #1100, opened 2018-05-26
Wrong text on "Loading Spinner UI"


Catfriend1:
The text shows "generating keys" during restart of the syncthing binary when it's not doing that step.

Reason:
On first app launch, an extra is put telling "hey this is first app launch". The Spinner UI dialog code picks it up and tells "generating keys". It also picks up the first launch extra (only used within the spinner UI code) when the first app assistant was completed and the user configures something and then restarts syncthing binary.

I'll make a PR to set the EXTRA_FIRST_START flag to false after the first-run syncthing binary initialization completed properly as this only will affect the spinner. Maybe we should rename it to something like EXTRA_KEY_GENERATION_IN_PROGRESS

### Version Information
    App Version: 0.10.10
    Syncthing Version: v0.14.47
    Android Version: Android 7.1.2 LOS

Catfriend1:
Closed by https://github.com/syncthing/syncthing-android/pull/1101

AudriusButkevicius:
You should put `(fixes #xxxx)` in your PR names.

PR #1101, opened 2018-05-26
Fix loading text

ghost:
Purpose
Fix issue described in https://github.com/syncthing/syncthing-android/issues/1100 .

Reason
The text shows "generating keys" during restart of the syncthing binary when it's not doing that step. The extra name first_start is placed misleading as the extra of the intent is about "generating keys" or "loading".

Testing
Verified working on lg-h815, Android 7.1.2


AudriusButkevicius:
Not sure how this fix works. What is changed here?

Nutomic:
The essential change is that the message "generating keys" is not shown after Syncthing was started for the first time. That was definitely a bug. Everything else is just renaming.

Nutomic (approved):

Catfriend1:
I'm sorry I just forgot about the "fixes " tag in the title. :(

PR #1102, opened 2018-05-27
Replace "grep" with java code (fixes #887)

### LINKED ISSUE #887: synthing isn't well terminated on android -ZenUI
I have installed syncthing on Zenpad (ZenUI launcher on )android)  and it works, but when I exit the programm regulary synything doesn't terminate right. When I use the quit button on the side menue, it seems to me that a new child process in background is starting.
Other connected devices (for example PC-client) still shows online status until zenpad device hibernates and background processes are shut down by the system. 
When I then turn on Zenpad device, there is still the message 'synthing is running' altough it isn't running anymore (even not in background - processes and child processes has been killed).This malfunction do make me very unhappy , because I like this piece of software very much and I would like to run it also on my Zenpad. But without the problem solved, its not useable for me. Has someone any similar experienes or an idea how to fix this ?
p.s. sorry for bad englisch (AUSTRIA)
Thanks for answers !! regards, joe
Version Information

    Syncthin-Android Version: 0.9.9
    Syncthing Version: v0.14.26
    Android Version: Android 6.0.1

ghost:
Purpose
This PR is intended to fix issue https://github.com/syncthing/syncthing-android/issues/887 by replacing the  "grep" in "ps | grep" shell code with java code in SyncthingRunnable.

Testing
Verified working on Android 7.1.2 LineAgeOS, device lg-h815. (Syncthing PID Found and Syncting PID #... killed lines are issued correctly in log) Used commit https://github.com/Catfriend1/syncthing-android/commit/5f07a896782984168d8d9d9582720c10ce31c4e0 to verify.

Not covered by Testing
I cannot test with Android ZenUI (on this platform the issue was originally reported). I can just confirm that "grep" is no longer necessary. This should be done by @joe147 @Joe7new .

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
@@ -12,6 +12,7 @@ import com.google.common.base.Charsets;
 import com.google.common.io.Files;
 import com.nutomic.syncthingandroid.R;
 import com.nutomic.syncthingandroid.SyncthingApp;
+import com.nutomic.syncthingandroid.service.Constants;
 
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
@@ -24,7 +25,9 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.LineNumberReader;
 import java.security.InvalidParameterException;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicReference;
 
@@ -49,6 +52,7 @@ public class SyncthingRunnable implements Runnable {
     private final Context mContext;
     private final File mSyncthingBinary;
     private String[] mCommand;
+    private List<String> mLibSyncthingPID = new ArrayList<String>();
     private final File mLogFile;
     @Inject SharedPreferences mPreferences;
     private final boolean mUseRoot;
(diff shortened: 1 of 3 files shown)


AudriusButkevicius:
It should be fixes #xxx not the full URL.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
```diff
@@ -208,6 +212,49 @@ private boolean useWakeLock() {
...
+                }
+            } catch (IOException e) {
+                Log.w(TAG_KILL, "Failed close the psOut stream", e);
+            }
+            if (ps != null) {
+                ps.destroy();
+            }
+        }
+    }
+
```
AudriusButkevicius: Why can't this just return the pid list? Whats the purpose of storing it on the class?

Catfriend1: I first thought of we could store our running syncthing instances on the class and later provide an interface "isBinaryRunning". This was when I discovered another issue (see other PR not relevant here)... but as this is unrelated to the duplicate starts of the binary, you're right. I'll improve that for simplicity :).

AudriusButkevicius (approved):

Catfriend1:
Verified working after latest commit https://github.com/syncthing/syncthing-android/pull/1102/commits/b745f01cf5022566df5ac64a7b72141533f1b011 , Thanks @AudriusButkevicius

Catfriend1:
@AudriusButkevicius I'll open a followup PR.

Joe7new:
Thanks for your efforts to you, catfriend1 ! sorry for commenting so late. I don't have the ZenUI-device anymore, so I couldn't test ....But again thanks for your work. Kind regards , Joe

PR #1105, opened 2018-05-27
remove duplicate SyncthingRunnable start trigger (fixes #1104)

### LINKED ISSUE #1104: Duplicate syncthing binary starts by SyncthingService
The log shows "Another syncthing instance is already running" and then next line "Restarting syncthing" every time I open the app or on first app start after installation.

To further diagnose the problem, I captured some logs with additional "log points" in place and indeed got multiple startup attempts of the SyncthingRunnable caused by SyncthingService.

Logs attached.
App Version: 0.10.10 (and newer commits)
Syncthing Version: v0.14.47
Android Version: Android 7.1.2 /  LineAgeOS

grafik
grafik
grafik
grafik
grafik

I've made a PR to fix this. Uploading it shortly...

ghost:
Purpose
In issue https://github.com/syncthing/syncthing-android/issues/1104, I discovered faulty startup behaviour of the SyncthingService. It starts the SyncthingRunnable.run() two times which results in one starting binary instance getting killed as "two instances running" will be caught by the app and fixed.
Cause is,
| new DeviceStateHolder(SyncthingService.this, this::updateState);
already has a callback to 
| updateState()
and does call it in foreground running mode and in background running condition.

Sidenote
I also suspect the "two simultaneously" starting syncthing binary instances to be the cause of some people's reported config corruption issue, e.g. https://forum.syncthing.net/t/all-settings-are-gone-after-updating-app/11514/9 .

Testing
Verified correctly working on commit https://github.com/syncthing/syncthing-android/commit/976b47532e0d6fa13c3bc84fe5cbb473635c466a under Android 7.1.2, device lg-h815.

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
@@ -9,6 +9,7 @@ import android.os.Bundle;
 import android.support.annotation.NonNull;
 import android.support.v4.app.ActivityCompat;
 import android.support.v4.content.ContextCompat;
+import android.util.Log;
 import android.view.View;
 import android.widget.Button;
 import android.widget.Toast;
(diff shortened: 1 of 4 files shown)


AudriusButkevicius:
I'll leave for @Nutomic

Nutomic:
Yes that makes sense.

PR #1108, opened 2018-05-28
RestApi - dont dump config to logcat (fixes #1107)

### LINKED ISSUE #1107: GUI password and API key is logged to logcat
Reproducer
Install Catlog app or use "adb logcat" and capture a log while syncthing is running. After a config change or reload, the config - including the sensitive password/key data - is dumped to the log output.

Suggestion
Depending on point of view, this can be considered a potential security leak under special circumstances. The API key and password should be replaced by generic tags before dumping parts of the config for easier debugging.

App Version: 0.10.10 (newer commits and previous releases)
Syncthing Version: v0.14.47
Android Version: Android 7.1.2

ghost:
Purpose
The config contains passwords and should not be dumped to logcat.
Following the proposal of log removal from https://github.com/syncthing/syncthing-android/issues/1107 .

Testing
Not necessary as this is just log stuff being removed. Builds okay.

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
@@ -163,11 +163,11 @@ public class RestApi implements SyncthingService.OnWebGuiAvailableListener {
     }
 
     private void onReloadConfigComplete(String result) {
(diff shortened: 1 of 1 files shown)


Catfriend1:
lgtm?

AudriusButkevicius (approved):

Issue #1116, opened 2018-06-02
DeviceStateHolder class gets instantiated one more time on launch by app drawer


Catfriend1:
Reproducer
Install syncthing app.
Go into settings by using the drawer on the left side.
Click home button
Open app drawer from android.
"Launch" the syncthing app.

Verified by putting a "Log.v" in DeviceStateHolder class' constructor and shutdown() function. I'm getting multiple new() registering more and more listeners and multiply calling back to SyncthingService for every single event occured but no shutdown() until I quit MainActivity.

Log
![image](https://user-images.githubusercontent.com/16361913/40877685-32f20442-6685-11e8-8468-aa47bc870904.png)
![image](https://user-images.githubusercontent.com/16361913/40877700-6bd8cbba-6685-11e8-8191-70b0c20607e6.png)
![image](https://user-images.githubusercontent.com/16361913/40877702-7741829e-6685-11e8-8afe-4ca8d4f88aa6.png)

    App Version: 0.10.10+
    Syncthing Version: v0.14.47
    Android Version: Android 7.1.2

Catfriend1:
I'm working improving the callback from DeviceStateHolder to SyncthingService so it only triggers a syncthing startup/termination one-time(!) if the run conditions really need this after a condition change.

Catfriend1:
Solved by commit https://github.com/Catfriend1/syncthing-android/commit/2fece28ab5f02ed905394983f0568e7816b09a7d

Issue #1115, opened 2018-06-02
SyncthingService endless restarts binary on run condition change


Catfriend1:
Reproducer on current master's code base
Install syncthing app while wifi is on.
Go to settings > run conditions > 
always run in background = yes
run on wifi = yes

Disable phone's wifi. Works.
Enable phone's wifi.  Endless syncthing binary restarts due to DeviceStateHolder issuing multiple change events to syncthing service even if the final decision to run or terminate the binary did not change.

Getting into it, I think the SyncthingService#updateState function containing an previously present in the code "HACK" is to blame. I'm currently investigating, if this fix helps:
`// HACK: Make sure there is no syncthing binary left running from an improper
            // shutdown (eg Play Store update).
            if (mCurrentState == State.INIT) {
                shutdown(State.INIT, () -> {
                    Log.i(TAG, "Starting syncthing according to current state and preferences");
                    new StartupTask().executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR);
                });
            }`

Log:
`06-02 14:06:52.609 I/DeviceStateHolder(9650): Listening for network state changes
06-02 14:06:52.619 I/DeviceStateHolder(9650): Listening to battery state changes
06-02 14:06:52.624 I/DeviceStateHolder(9650): Listening to power saving changes
06-02 14:06:52.652 I/DeviceStateHolder(9650): State updated, allowed network connection: true, charging: false, power saving: false
06-02 14:06:52.654 D/DeviceStateHolder(9650): All SSIDs allowed for syncing
06-02 14:06:52.655 I/DeviceStateHolder(9650): State updated, allowed network connection: true, charging: false, power saving: false
06-02 14:06:52.659 D/DeviceStateHolder(9650): All SSIDs allowed for syncing
06-02 14:06:52.659 I/DeviceStateHolder(9650): State updated, allowed network connection: true, charging: false, power saving: false
06-02 14:06:52.662 D/DeviceStateHolder(9650): All SSIDs allowed for syncing
06-02 14:06:52.669 I/DeviceStateHolder(9650): State updated, allowed network connection: true, charging: false, power saving: false
06-02 14:06:52.671 D/DeviceStateHolder(9650): All SSIDs allowed for syncing
06-02 14:06:52.777 I/SyncthingRunnableKill(9650): Killed Syncthing process 9714
06-02 14:06:52.792 I/DeviceStateHolder(9650): Listening for network state changes
[...]

PR #1117, opened 2018-06-02
Handle revoked storage permission gracefully (fixes #871 #1115 #1116)

### LINKED ISSUE #871: Syncthing silently breaks if permission gets removed
I recently tried a workaround for Nexus 6p battery drain issues - reset all app permissions. I went for 2 weeks before I noticed that syncthing isn't doing anything. The notification just says "syncthing is running" when charging & on wifi - per my config, but it isn't syncing because - of course, permission denied. Error is shown above folder icon once syncthing is opened. Clicking on that icon/text causes syncthing to force close.

Please **please** _please_ make syncthing loudly complain when the permission is not granted. There is zero reason to continue even running if the permission is not granted since, obviously, syncthing cannot do the one job it is supposed to do.

I would suggest to add an extra error notification (the non-permanent like "new email") kind - similar to e.g. gmail's "sign-in error" with a warning triangle or exclamation mark or a hypnotoad asking the user to click it and grant the permission.

Of course, the notification should show only if there is a valid sync configuration - meaning there is a shared folder and at least one peer with whom the folder is shared.

ghost:
Purpose

a) Handle revoked storage permission gracefully
Keep the SyncthingService off until the permission is re-granted by the user.
Immediately post a notification to the status bar explaining that syncthing was terminated as it needs the write permission to function.
Clicking the notification, the user is took to FirstStartActivity where the permission is rechecked and reasked. Denial keeps syncthing completely terminated, Granting will turn over to MainActivity and show the wrapper UI and run syncthing as the user normally expects it.
This will fix issue https://github.com/syncthing/syncthing-android/issues/871

b) Implementing the above, I came across the bug, that multiple SyncthingService and DeviceStateHolder instances are started when the user switches apps using the home button and gets into the syncthing app again by launching it from the app drawer. This is a big problem as unintended creation of multiple instances of the wrappers classes is happening and the permission revoke is not correctly handled on internal SyncthingService#onApiRequest calls thus resuming after re-granting the permission.
See issue https://github.com/syncthing/syncthing-android/issues/1116
This issue is present since the 0.10.10 (and earlier) releases.

c) Implementing a fix for the multiple class instantiating problem that came from b) I noticed a new bug. Limiting the instantiation to one FirstStartActivity (which "always" calls one SyncthingService, one MainActivity and one DeviceStateHolder subsequently) does not solve the "endless syncthing binary shutdown and restart requests" problem. See issue https://github.com/syncthing/syncthing-android/issues/1115 .
This issue is present since the current master commit https://github.com/syncthing/syncthing-android/commit/6a4c99848d7e2a4b9135b3a7cdc86b6889e05fcc . (tag unreleased)
I fixed it renaming the SyncthingService#updateState function to a resonable name and calling it only on change of the "binary should be running" decision. Before this PR, updateState was wrongly called on every run condition change that got monitored even if this had no influence on the "binary should run decision".
To sum up: The new function onUpdatedShouldRunDecision is called by DeviceStateHolder if the result of "DeviceStateHolder#shouldRun" is the opposite of the previously determined result.

Testing
Verified this works correctly on my Android 7.1.2 lg-h815 device.
- Case 1
Revoke storage permission while syncthing is running (without running in the background set)
image
Notification pops up
image
SyncthingService is restarted by android and correctly detects and handles the revoked permission situation
image
- Case 1a - Starting syncthing from the launcher and clicking "deny" when storage permission request is shown
image
- Case 1b - Clicking the notification from above (case 1ba), does the same as starting syncthing from the launcher (case 1bb)
Permission request for storage perm is shown -  pressed "accept".
Syncthing is up and running showing the MainActivity UI again. Log checked, SyncthingService running fine, DeviceStateHolder also listening for changed run conditions (if any are set) (case 1bba with run conditions + 1bbb without run conditions)
image

- Case 2
Set syncthing to run in background and only on wifi.
Rebooted the device to see if the changed SyncthingService/DeviceStateHolder instantiation behaviour (class gets only instantiated one time) works correctly on boot-starting the app automatically. Test successful, app started correctly in the background (wifi was still on on boot).
image
Turned off wifi on android UI. Syncthing disabled correctly popped up in the status bar as expected.
Log checked, everything fine as the syncthing binary got killed.
image
Started syncthing app from app drawer to get into the UI - works okay - says syncthing is disabled and quit.
Turned wifi back on, according to the log one instance of syncthing is running. No more problems found.
image

Comments
Sorry it was too hard for me to fix all this problems one by one as I originally intended to just solve issue https://github.com/syncthing/syncthing-android/issues/871 but as I see now cleary, the problems are related so changing one part of code will put the issue to another niche.

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/MainActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/MainActivity.java
@@ -14,6 +14,7 @@ import android.content.pm.PackageManager;
 import android.content.res.Configuration;
 import android.graphics.Bitmap;
 import android.graphics.drawable.BitmapDrawable;
+import android.Manifest;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
@@ -24,6 +25,7 @@ import android.support.design.widget.TabLayout;
 import android.support.v4.app.Fragment;
 import android.support.v4.app.FragmentManager;
 import android.support.v4.app.FragmentPagerAdapter;
+import android.support.v4.content.ContextCompat;
 import android.support.v4.view.GravityCompat;
 import android.support.v4.view.ViewPager;
 import android.support.v4.widget.DrawerLayout;
@@ -253,6 +255,17 @@ public class MainActivity extends StateDialogActivity
         setOptimalDrawerWidth(findViewById(R.id.drawer));
 
         onNewIntent(getIntent());
+    }
+
+    @Override
+    public void onResume() {
+        // Check if storage permission has been revoked at runtime.
+        if ((ContextCompat.checkSelfPermission(this, Manifest.permission.WRITE_EXTERNAL_STORAGE) !=
+            PackageManager.PERMISSION_GRANTED)) {
+            startActivity(new Intent(this, FirstStartActivity.class));
+            this.finish();
+        }
+        super.onResume();
     }
 
     @Override
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/NotificationHandler.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/NotificationHandler.java
@@ -26,7 +26,7 @@ public class NotificationHandler {
     private static final int ID_RESTART = 2;
     private static final int ID_STOP_BACKGROUND_WARNING = 3;
     private static final int ID_CRASH = 9;
-    private static final int ID_MISSINGPERM = 10;
+    private static final int ID_MISSING_PERM = 10;
     private static final String CHANNEL_PERSISTENT = "01_syncthing_persistent";
     private static final String CHANNEL_INFO = "02_syncthing_notifications";
     private static final String CHANNEL_PERSISTENT_WAITING = "03_syncthing_persistent_waiting";
@@ -182,7 +182,7 @@ public class NotificationHandler {
                 .setAutoCancel(true)
                 .setOnlyAlertOnce(true)
                 .build();
-        mNotificationManager.notify(ID_MISSINGPERM, n);
+        mNotificationManager.notify(ID_MISSING_PERM, n);
     }
 
     public void showRestartNotification() {
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/NotificationHandler.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/NotificationHandler.java
@@ -26,6 +26,7 @@ public class NotificationHandler {
     private static final int ID_RESTART = 2;
     private static final int ID_STOP_BACKGROUND_WARNING = 3;
     private static final int ID_CRASH = 9;
+    private static final int ID_MISSINGPERM = 10;
     private static final String CHANNEL_PERSISTENT = "01_syncthing_persistent";
     private static final String CHANNEL_INFO = "02_syncthing_notifications";
     private static final String CHANNEL_PERSISTENT_WAITING = "03_syncthing_persistent_waiting";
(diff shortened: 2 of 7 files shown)


app/src/main/java/com/nutomic/syncthingandroid/activities/MainActivity.java
```diff
@@ -255,6 +257,17 @@ public void onCreate(Bundle savedInstanceState) {
...
 
+    @Override
+    public void onResume() {
+        // Check if storage permission has been revoked at runtime.
+        if ((ContextCompat.checkSelfPermission(this, Manifest.permission.WRITE_EXTERNAL_STORAGE) !=
+            PackageManager.PERMISSION_GRANTED)) {
+            startActivity(new Intent(this, FirstStartActivity.class));
+            this.finish();
+        }
+        super.onResume();
```
AudriusButkevicius: No idea, but should this be an else?

app/src/main/java/com/nutomic/syncthingandroid/service/NotificationHandler.java
```diff
@@ -26,6 +26,7 @@
     private static final int ID_RESTART = 2;
     private static final int ID_STOP_BACKGROUND_WARNING = 3;
     private static final int ID_CRASH = 9;
+    private static final int ID_MISSINGPERM = 10;
```
AudriusButkevicius: ID_MISSING_PERM?

app/src/main/java/com/nutomic/syncthingandroid/service/NotificationHandler.java
```diff
@@ -171,6 +172,19 @@ public void showEventNotification(String text, PendingIntent pi, int id) {
         mNotificationManager.notify(id, n);
     }
 
+    public void showStoragePermissionRevokedNotification() {
+        Intent intent = new Intent(mContext, FirstStartActivity.class);
+        Notification n = getNotificationBuilder(mInfoChannel)
+                .setContentTitle(mContext.getString(R.string.syncthing_terminated))
+                .setContentText(mContext.getString(R.string.toast_write_storage_permission_required))
```
AudriusButkevicius: Don't see the string anywhere?

app/src/main/java/com/nutomic/syncthingandroid/activities/MainActivity.java
Catfriend1: According to the lint warning the super function is required to be called when overriding.

app/src/main/java/com/nutomic/syncthingandroid/service/NotificationHandler.java
Catfriend1: Its been reused from the Message that explains the storage permission in FirstStartActivity

app/src/main/java/com/nutomic/syncthingandroid/service/NotificationHandler.java
Catfriend1: Yes ill fix that...

app/src/main/java/com/nutomic/syncthingandroid/activities/MainActivity.java
Catfriend1: I took a quick search: https://stackoverflow.com/a/12975029 does explain the android activity lifecycle musn't be prevented by a missing super call.
https://developer.android.com/guide/components/activities/activity-lifecycle has a nice diagram where the activity lifecycle is shown.

Catfriend1:
@AudriusButkevicius Are we ready or can I still improve it?

PR #1130, opened 2018-06-10
fix lint "unused quantity translations"

ghost:
in /src/main/res/values-cs/strings.xml

Purpose - fix unnecessary build lint warnings by ignoring them. The cause of the warnings is people on transifex don't know certain strings needn't be translated in certain languages. If they do, they are unused string and even if this doesn't hurt, lint complains about them.

image

Verified building correctly. Build warnings decreased from 44 to 40.


Catfriend1:
[lint-results.zip](https://github.com/syncthing/syncthing-android/files/2087731/lint-results.zip)

PR #1131, opened 2018-06-11
Fix Android Studio 3.1.3 build-tool warnings

capi:
Adjust build-tools version and Gradle version to work with latest Android Studio (3.1.3).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,13 +2,11 @@
 
 buildscript {
     repositories {
+        google()
         jcenter()
-        maven {
-            url "https://maven.google.com"
-        }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.0.1'
+        classpath 'com.android.tools.build:gradle:3.1.3'
         classpath 'com.github.triplet.gradle:play-publisher:1.2.0'
         classpath 'com.github.ben-manes:gradle-versions-plugin:0.15.0'
 
@@ -19,10 +17,8 @@ buildscript {
 
 allprojects {
     repositories {
+        google()
         jcenter()
-        maven {
-            url "https://maven.google.com"
-        }
     }
 }
 
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,6 @@
+#Fri Jun 08 21:11:25 CEST 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.3.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.5-all.zip
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ android:
   components:
     - tools
     - platform-tools
-    - build-tools-27.0.2
+    - build-tools-27.0.3
     - android-27
     - extra-android-m2repository
 
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -21,7 +21,7 @@ dependencies {
 android {
     // Changes to these values need to be reflected in `.travis.yml`
     compileSdkVersion 27
-    buildToolsVersion "27.0.2"
+    buildToolsVersion '27.0.3'
 
     buildTypes.debug.applicationIdSuffix ".debug"
     dataBinding.enabled = true


capi:
Hmm build failes due to not-accepted license on build server. How is this fixed?
```
> Failed to install the following Android SDK packages as some licences have not been accepted.
     build-tools;27.0.3 Android SDK Build-Tools 27.0.3
  To build this project, accept the SDK license agreements and install the missing components using the Android Studio SDK Manager.
````

cc @AudriusButkevicius @Catfriend1

**Comment by imsodin** [comment]
Most likely the sdk version in the docker file also needs to be adjusted and the image needs to be rebuilt due to

```diff
-    buildToolsVersion "27.0.2"
+    buildToolsVersion '27.0.3'
```

capi:
Ouch, I did not yet care about the build servers so much, since I always build locally (because I installed initially from some modification that already got merged). If I fix the Docker file, who can publish it?

AudriusButkevicius:
I'll have to rebuild the docker image if we merge.

build.gradle
```diff
@@ -6,9 +6,10 @@ buildscript {
         maven {
             url "https://maven.google.com"
         }
+        google()
```
AudriusButkevicius: Isn't this the same as the section above? Can we drop that section and move google up top to address the fdroid issue (on the forum)?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Fri Jun 08 21:11:25 CEST 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.3.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
```
AudriusButkevicius: Why not 4.8 or whatever is latest?

Catfriend1:
We should be careful about updating the build parameters as @Nutomic bumped up the tools some time before and ended up in revert.
See for example https://github.com/syncthing/syncthing-android/commit/473b2040fcedda3c84af7e93c6814d7a69411ab0 and https://github.com/syncthing/syncthing-android/commit/c3be04a3f23adefefca07a209b54a87e21507194

build.gradle
capi: Will try that.

gradle/wrapper/gradle-wrapper.properties
capi: I just had Android Studio apply it's automatic changes, and I think it's just going to the lowest possible version, but not sure.

capi:
I agree. I did not think that it caused so much issues with the build-server, but in hindsight it's obvious that it would have influence on that. I just applied the suggested changes to remove warnings from current Android Studio about too low versions for Gradle and build tools, and I supposed it would be better to have those changes in repository and not only locally. I was over-optimistic :-)

capi:
@imsodin Can you explain to me, what exactly the difference between the continuous build of Travis vs Build is? Are both needed? Is Travis maybe simply outdated?

Catfriend1:
@capi Please have a look at travis.yml
There is a line 
yes | sdkmanager "platforms;android-27"
I think you need to accept the missing license there.

gradle/wrapper/gradle-wrapper.properties
AudriusButkevicius: Let's check if it works with 4.8 and move to that.

capi: @AudriusButkevicius Seems that Android Studio is not too happy with Gradle >= 4.6:
```
Configuration on demand is not supported by the current version of the Android Gradle plugin since you are using Gradle version 4.6 or above. Suggestion: disable configuration on demand by setting org.gradle.configureondemand=false in your gradle.properties file or use a Gradle version less than 4.6.
```
I'll try with 4.5 instead.

Catfriend1:
@capi: did you make somr rough tests to make sure the app works when built this way? If yes, lgtm

build.gradle
```diff
@@ -2,13 +2,11 @@
 
 buildscript {
     repositories {
+        google()
```
AudriusButkevicius: This needs to be fixed in a few places I think.

capi: One more in the same file, true. I have pushed an updated version of `build.gradle`.

capi:
@Catfriend1 I have the locally built version installed on my phone. Behaves exactly like `master`.

PR #1134, opened 2018-06-13
Rename sttrace pref to debug_facilities_enabled (fixes #1133)

### LINKED ISSUE #1133: Syncthing crashes on startup 0.10.11
syncthing crashes on startup since last update from play store

App Version 0.10.11
Version in Playstore 0.14.48
Android 7.1.2 
LineageOS 14.1-20180606-NIGHTLY-bacon

Syncthing crashes on startup (also after reboot of device)
Reproducable on my sytstem

ghost:
in Constants.java

Purpose
Users having once set the sttrace/string pref in app versions before 0.10.11 cannot upgrade without this PR as the type mismatch between app versions results in the crash described in issue #1133 .

As discussed before, we'll fix it by renaming the pref to "debug_facilities_enabled".
Sidenote: I did not rename the pref storing available debug facility options as that can only be done before the next syncthing binary update gets released with the app. The reason is, this pref is only propagated once per syncthing binary upgrade. If we need an interim release, we would be out of luck. I'll schedule it for later to fix up the pref cache's name.

Testing
Verified working @ https://github.com/syncthing/syncthing-android/pull/1134/commits/2556a34a1731b0d64d1b20892ea80e29f199b94f

Related issue #1146

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
@@ -78,7 +78,6 @@ public class SettingsActivity extends SyncthingActivity {
             Preference.OnPreferenceClickListener, SharedPreferences.OnSharedPreferenceChangeListener {
 
         private static final String TAG = "SettingsFragment";
-        private static final String KEY_STTRACE = "sttrace";
         private static final String KEY_EXPORT_CONFIG = "export_config";
         private static final String KEY_IMPORT_CONFIG = "import_config";
         private static final String KEY_ST_RESET_DATABASE = "st_reset_database";
@@ -193,10 +192,10 @@ public class SettingsActivity extends SyncthingActivity {
             Preference exportConfig = findPreference("export_config");
             Preference importConfig = findPreference("import_config");
 
-            Preference stTrace              = findPreference("sttrace");
-            Preference environmentVariables = findPreference("environment_variables");
-            Preference stResetDatabase      = findPreference("st_reset_database");
-            Preference stResetDeltas        = findPreference("st_reset_deltas");
+            Preference debugFacilitiesEnabled   = findPreference(Constants.PREF_DEBUG_FACILITIES_ENABLED);
+            Preference environmentVariables     = findPreference("environment_variables");
+            Preference stResetDatabase          = findPreference("st_reset_database");
+            Preference stResetDeltas            = findPreference("st_reset_deltas");
 
             mUseRoot                        = (CheckBoxPreference) findPreference(Constants.PREF_USE_ROOT);
             mUseWakelock                    = (CheckBoxPreference) findPreference(Constants.PREF_USE_WAKE_LOCK);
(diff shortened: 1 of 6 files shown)


AudriusButkevicius (approved):

AudriusButkevicius:
So the change looks ok to me, yet I am not sure I understand what this breaks in terms of caching.

AudriusButkevicius:
If it crashed, it should not have stuff cached, hence should be fixed after this amd cache after this. If it did cache, it should still be ok as I would not have expected things to change.

Catfriend1:
@AudriusButkevicius Verified this working. Lgtm

AudriusButkevicius:
The other var was always a set and not a string, so why should it matter/break?

Catfriend1:
It would be empty until syncthing binary is upgraded.

AudriusButkevicius:
Sure, but why would that cause a crash? Empty falls back to thw hardcoded list

Catfriend1:
It would not crash, but fail to read to read the cache from rest. For this PR everything is fine.

AudriusButkevicius:
Well it would fail to read from the cache on 0.10.10 to 0.10.11 upgrade, how is it going to be different this time?

Catfriend1:
The cache is not handled in this pr its the selection. I'll explain it in the follow up pr again before St version gets bumped up. Btw it only caches once on binary upgrade

AudriusButkevicius:
We should fix that it caches on startup.

AudriusButkevicius:
As there might be multiple android versions before syncthing gets upgraded.

Catfriend1:
We'll leave caching as it is and rename the var in a follow up PR as its a one time problem. Doesnt need to get fatter. This pr is - again - not about caching but fixing the problem the app does not start coming from 0.10.10 to 0.10.11

Catfriend1:
And yes there can be multiple android versions in between. That's no problem.

Catfriend1:
I'll make the follow-up PR ready and setting the milestone marker to "0.14.49" so it can get merged at the right time. We have then less code complexity.

Catfriend1:
Follow-up PR #1136
Comparison explains what was misunderstood above : https://github.com/Catfriend1/syncthing-android/compare/fixSttracePrefMig...Catfriend1:renameSttraceCache?expand=1

Catfriend1:
lgtm

Catfriend1:
We can't because this is not necessary and you don't like hardcoded lists.

Timeline:
a) Merge this PR 1134
b) Make a release (so people locked out of syncthing completely can start it again)
c) (Optional) Do more android app releases
d) Bump syncthing core version to v0.14.49 when it's time
e) Merge PR #1136 (see milestone)
f) Cache will be updated with new debug facilities

Following this strategy, no more code changes are necessary and we ensure having the debug facilities read from the REST API available all the time through. I am not against your suggestion with the hardcoded list but also don't want to introduce new problems on the way.

Reading the syncthing core PR's I also expect we have to add more migration work to the v0.14.49 milestone like the readwrite folder rename stuff?!

AudriusButkevicius:
Sorry, your argument makes no sense.
If we rename the variable now, this will still work:
https://github.com/syncthing/syncthing-android/blob/6a4c99848d7e2a4b9135b3a7cdc86b6889e05fcc/app/src/main/java/com/nutomic/syncthingandroid/views/SttracePreference.java#L90

and there will still be stuff, hence no coordination is required?

Sure, I don't like hardcoded lists, but I think we are already past that point, as this is already merged?

This sort of coordination fiasco you are suggesting we do is exactly the reason why I didn't want to mess around with this.

Catfriend1:
I'll leave it up to you when the two pr get merged. Just wanted to suggest a merge plan like syncthing core already has it as calmh?! merges with the master and picks specific master commits that were planned to go into release next. If you don't like the idea you can merge the two fixing the sttrace problems as they are verified working.

PR #1136, opened 2018-06-13
Rename sttrace_cache variable

ghost:
Purpose
The variable should be renamed to correspond to "debug_facilities_enabled" from PR #1134.

Side Effects
As it changes its name, it would end up empty on app upgrade.

Milestone for merging
This is the reason for that this PR may not be merged before the "Bumped syncthing to 0.14.49" milestone. If merged on that milestone reach, the syncthing binary update will trigger the new variable to be refilled and the existing code will work properly.

Related issue #1146


Catfriend1:
Comments that belong to this PR taken from #1134

AudriusButkevicius
commented 1 day ago

I still don't understand why we can't rename the var now, run with that value empty, and populate from a static list in the meanwhile until syncthing gets bumped?
@Catfriend1
Catfriend1
commented about 14 hours ago

We can't because this is not necessary and you don't like hardcoded lists.

Timeline:
a) Merge this PR 1134
b) Make a release (so people locked out of syncthing completely can start it again)
c) (Optional) Do more android app releases
d) Bump syncthing core version to v0.14.49 when it's time
e) Merge PR #1136 (see milestone)
f) Cache will be updated with new debug facilities

Following this strategy, no more code changes are necessary and we ensure having the debug facilities read from the REST API available all the time through. I am not against your suggestion with the hardcoded list but also don't want to introduce new problems on the way.

Reading the syncthing core PR's I also expect we have to add more migration work to the v0.14.49 milestone like the readwrite folder rename stuff?!

AudriusButkevicius:
This can be closed as it does the same thing?

Catfriend1:
This just renames the pref where the cache lies in. Recommended for merge with 0.14.49 , can be merged now if you insist. Just cosmetics because one var is now called sttrace_available and one debug_facilities_enabled .

AudriusButkevicius:
This has merge conflict.

Catfriend1:
Ah okay its the base. I need to rebase this as the only change relevant is

public static final String PREF_DEBUG_FACILITIES_AVAILABLE  = "debug_facilities_available";

Catfriend1:
Merge conflicts resolved. Ready.

AudriusButkevicius (approved):

PR #1140, opened 2018-06-16
Fix static leak in ShareActivity#CopyFilesTask

ghost:
Purpose
same as #1137 This is PR 2 out of 4 to fix the build warnings about static leaks in AsyncTask's.
Related #1140 #1141 #1142 

Testing
I still need to test this.

PR #1137, opened 2018-06-14
Fix leak in SyncthingService.StartupTask (fixes #1135)

ghost:
Purpose
Fix issue https://github.com/syncthing/syncthing-android/issues/1135 . Will also get rid of the lint warnings about the wrong implementation of the AsyncTask in SyncthingService.StartupTask .
This is no urgent problem, according to StackOverflow https://stackoverflow.com/questions/44309241/warning-this-asynctask-class-should-be-static-or-leaks-might-occur/46166223#46166223 its not leaking in most use cases.
See issue for more details and discussion about the way chosen to solve this.
Related #1140 #1141 #1142 

Testing
Verified working correctly under Android 7.1.2 , lg-h815.

Merging aspects
This is the first out of four PR's to fix the "AsyncTask class should be implemented as private static" problems from the current lint report. So please wait until I got all four PR ready, then I'll test through everything and we can merge them.

PR #1141, opened 2018-06-16
Fix static leak in SettingsActivity#TestRootTask

ghost:
Purpose
same as #1137 #1140 #1142 This is PR 3 out of 4 to fix the build warnings about static leaks in AsyncTask's.

Testing
I still need to do this.


AudriusButkevicius (approved):

Catfriend1:
Verified working. Lgtm.

PR #1142, opened 2018-06-16
Fix static leak in LogActivity#AsyncTask

ghost:
Purpose
same as #1137 #1140 #1141 This is PR 4 out of 4 to fix the build warnings about static leaks in AsyncTask's.

Testing
I still need to do this.

Note
I put code that was already existing and okay as it is.

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
@@ -17,6 +17,7 @@ import com.nutomic.syncthingandroid.R;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
+import java.lang.ref.WeakReference;
 
 /**
  * Shows the log information from Syncthing.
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/nutomic/syncthingandroid/activities/LogActivity.java
```diff
@@ -97,63 +98,80 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
 
-    private void scrollToBottom() {
-        mScrollView.post(() -> mScrollView.scrollTo(0, mLog.getBottom()));
-    }
-
     private void updateLog() {
-        if (mFetchLogTask != null)
+        if (mFetchLogTask != null) {
             mFetchLogTask.cancel(true);
+        }
```
AudriusButkevicius: no return?

Catfriend1: I thought about this when making the PR. The original code intends to cancel and relaunch the asynctask, so no return here. I'll test this currently on my phone.

AudriusButkevicius (approved):

Issue #1146, opened 2018-06-16
A lot more crashes more since 0.10.11


AudriusButkevicius:
![image](https://user-images.githubusercontent.com/1144861/41501915-440674a4-71a5-11e8-8de7-c0918bb9dc7a.png)

```
java.lang.NullPointerException: 
  at com.nutomic.syncthingandroid.activities.MainActivity.lambda$onServiceStateChange$0$MainActivity (MainActivity.java:116)
  at com.nutomic.syncthingandroid.activities.MainActivity$$Lambda$0.onResult (Unknown Source:4)
  at com.nutomic.syncthingandroid.service.RestApi.lambda$getSystemInfo$7$RestApi (RestApi.java:398)
  at com.nutomic.syncthingandroid.service.RestApi$$Lambda$7.onSuccess (Unknown Source:2)
  at com.nutomic.syncthingandroid.http.ApiRequest.lambda$connect$0$ApiRequest (ApiRequest.java:98)
  at com.nutomic.syncthingandroid.http.ApiRequest$$Lambda$0.onResponse (Unknown Source:4)

java.lang.ClassCastException: 
  at android.app.SharedPreferencesImpl.getStringSet (SharedPreferencesImpl.java:234)
  at com.nutomic.syncthingandroid.service.SyncthingRunnable.buildEnvironment (SyncthingRunnable.java:435)
  at com.nutomic.syncthingandroid.service.SyncthingRunnable.run (SyncthingRunnable.java:133)
  at com.nutomic.syncthingandroid.service.SyncthingRunnable.run (SyncthingRunnable.java:106)

Caused by: java.lang.NullPointerException: 
  at com.nutomic.syncthingandroid.service.RestApi.editFolder (RestApi.java:285)
  at com.nutomic.syncthingandroid.activities.FolderActivity.updateFolder (FolderActivity.java:486)
  at com.nutomic.syncthingandroid.activities.FolderActivity.onPause (FolderActivity.java:253)

java.lang.NullPointerException: 
  at com.nutomic.syncthingandroid.service.RestApi.getOptions (RestApi.java:371)
  at com.nutomic.syncthingandroid.activities.MainActivity.lambda$showUsageReportingDialog$10$MainActivity (MainActivity.java:477)
  at com.nutomic.syncthingandroid.activities.MainActivity$$Lambda$9.onClick (Unknown Source)

java.lang.NullPointerException: 
[...]

PR #1149, opened 2018-06-17
Fix NPE in DrawerFragment if syncthing is terminated (fixes #1148)

### LINKED ISSUE #1148: NPE in DrawerFragment if syncthing is terminated
If syncthing was terminated while Drawer is expanded and run conditions change:
When the user then would hit "show qr code" a NPE occured before the PR. (see list in issue #1146)

Error Log
java.lang.NullPointerException: at com.nutomic.syncthingandroid.service.RestApi.getGui (RestApi.java:375) at com.nutomic.syncthingandroid.fragments.DrawerFragment.showQrCode (DrawerFragment.java:210) at com.nutomic.syncthingandroid.fragments.DrawerFragment.onClick (DrawerFragment.java:241)

ghost:
Purpose
Fix NPE if syncthing was terminated while Drawer is expanded and run conditions change. When the user then would hit "show qr code" a NPE occured before the PR. (see list in issue #1146, issue #1148)

Error Log
java.lang.NullPointerException: at com.nutomic.syncthingandroid.service.RestApi.getGui (RestApi.java:375) at com.nutomic.syncthingandroid.fragments.DrawerFragment.showQrCode (DrawerFragment.java:210) at com.nutomic.syncthingandroid.fragments.DrawerFragment.onClick (DrawerFragment.java:241)

Testing
Just a null check, don't think it needs testing. Please tell me if it does.

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/fragments/DrawerFragment.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/fragments/DrawerFragment.java
@@ -4,6 +4,7 @@ import android.content.Intent;
 import android.os.Bundle;
 import android.support.v4.app.Fragment;
 import android.support.v4.content.ContextCompat;
+import android.util.Log;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
@@ -37,6 +38,8 @@ import java.util.TimerTask;
  * Displays information about the local device.
  */
 public class DrawerFragment extends Fragment implements View.OnClickListener {
+
+    private static final String TAG = "DrawerFragment";
 
     private TextView mCpuUsage;
     private TextView mRamUsage;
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/nutomic/syncthingandroid/fragments/DrawerFragment.java
```diff
@@ -206,10 +206,15 @@ private void onReceiveConnections(Connections connections) {
      */
 
     private void showQrCode() {
+        RestApi restApi = mActivity.getApi();
+        if (restApi == null) {
+            Toast.makeText(mActivity, R.string.syncthing_terminated, Toast.LENGTH_SHORT).show();
+            return;
+        }
+        String apiKey = restApi.getGui().apiKey;
+        String deviceId = restApi.getLocalDevice().deviceID;
```
AudriusButkevicius: Wouldn't these throw exceptions if syncthing was shutdown after checking rest api but before calling getLocalDevice etc?

Catfriend1: Maybe that can happen, but the shutdown then has to occur simultaneously from another thread which we currently don't have.

Catfriend1: Any suggestions how to improve this or can we merge it as a first-aid?

AudriusButkevicius: All of this startup, shutdown, stuff and element access should be behind a lock, there is a ton of NPE's happening because we just happen to access these global things without any locks at random times.

AudriusButkevicius: Perhaps wrap this whole thing in a try catch, and say an error has occoured if you get an exception.

app/src/main/java/com/nutomic/syncthingandroid/fragments/DrawerFragment.java
```diff
@@ -211,15 +214,19 @@ private void showQrCode() {
...
+            String deviceId = restApi.getLocalDevice().deviceID;
+            URL url = restApi.getUrl();
+            //The QRCode request takes one paramteer called "text", which is the text to be converted to a QRCode.
+            new ImageGetRequest(mActivity, url, ImageGetRequest.QR_CODE_GENERATOR, apiKey,
+                    ImmutableMap.of("text", deviceId),qrCodeBitmap -> {
+                mActivity.showQrCodeDialog(deviceId, qrCodeBitmap);
+                mActivity.closeDrawer();
+            }, error -> Toast.makeText(mActivity, R.string.could_not_access_deviceid, Toast.LENGTH_SHORT).show());
+        } catch (Exception e) {
+            Log.e(TAG, "showQrCode", e);
```
AudriusButkevicius: Do we need to show something to the user, or will it be obvious once the drawer is closed?

Catfriend1: I think in most cases a problem occurs here we'll hit the part that shows the toast to the user. That's enough for me.
> if (restApi == null) {
 -        String deviceId = mActivity.getApi().getLocalDevice().deviceID;		 +            Toast.makeText(mActivity, R.string.syncthing_terminated, Toast.LENGTH_SHORT).show();
 -        URL url = mActivity.getApi().getUrl();		 +            return;
 -        new ImageGetRequest(mActivity, url, ImageGetRequest.QR_CODE_GENERATOR, apiKey,		 +        }

AudriusButkevicius (approved):

PR #1151, opened 2018-06-17
Fix ClassCastException in SyncthingService onCreate (fixes #1150)

### LINKED ISSUE #1150: java.lang.ClassCastException in SyncthingService onCreate()
Error Log from GPlay Console
Caused by: java.lang.ClassCastException: 
  at com.nutomic.syncthingandroid.service.SyncthingService.onCreate (SyncthingService.java:136)

AppV 0.10.11

ghost:
Purpose
Fix issue discovered in GPlay Console #1150 .

According to https://stackoverflow.com/questions/8040226/android-classcast-exception-when-binding-to-service the class cast exception occurs because of the generation of a SyncthingServiceBinder class (which extens IBinder) and only referring to IBinder (instead of SyncthingServiceBinder) in onServiceConnected() functions. This might only be a problem on some Android versions, as its no problem on my personal device running Android 7.1.2. Fixed it by type-casting the IBinder correctly after the function override that has to declare IBinder.

Error Log from GPlay Console
Caused by: java.lang.ClassCastException:
at com.nutomic.syncthingandroid.service.SyncthingService.onCreate (SyncthingService.java:136)

Testing
Verified working on Android 7.1.2, device lg-h815.

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/FolderPickerActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/FolderPickerActivity.java
@@ -32,6 +32,7 @@ import com.google.common.collect.Sets;
 import com.nutomic.syncthingandroid.R;
 import com.nutomic.syncthingandroid.SyncthingApp;
 import com.nutomic.syncthingandroid.service.SyncthingService;
+import com.nutomic.syncthingandroid.service.SyncthingServiceBinder;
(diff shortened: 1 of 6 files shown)


AudriusButkevicius:
If you read the comments on the accepted answer, it does not help. It seems the fix is something todo with the manifest.

Catfriend1:
Yes, I read them.  We don't have the wrong process flag in the manifest. It's the type cast that has to be explicitly.

PR #1153, opened 2018-06-17
Fix NPE during FolderActivity#addDeviceViewAndSetListener (fixes #1152)

### LINKED ISSUE #1152: NPE during FolderActivity#addDeviceViewAndSetListener
Error Log
java.lang.NullPointerException: 
  at com.nutomic.syncthingandroid.activities.FolderActivity$2.onCheckedChanged (FolderActivity.java:111)

AppV 0.10.11 (and expected to be a bug in older versions)

Mentioned in issue list #1146

ghost:
Purpose
In FolderActivity: 
addDeviceViewAndSetListener#deviceView.setChecked
there is a mistake re-registering the listener if it is already registered when the syncthing run condition changes from ACTIVE>DISABLED>INIT>STARTING>ACTIVE during a folder edit. The re-registration itself is not the problem, but calling setChecked while the onCheckedChange listener is still registered will cause the messy NPE when a device has been removed in the meantime, e.g. using the WebUI or a config editor.
Fixed by the approach to properly unregister the listener (for device sliders) before calling setChecked from the code where we don't want the onCheckedChange-Event to fire. (The same is already implemented correctly for the other sliders {paused, sendonly, fswatcher}).

Error Log
java.lang.NullPointerException: 
  at com.nutomic.syncthingandroid.activities.FolderActivity$2.onCheckedChanged (FolderActivity.java:111)

Testing
I did not test this, think it's obvious what was causing the problem here and that the fix is okay. Please tell me if it needs testing.

Fixes issue #1152

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
@@ -480,6 +480,7 @@ public class FolderActivity extends SyncthingActivity
     private void addDeviceViewAndSetListener(Device device, LayoutInflater inflater) {
         inflater.inflate(R.layout.item_device_form, mDevicesContainer);
         SwitchCompat deviceView = (SwitchCompat) mDevicesContainer.getChildAt(mDevicesContainer.getChildCount()-1);
+        deviceView.setOnCheckedChangeListener(null);
         deviceView.setChecked(mFolder.getDevice(device.deviceID) != null);
         deviceView.setText(device.getDisplayName());
         deviceView.setTag(device);
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/FolderPickerActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/FolderPickerActivity.java
@@ -32,6 +32,7 @@ import com.google.common.collect.Sets;
 import com.nutomic.syncthingandroid.R;
 import com.nutomic.syncthingandroid.SyncthingApp;
 import com.nutomic.syncthingandroid.service.SyncthingService;
+import com.nutomic.syncthingandroid.service.SyncthingServiceBinder;
 import com.nutomic.syncthingandroid.util.Util;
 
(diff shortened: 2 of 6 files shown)


app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
```diff
@@ -480,6 +480,7 @@ private void addEmptyDeviceListView() {
     private void addDeviceViewAndSetListener(Device device, LayoutInflater inflater) {
         inflater.inflate(R.layout.item_device_form, mDevicesContainer);
         SwitchCompat deviceView = (SwitchCompat) mDevicesContainer.getChildAt(mDevicesContainer.getChildCount()-1);
+        deviceView.setOnCheckedChangeListener(null);
```
AudriusButkevicius: Why would you not do it why all other deregistrations are done?

Catfriend1: If you compare it to the other AddListeners function, it's the same way of successfully suppressing the event when we (initially) change the slider from code. There is no onDestroy() deregistration of the listeners in place as it's not necessary to do so. (SettingsActivity also uses the the above approach...) Otherwise, we would need a boolean in the class' context to be set true if we fire the event by code instead of the UI triggering it.

AudriusButkevicius: So this essentially removes the listener that was previously registered? Why don't re unset it as the view gets destroyed?

Unsetting on creation feels wrong.

Catfriend1: I think it's the function name I kept from the existing code that is quite confusing. This function is called whenever the UI is refreshed, so it's not a unset on creation. To be more precise, we don't kill a listener here but the function that fires if the event occurs. So that's okay, just believe me :)

AudriusButkevicius:
Please, start your PRs with a capital letter.

PR #1161, opened 2018-06-19
Add randomization to default camera folder name (fixes #1159)

### LINKED ISSUE #1159: Default camera folder name is the same on multiple identical devices and should be unique
After installing Syncthing for Android on two family member's phones that are the same make and model, I encountered confusing behavior where photos taken on one phone were showing up on the other.

The problem was caused by Syncthing on both phones coming by default with a camera folder with the same Folder ID, based on the device make and model ("lg-m210-photos" in this case).

I've now manually fixed the problem, by removing the folder on one device and recreating it with a different ID.

I hope future versions of the app will assign a default camera folder name that is almost certain to be unique, even on multiple identical devices.

For example, how about appending a couple of random letters, e.g. on these LG M210 phones the folder names might have defaulted to "lg-m210-photos-hz" and "lg-m210-photos-ej".

### Version Information

    App Version: 0.10.4
    Syncthing Version: v0.14.43
    Android Version: Android 7.0

nobikik9:
Purpose
As it was mentioned in [#1159](https://github.com/syncthing/syncthing-android/issues/1159) it's better not to have a default camera folder's id as it it, because there is a big risk of collisions. It was also discussed, to randomize somehow that id. The reasonable trade off between fully random id and meaningful is to append same short random string to the meaningful id. I appended 7 random characters preceded by under dash.
Testing
Was tested on Android emulator (Nexus 5, Android 7.1.1)

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
@@ -51,7 +51,7 @@ public class ConfigXml {
     }
 
     private static final String TAG = "ConfigXml";
-    private static final int DEVICE_ID_APPENDIX_LENGTH = 7;
+    private static final int FOLDER_ID_APPENDIX_LENGTH = 4;
 
     private final Context mContext;
     @Inject SharedPreferences mPreferences;
@@ -306,9 +306,9 @@ public class ConfigXml {
                 .replace(" ", "_")
                 .toLowerCase(Locale.US)
                 .replaceAll("[^a-z0-9_-]", "");
-        String deviceId = deviceModel + "_" + generateRandomString(DEVICE_ID_APPENDIX_LENGTH);
+        String defaultFolderId = deviceModel + "_" + generateRandomString(FOLDER_ID_APPENDIX_LENGTH);
         folder.setAttribute("label", mContext.getString(R.string.default_folder_label));
-        folder.setAttribute("id", mContext.getString(R.string.default_folder_id, deviceId));
+        folder.setAttribute("id", mContext.getString(R.string.default_folder_id, defaultFolderId));
         folder.setAttribute("path", Environment
                 .getExternalStoragePublicDirectory(Environment.DIRECTORY_DCIM).getAbsolutePath());
         folder.setAttribute("type", "readonly");
--- a/app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
@@ -26,6 +26,7 @@ import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.Locale;
+import java.util.Random;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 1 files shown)


Catfriend1:
DeviceId = ....
Can you please rename those vars to FolderId?

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -50,6 +51,7 @@
     }
 
     private static final String TAG = "ConfigXml";
+    private static final int DEVICE_ID_APPENDIX_LENGTH = 7;
```
Catfriend1: private static final int FOLDER_ID_APPENDIX_LENGTH = 4;

Name change; Suggest to make it 4 chars long as this only applies for the default camera folder (one single folder created once on first launch) How much chance do we have 4 random char's collide? I think that's okay.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -304,8 +306,9 @@ private boolean changeDefaultFolder() {
                 .replace(" ", "_")
                 .toLowerCase(Locale.US)
                 .replaceAll("[^a-z0-9_-]", "");
+        String deviceId = deviceModel + "_" + generateRandomString(DEVICE_ID_APPENDIX_LENGTH);
```
Catfriend1: String defaultFolderId = deviceModel + "_" + generateRandomString(FOLDER_ID_APPENDIX_LENGTH);

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -304,8 +306,9 @@ private boolean changeDefaultFolder() {
...
         folder.setAttribute("label", mContext.getString(R.string.default_folder_label));
-        folder.setAttribute("id", mContext.getString(R.string.default_folder_id, deviceModel));
+        folder.setAttribute("id", mContext.getString(R.string.default_folder_id, deviceId));
```
Catfriend1: folder.setAttribute("id", mContext.getString(R.string.default_folder_id, defaultFolderId));

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -314,6 +317,19 @@ private boolean changeDefaultFolder() {
         return true;
     }
 
+    /**
+     * Generates a random String with a given length
+     */
+    private String generateRandomString(int length) {
+        char[] chars = "abcdefghjkmnpqrstuvwxyz123456789".toCharArray();
+        Random random = new Random();
```
Catfriend1: Do we have to seed the random generator?

AudriusButkevicius: I think there is some PRNG hack in the app, not sure if it affects Random

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
nobikik9: Yes, probably it's ok to have an appendix 4 chars long. We have around ~1/10^6 chance that two same devices will be given the same id. Moreover, it's an instance of the generalized birthday problem and we need ~450 same devices to have at least one pair of devices with the same id with probability 1/10.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
nobikik9: I've installed and uninstalled Syncthing app several times. It produces different ids.

AudriusButkevicius:
Nvm

Issue #453, opened 2015-08-29
Root


licaon-kter:

Nutomic:
Sorry, forgot to comment:

The problem is that root is only given to the syncthing binary, but the folder listing in the app is done from Java. We could solve that by listing the folders from a shell (pull requests welcome).

Root works fine on my Nexus 5, can you record a logcat on a PC, and post that in a seperate issue?

licaon-kter:
While C/++ maybe I could read, Java is kinda foreign to me.

*"the folder listing in the app is done from Java"*: so it lists the folders in / (root) ok but not subfolders, and that is "as expected" ? I would think that listing /data/subfolder is the same as / (root).

Nutomic:
Listing `/data/` requires root permission, and we can only get that via shell, not via Java APIs. So it needs a different implementation (which we don't have).

For the freeze, it should be fine if you just record a log with adb logcat or a logcat app, and post it here.

gastus1:
Same problem with accessing folders under /data. 
Can configure the folder in WebGUI, but after that the app hangs on white screen, so the sync does not work.
I got an x86 phone (Lenovo K900), so if i could be of any help in resolving the issue, i'm ready!

Nutomic:
The app should start fine if you entered the path correctly. Can you post a logcat of the app start?

gastus1:
I'll be glad to post it, but how? The Logcat app shows me the `/dev/log/main` log.
But i dont see there anything related to syncthing. May be i have to put something in the "Debug options" field of your app?

gastus1:
I'll do it today later or tomorrow!
Thanks!

licaon-kter:
@gastus1 See this post I guess: https://github.com/syncthing/syncthing-android/issues/449#issuecomment-138124151

gastus1:
Here is the logcat from the moment i started the freshly installed app for the 1st time. It includes the key generetion, the selection of root access, entry into the webgui and adding the folder in /data/data/..., and restarting from webgui. After that the app stuckes on white screen, the SuperSU toasters keep on showing saying that the "app given root interactive shell" AND "app given root interactive shell
/data/data/com.nutomic.syncdroidandroid/lib/libsyncthing.so -generate /data/data/com.nutomic.syncdroidandroid/files/", and then the system dialog pops up saying that the app does not respond.



[...]

PR #1170, opened 2018-06-26
Implement proper permission handling using SAF (fixes #1160)

### LINKED ISSUE #1160: Missing .stfolder in default Camera folder
Hello everyone!
Yesterday I bought a new phone. So tried to configure Syncthing to synchronize my default Camera folder (internal storage, it was already created) with a laptop (it's the first time I used Syncthing) . But I encountered a problem with synchronization, because "Folder Marker Missing" error was raised (to see the error I switched to the Web GUI). I found [#168](https://github.com/syncthing/syncthing-android/issues/168) and creating .stfolder folder resolved the issue. But anyway it required some some effort. Is it an intended behavior not to create .stfolder?

I also want to apologize if my question is bad/trivial/stupid (probably, the folder wasn't created because of default r/w permissions, not sure), but it's the first time for me to report a problem.
Thank you!

### Version Information

    App Version: 0.10.11
    Syncthing Version: v0.14.48
    Android Version: Android 7.1
    Phone: Xiaomi Readmi 5A

ghost:
Purpose
Improves the FolderActivity UI when creating or editing a syncthing folder making it clear to the user why a folder can or cannot be written to.
Implements the Storage Access Framework API available on Android 5+ (API level 21) and gracefully falls back to the existing FolderPickerActivity implementation if the app runs on a lower API level.
Intends to fix issue #1160 by creating the ".stfolder" on folder creation properly using the Storage Access Framework on API level 21+. This helps users a lot creating sendonly folders for backup means (yes I know and fully support it is SyncThing and NOT BackupThing).

SAF-related: If a folder can be written to or not by the syncthing binary is intentionally implemented by using the shell instead of SAF as SAF cannot tell us if the native syncthing binary has access to something or not after the user granted permission by making his folder selection on the SAF UI. This also makes the new code backwards compatible as checking write permission using the shell is also available in API levels < 21.

The "Documents" Library shown in Android's native directory picker (SAF UI) is properly supported to be converted to an absolute unix path since commit https://github.com/syncthing/syncthing-android/pull/1170/commits/c786e353ef5e09dc8a7b6cf5352be04f01394365 .

Related Issue
#453 We should close this issue as "invalid" as implementing this would a) risk user data on misusage, b) is a niche use case syncing the /data/ partition from Android and c) will get in our way implementing (or in a future android version maybe HAVING to implement) more SAF use in the code "when Google likes us to do so".

Testing
Verified working on device lg-h815 running Android 7.1.2 at https://github.com/syncthing/syncthing-android/pull/1170/commits/9ee829afaf554eaf527a02a2341210f53fa5d81e .

Verified working on device samsung-i9100 running Android 8.1 https://github.com/syncthing/syncthing-android/pull/1170/commits/9ee829afaf554eaf527a02a2341210f53fa5d81e .


app/src/main/java/com/nutomic/syncthingandroid/util/Util.java
```diff
@@ -122,6 +124,81 @@ public static boolean fixAppDataPermissions(Context context) {
...
+            // This is very unlikely to happen, so we have less error handling.
+            Log.i(TAG, "Failed to remove test file");
+        }
+        return true;
+    }
+
+    /**
+     * Run command in a shell and return the exit code.
+     */
+    public static int runShellCommand(String cmd, Boolean useRoot) {
```
Catfriend1: I'd like to eliminate boilerplate code in a follow-up PR as we have root shell calls spread all over the code and most of them (e.g. niceSyncthing, killSyncthing) can be stripped down to call this new function "runShellCommand".

app/src/main/java/com/nutomic/syncthingandroid/util/FileUtils.java
```diff
@@ -0,0 +1,148 @@
...
+                Boolean isExternalVolume = ((uuid != null) && uuid.equals(volumeId));
+                Boolean isHomeVolume = (uuid == null && HOME_VOLUME_NAME.equals(volumeId));
+                if (isPrimaryVolume || isExternalVolume) {
+                    // Return path if the correct volume corresponding to volumeId was found.
+                    return (String) getPath.invoke(storageVolumeElement);
+                } else if (isHomeVolume) {
+                    // Reading the environment var avoids hard coding the case of the "documents" folder.
+                    return Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOCUMENTS).getAbsolutePath();
+                }
+                Log.d(TAG, "Skipping volume, uuid = '" + uuid + "', volumeId = '" + volumeId + "'");
```
Catfriend1: We should leave this in place. It helped me catch crashes if the user chose from a document provider category like the "home directory" which was not supported - now it's implemented correctly but more foreign apps could hook into this SAF UI place.

app/src/main/java/com/nutomic/syncthingandroid/util/FileUtils.java
```diff
@@ -0,0 +1,148 @@
...
+    private static final String TAG = "FileUtils";
+
+    private static final Boolean isCompatible = (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP);
+
+    private FileUtils() {
+        // Private constructor to enforce Singleton pattern.
+    }
+
+    private static final String PRIMARY_VOLUME_NAME = "primary";
+    private static final String HOME_VOLUME_NAME = "home";
```
Catfriend1: This refers to the "/documents" directory.

app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
```diff
@@ -190,6 +205,39 @@ public void onCreate(Bundle savedInstanceState) {
...
+        externalFilesDir.addAll(Arrays.asList(getExternalFilesDirs(null)));
+        externalFilesDir.remove(getExternalFilesDir(null));
+
+        // Display storage access framework directory picker UI.
+        Intent intent = new Intent(Intent.ACTION_OPEN_DOCUMENT_TREE);
+        if (externalFilesDir.size() > 0) {
+            initialPath = externalFilesDir.get(0);
+            Toast.makeText(this, getString(R.string.folder_hint_writeable_initial_folder, initialPath.getAbsolutePath()),
+                Toast.LENGTH_LONG).show();
+            intent.putExtra("android.provider.extra.INITIAL_URI", Uri.fromFile(initialPath).toString());
```
Catfriend1: This does not work on my Android version. SAF UI starts in the place where the user was the last time the UI was opened. Maybe a future android feature? I only made sure it doesn't crash.

app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
```diff
@@ -190,6 +205,39 @@ public void onCreate(Bundle savedInstanceState) {
...
+        File initialPath = null;
+        ArrayList<File> externalFilesDir = new ArrayList<>();
+        externalFilesDir.addAll(Arrays.asList(getExternalFilesDirs(null)));
+        externalFilesDir.remove(getExternalFilesDir(null));
+
+        // Display storage access framework directory picker UI.
+        Intent intent = new Intent(Intent.ACTION_OPEN_DOCUMENT_TREE);
+        if (externalFilesDir.size() > 0) {
+            initialPath = externalFilesDir.get(0);
+            Toast.makeText(this, getString(R.string.folder_hint_writeable_initial_folder, initialPath.getAbsolutePath()),
```
Catfriend1: I intend to improve this by a future PR.
Reason: With the old FolderPickerActivity now only opening on Android < Lollipop, there was a line pointing to the external storage directory "/storage/....-..../Android/com.nutomic.syncthingandroid/files" to help the user find out where he has write access for doing two-way sync.
I absolutely couldn't figure out doing a lot of research how we can start the SAF folder picker UI with an initial folder to open automatically as I would have chosen to open in this special (long) path at default.

Final solution for a follow-up PR:
We precreate a "Share" folder here initially when the app starts for the first time (and once after app upgrade took place) and then the user as two folders automatically precreated available:
a) "device_name_random_device_model" as Send-Only folder pre-configured to point at "/storage/emulated/0/DCIM" (already implemented since years)
b) "device_name_random_Share" as Send-Receive folder pre-configured to point at "/storage/....-..../Android/com.nutomic.syncthingandroid/files/Share"
Then we can get rid of the nasty toast popup reminder.

Catfriend1:
Screenshots how it will look on Android 5+:
image
image
image
image

app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
Catfriend1: I still need to improve this. Testing on Android 8+ I discovered we need to put a parcelable here.

Catfriend1:
Still ToDo
Put parcelable for the initial folder.
Clear mPathView if an error occured during SAF folder selection (as the var is cleared backstage) Also clear mFolder.path.

Catfriend1:
Ok, its been tested on two devices with success now. See updated first post, section "Testing". I also verified this working correctly if the user has no external sdcard in the phone or took it out after folder creation.
On Android 8+ we use the EXTRA_INITIAL_DIR to start the folder picker in the writeable external storage "/Android/data/<package_name>/files" directory helping the user to create a new sync folder in the right place. Power users (have in the past) and can still go to another location and setup a new sync folder there. To sum up, the .stfolder marker is also created properly regardless from android's native linux process write perms when a new folder is setup by the user.
This will remove the sluggish discussion and tickets about "how and where can I setup sendreceive folders on android".
Ready for review.

capi:
I have not yet reviewed the code itself, I'll try this patch on my devices soon. I just want to say that I think it's a sensible approach to the current situation and should improve usability. "Power-users" that want to do other stuff can still do it from the Web-UI of Syncthing, so you don't remove any use-cases but guide "default users" towards something that is supported. Great job!

Catfriend1:
Doing more testing and comparing to the old ficker mechanism, I found out, we had another problem in 0.10.11 (and before) which is fixed by this PR:
The old FolderPickerActivity (that is still in place for Android < 5) cannot create folders on Android 5+.
Here is a screenshot what happens on 0.10.11 release running Android 8.1:
image

app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
```diff
@@ -381,7 +415,21 @@ public boolean onOptionsItemSelected(MenuItem item) {
                             .show();
                     return true;
                 }
-                getApi().addFolder(mFolder);
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP &&
+                    mFolderUri != null) {
```
AudriusButkevicius: Should we only do this for send only folders?

Catfriend1: Yes, we pre-check if syncthing can write to a folder and only then, the user can move the slider to enable sendreceive mode. If syncthing can write itself, it will create .stfolder correctly. So it is sufficient to do the creation of .stfolder from the java code only for send only folders as you suggested. I'll commit that shortly.

app/src/main/java/com/nutomic/syncthingandroid/service/Constants.java
```diff
@@ -30,6 +30,14 @@
      */
     public static final String PREF_DEBUG_FACILITIES_AVAILABLE  = "debug_facilities_available";
 
+    /**
+     * Available folder types.
+     */
+    public static final String FOLDER_TYPE_SEND_ONLY            = "readonly";
```
AudriusButkevicius: I think its "sendonly"? Or atleast we were planning to rename, yet supported the old names.

Catfriend1: I was unsure if "sendonly" can be set already for 14.48. Let's put the new names when we have 14.49 (or future version) out and the rename took place.

Catfriend1: I checked how it's currently implemented upstream:
v14.48:
- Web UI writes "readonly", "readwrite" (initially, on change)
- syncthing understands "readonly", "readwrite", "sendonly", "sendreceive" in config.xml
v14.49-rc.3:
- Web UI writes "sendonly", "sendreceive" (initially, on change)
- syncthing understands "readonly", "readwrite", "sendonly", "sendreceive" in config.xml
So I would recommend to put "sendreceive" and "sendonly" when we bump syncthing to v14.49 . If we would do it now, the user can jump into a pitfall when we create a folder "sendonly" and he edits it later through the web UI to be "readwrite". Our code would - according to the constants in place - compare "readwrite" to the string "sendreceive" and assume we were "readonly" then. This can be solved by changing the constans later after 14.49 got released.

Catfriend1:
@AudriusButkevicius Anything else I can do for this PR or are we ready now?

app/src/main/java/com/nutomic/syncthingandroid/service/Constants.java
AudriusButkevicius: So I'll only release the new version of the app once .49 rolls off the press.

Catfriend1: Ok, I prepared this PR for syncthing v0.14.49+, see commit https://github.com/syncthing/syncthing-android/pull/1170/commits/463d9cc1d3f1f139b19aa49a522cb0acce23c1c9

Catfriend1:
Ok let's wait what @capi says after his test. Please note that a test should be based on the commit before the constants change as we still build with 14.48 here.

AudriusButkevicius:
@capi, can you give the all clear, or tell us if you don't have time and merge it as is?

Catfriend1:
@capi If you don't have time I'll do another test round on my second phone the next days.

capi:
I'm so sorry, my job is a little "crazy" at the moment, I hardly have time to do anything besides it. Haven't been online in quite some time. If you guys already tested it, please merge as is, it's unlikely I'll have time for anything until about 2 weeks from now. Sorry again for not commenting that sooner.

Catfriend1:
SideNote - If you are testing this code on top of the current syncthing v0.14.48 syncthing, you'll experience expected behaviour where the "sendonly" slider is not saved and restored correctly. This is because we prepared the code for the new folder types in syncthing v0.14.49+ and therefore not considered a bug.

spaetz:
Juszt testing this in syncthing-android 0.10.13 (syncthing 0.14.49dirty): I can create and select a folder using the SAF framework just fine. However, I am still getting "permission denied" errors in /storage/1234-5678/DCIM.
Storage permission has been granted to the app.

Catfriend1:
It is working correctly but Audrius broke my code by a workaround in the hope to fix another bug so you'll have to wait for the next release. The workaround makes the gui show you a wrong permission text.
His change got reverted in the meantime and the bug fixed properly, this build has the changes in it: https://build.syncthing.net/repository/download/SyncthingAndroid_Build/24761:id/outputs/apk/debug/app-debug.apk

Catfriend1:
@spaetz Please head to the forum and find an unofficial beta at https://forum.syncthing.net/t/syncthing-android-unofficial-beta/11920 . I'd appreciate your feedback on this if it solves your issue.

PR #1176, opened 2018-07-03
Add welcome wizard slides on first start requesting permissions

ghost:
Purpose
Improve the app by adding a welcome wizard explaining required and optional permissions.

Screenshots
image
image
image
image

Testing
Verified working on device lg-h815 running Android 7.1.2 at commit https://github.com/syncthing/syncthing-android/pull/1176/commits/fd490d8a1f87cc08bddfc6129c857b0047ff569a .

EDIT: Removed the relation to issue #1129 . I'll do that fix based on this PR in a follow-up so this doesn't get too big.


AudriusButkevicius:
Its all in german so I have no idea what the guide says.

Catfriend1:
@AudriusButkevicius see strings.xml for the English texts. The german parts got loaded because I reused the texts that syncthing already had for german since many releases.

Catfriend1:
@AudriusButkevicius The code for the slides requesting permissions from the user is not huge, so I decided to add it to FirstStartActivity (which was nearly "empty" before). It's all in one place now :).

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
```diff
@@ -1,83 +1,311 @@
...
+            container.removeView(view);
+        }
+    }
+
+    /**
+     * Preconditions:
+     * Storage permission has been granted.
+     */
+    private void startApp() {
+        Boolean doInitialKeyGeneration = !Constants.getConfigFile(this).exists();
```
Catfriend1: Sidenote: I've replaced the "isFirstStart" check (previously done via preference check) because it's not very reliable detecting if syncthing needs to generate new keys and config. We now use the same indicator we already have in other classes (e.g. ConfigXml). If no config is present (or it got corrupted in a way crucial files are missing) we'll trigger the "key and config generation" process on app start. This gracefully auto-fixes an unrecoverable config break so we don't have to advise users to clear app's data and cache.

app/src/main/res/values/strings.xml
```diff
@@ -11,16 +11,29 @@
...
+    <!-- Slide 1 -->
+    <string name="introduction">Introduction</string>
     <string name="welcome_text">Syncthing is an open-source file synchronization application.\n\
 To share data with other devices, you need to add their unique device IDs to the device list. Afterwards you can select which folders to share with which devices.\n\
 Please report any problems you encounter via Github.</string>
 
-    <string name="cont">Continue</string>
+    <!-- Slide 2 -->
+    <string name="storage_permission_title">Storage permission</string>
+    <string name="storage_permission_desc">Syncthing requires the storage permission to do file synchronization. We ask you to grant the permission.</string>
```
imsodin: As this is a required permission, I wouldn't ask and have a button - just explain and then request the permission on pressing "next".

Alternative wording that doesn't just repeat the title and seems "clearer" and easier to me:  
`Syncthing needs to access your storage to do file synchronization.`

app/src/main/res/values/strings.xml
```diff
@@ -11,16 +11,29 @@
...
+    <!-- Welcome wizard -->
+    <!-- Slide 1 -->
+    <string name="introduction">Introduction</string>
     <string name="welcome_text">Syncthing is an open-source file synchronization application.\n\
 To share data with other devices, you need to add their unique device IDs to the device list. Afterwards you can select which folders to share with which devices.\n\
 Please report any problems you encounter via Github.</string>
 
-    <string name="cont">Continue</string>
+    <!-- Slide 2 -->
+    <string name="storage_permission_title">Storage permission</string>
```
imsodin: Potentially capitalize fully, as it's a title.

app/src/main/res/values/strings.xml
```diff
@@ -11,16 +11,29 @@
...
+    <string name="storage_permission_desc">Syncthing requires the storage permission to do file synchronization. We ask you to grant the permission.</string>
+
+    <!-- Slide 3 -->
+    <string name="location_permission_title">Location permission</string>
```
imsodin: Also maybe capitalize.

app/src/main/res/values/strings.xml
```diff
@@ -11,16 +11,29 @@
...
+    <string name="location_permission_desc">Syncthing can be configured to run only on selected wifi networks. Due to android requirements, you need to grant the permission if you\'d like to use this feature. Otherwise, feel free to deny it.</string>
```
imsodin: Here the button makes sense, as it's voluntary. I'd turn the latter part around to focus on that choice instead of the android requirement. Something like `If you want to use this feature, press the button above to give the required location access to Syncthing. Otherwise you can skip this step.`

imsodin (commented):
If you have high-level stuff that you think would benefit from another set of eyes, you can ping me. I am not the ideal candidate for it, as I am not very invested in advance functionalities of the app (I am happy as long as it starts and stops when I manually tell it to). I don't follow the syncthing-android activities closely and have no plans to dig into it, but as far as I have time and something to say I am happy to provide feedback. Just to have said it explicitly: Consider all of my feedback as suggestions - I don't have any stakes in the android wrapper.

I believe this is a good improvement, it's always nice to be up front about required permissions. Are these pages conditional, i.e. only show up if the permissions weren't already granted?

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
```diff
@@ -1,83 +1,311 @@
...
 
 import javax.inject.Inject;
 
-public class FirstStartActivity extends Activity implements Button.OnClickListener {
+public class FirstStartActivity extends Activity {
 
     private static String TAG = "FirstStartActivity";
+    private static final int REQUEST_COARSE_LOCATION = 141;
     private static final int REQUEST_WRITE_STORAGE = 142;
+    private static final int SLIDE_POS_LOCATION_PERMISSION = 1;
```
AudriusButkevicius: What are these magic numbers?

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
```diff
@@ -1,83 +1,311 @@
...
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         ((SyncthingApp) getApplication()).component().inject(this);
 
-        if (!isFirstStart()) {
+        /**
+         * Recheck storage permission. If it has been revoked after the user
+         * completed the welcome slides, displays the slides again.
+         */
```
AudriusButkevicius: Are we going to show all slides if the permission is revoked?

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
```diff
@@ -1,83 +1,311 @@
...
     private static final int REQUEST_WRITE_STORAGE = 142;
+    private static final int SLIDE_POS_LOCATION_PERMISSION = 1;
+
+    private ViewPager mViewPager;
+    private ViewPagerAdapter mViewPagerAdapter;
+    private LinearLayout dotsLayout;
+    private TextView[] dots;
+    private int[] layouts;
+    private Button btnBack;
+    private Button btnNext;
```
AudriusButkevicius: Why no m prefix for these?

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
```diff
@@ -1,83 +1,311 @@
...
-             */
-            return;
+    public void onBtnNextClick() {
+        // Check if we are allowed to advance to the next slide.
+        if (mViewPager.getCurrentItem() == SLIDE_POS_LOCATION_PERMISSION) {
+            // As the storage permission is a prerequisite to run syncthing, refuse to continue without it.
+            if (!haveStoragePermission()) {
+                Toast.makeText(this, R.string.toast_write_storage_permission_required,
+                        Toast.LENGTH_LONG).show();
+                this.finish();
```
AudriusButkevicius: What does finish do in this case?

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
```diff
@@ -1,83 +1,311 @@
...
+
+        if (dots.length > 0)
+            dots[currentPage].setTextColor(colorsActive[currentPage]);
+    }
+
+    private int getItem(int i) {
+        return mViewPager.getCurrentItem() + i;
+    }
+
+    //  viewpager change listener
```
AudriusButkevicius: capitalisation

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
```diff
@@ -1,83 +1,311 @@
...
+            /* Slide: storage permission */
+            Button btnGrantStoragePerm = (Button) view.findViewById(R.id.btnGrantStoragePerm);
+            if (btnGrantStoragePerm != null) {
+                btnGrantStoragePerm.setOnClickListener(new View.OnClickListener() {
+                    @Override
+                    public void onClick(View v) {
+                        requestStoragePermission();
+                    }
+                });
+            }
```
AudriusButkevicius: If it's null, what should we do in that case?

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
```diff
@@ -1,83 +1,311 @@
...
+
+        @Override
+        public int getCount() {
+            return layouts.length;
+        }
+
+        @Override
+        public boolean isViewFromObject(View view, Object obj) {
+            return view == obj;
+        }
```
AudriusButkevicius: Why do we need to override this?

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
```diff
@@ -1,83 +1,311 @@
...
+
+
+        @Override
+        public void destroyItem(ViewGroup container, int position, Object object) {
+            View view = (View) object;
+            container.removeView(view);
```
AudriusButkevicius: Same "why"

app/src/main/res/values/colors.xml
```diff
@@ -10,4 +10,38 @@
     <color name="text_blue">#ff33b5e5</color>
     <color name="text_green">#ff99cc00</color>
     <color name="light_grey">#cccccc</color>
+
+    <!-- FirstStartActivity welcome wizard start -->
+    <!-- Screens background color-->
+    <color name="bg_screen1">#3395ff</color>
+    <color name="bg_screen2">#20d2bb</color>
+    <color name="bg_screen3">#c873f4</color>
+    <!-- <color name="bg_screen4">#f64c73</color> -->
```
AudriusButkevicius: Is there harm in having these uncommented if there are only 3 layouts?

app/src/main/res/values/strings.xml
AudriusButkevicius: I think this summons a separate system dialog, hence has to be "initiated" somehow.

Agree with the better wording.

app/src/main/res/values/strings.xml
AudriusButkevicius: I think this needs an explanation why...


> Syncthing can be configured to run only on selected wifi networks. Sadly, Android requires applications to have location permissions to be able look up active WiFi network name, as you can sometimes infer users location from the name of the network they are connected to. If you want to use this feature, press the button above to give the required location permissions to Syncthing. Otherwise you can skip this step.

Catfriend1: Hmm.. both suggestions are good. imsodin's is comforting the user but the user doesn't really get why a location permission should be granted for a wifi feature. Audrius' suggestion tells the truth (what for example complies with fdroid policy for app approval) but also looks a tiny bit scary as we say what nasty things we could do (but won't do :). For the first try on "later" public feedback I'll take Audrius' suggestion.

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
Catfriend1: This is an exception in the wizard as we have to make sure the app quits if the user didn't grant storage permissions during the second slide and tell him when exiting why we do it (that was already in the original welcome screen). So it is a slide index to make it easier if we need to rearrange slides later. I chose this way as the ViewPager doesn't really help here returning something unique and easily available to get the xml_layout_slide_name.

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
Catfriend1: Yes, I decided to do so in the first proposal as I consider it rare that users revoke the permission if they actively use the app. (see rare one bug report on the tracker where one user did). Reason is it will never harm anything going through the slides again - also being able to grant the location permission if this was also revoked "later" after completing the wizard for the first time. Cluttering the code to handle rare cases is a maybe can do but currently don't see necessary.

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
Catfriend1: I've had them temporarily in another context, added the "m" and gave the buttons a new var name.

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
Catfriend1: I exits the app by ending the FirstStartActivity like it was the case before the PR if the user did not grant the required storage permission.

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
Catfriend1: ok

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
Catfriend1: We do nothing as this is intended behaviour. The button exists only on one slide (slide 2) so when for example slide 3 gets instantiated it WILL return null as it doesn't exist. In this case we don't register the onClickListener as we don't need it.

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
Catfriend1: The method  instantiateItem(ViewGroup, int) returns an Object for a particular view. PagerAdapter implementation is considering this Object as a key value when viewpager changes a page. If we return the view itself from instantiateItem(ViewGroup, int), then our key for that page becomes the view itself. We can check return view == obj; from isViewFromObject which will always return true and our pages will display.

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
Catfriend1: Proper uninitialization and the above mentioned reason.

app/src/main/res/values/colors.xml
Catfriend1: Uncommenting will throw a lint warning but not be a problem. I'll reserved the color as I already know what the fourth slide will be.

AudriusButkevicius:
Other than the next button exiting if storage permission is not granted, looks good to me.

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
Catfriend1: Android development guidelines for permission request say we're not allowed to ask this multiple times on our own. That's why the button makes sense.

Catfriend1:
@AudriusButkevicius Okay, then let's go. As we currently have many discussion I'd suggest a follow-up PR would be great to meet again making the "app exits when storage permission not granted" better.

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
AudriusButkevicius: We should just disable the next button if the user has not granted permissions.

Catfriend1: A message is already printed on the screen when exiting, so it's no crash.
Disabling the next button would leave no way to exit the foreground app window. User would then have to press home or the panic button.

AudriusButkevicius: Why can't he press the back button?

AudriusButkevicius: Or we could just make clicking next print the toast and not progress to the new window. Then the user can back out, or grant the permission.

Catfriend1: I'll think about that. Good for follow-up.

AudriusButkevicius (dismissed):

AudriusButkevicius:
Anything left todo, or are we good to merge this?

Catfriend1:
Ok from my side. (I'll already mentioned I'd like to take care of more welcome slides, especially those required to make the user experience smooter on first app start relating to key generation and run conditions).

@imsodin: Is is okay from your side? Some of your feedback already got in.

app/src/main/res/values/strings.xml
imsodin: Fine with me, but please drop the sadly. We want the user to be cheerful about using Syncthing ;)

imsodin (changes requested):
I left my notes on an outdated review, meaning they are hidden. Thus to repeat here:

The location permission explanation is fine with me, but please drop the sadly. We want the user to be cheerful about using Syncthing ;)

And I didn't mean to put the strings in all caps, but to capitalize them, i.e. make the first letter uppercase: Storage permissions -> Storage Permissions (again, that's just how I am used to do things for title-like properties, maybe mixed capitalization looks better. I don't care much about it, except that it is **not** all caps :)

Catfriend1:
Thanks for reviewing, @imsodin I got all comments via mail that were hidden and will take care later.

Catfriend1:
@AudriusButkevicius Would you mind leaving the merges up to me after a PR has got the required approvements? Then I would plan to do a final testing after the reviews and before hitting the merge.

AudriusButkevicius:
Sure, but everytime you merge something in the branch the approval gets dismissed.

AudriusButkevicius (approved):

imsodin (approved):

Catfriend1:
Ok, testing completed, no crashes or problems so far found.

PR #1177, opened 2018-07-06
Fix recurring device, folder accept notifications, add ignore action (fixes #679)

### LINKED ISSUE #679: Same share folder notification displayed multiple times

ghost:
Purpose
Fix recurring device, folder consent notifications if the user dismissed the notification or swiped it away.

Implementation
Adds the device name instead of the device id for better readability in the "device xxx wants to connect" notification. Falls back to displaying the device id if the device name isn't available (like it was before).
Notification id generation logic for device, folder accepts is moved to NotificationHandler class.
Displays two action buttons on API level 16+ below the notification labeled "accept" and "ignore". (We already have API level 16 so this would be no compatibility problem.)
Implemented the correct "ignoreFolder" and "ignoreDevice" action handling code in RestApi.java.

Related issue
https://github.com/syncthing/syncthing-android/issues/679

Testing
Verified working on Android 7.1.2, device lg-h815 at commit https://github.com/syncthing/syncthing-android/pull/1177/commits/0e36400f435ea0b0e1a55f2215723859599ea6b7 .

Screenshots
image
image

image

Ignored device:
image

New implementation of event-unique notification IDs which are passed on to the intent or service and later correctly used to dismiss a notification when it was handled ("answered") by the user.
1) Shared folder request: 2) Device connection request:
image
image

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
@@ -227,14 +227,16 @@ public class EventProcessor implements  Runnable, RestApi.OnReceiveEventListener
                 .putExtra(DeviceActivity.EXTRA_DEVICE_NAME, deviceName);
         // Use a deterministic ID to make multiple PendingIntents distinguishable.
         int requestCode = deviceId.hashCode();
-        PendingIntent piAccept = PendingIntent.getActivity(mContext, requestCode, intentAccept, 0);
+        PendingIntent piAccept = PendingIntent.getActivity(mContext, requestCode,
+            intentAccept, PendingIntent.FLAG_UPDATE_CURRENT);
 
         // Prepare "ignore" action.
         Intent intentIgnore = new Intent(mContext, SyncthingService.class)
                 .putExtra(SyncthingService.EXTRA_NOTIFICATION_ID, notificationId)
                 .putExtra(SyncthingService.EXTRA_DEVICE_ID, deviceId);
         intentIgnore.setAction(SyncthingService.ACTION_IGNORE_DEVICE);
-        PendingIntent piIgnore = PendingIntent.getService(mContext, 0, intentIgnore, 0);
+        PendingIntent piIgnore = PendingIntent.getService(mContext, 0,
+            intentIgnore, PendingIntent.FLAG_UPDATE_CURRENT);
 
         // Show notification.
         mNotificationHandler.showConsentNotification(notificationId, title, piAccept, piIgnore);  // ToDo
@@ -269,14 +271,16 @@ public class EventProcessor implements  Runnable, RestApi.OnReceiveEventListener
                 .putExtra(FolderActivity.EXTRA_FOLDER_LABEL, folderLabel);
         // Use a deterministic ID to make multiple PendingIntents distinguishable.
         int requestCode = (deviceId + folderId).hashCode();
-        PendingIntent piAccept = PendingIntent.getActivity(mContext, requestCode, intentAccept, 0);
+        PendingIntent piAccept = PendingIntent.getActivity(mContext, requestCode,
+            intentAccept, PendingIntent.FLAG_UPDATE_CURRENT);
 
         // Prepare "ignore" action.
         Intent intentIgnore = new Intent(mContext, SyncthingService.class)
                 .putExtra(SyncthingService.EXTRA_NOTIFICATION_ID, notificationId)
                 .putExtra(SyncthingService.EXTRA_FOLDER_ID, folderId);
         intentIgnore.setAction(SyncthingService.ACTION_IGNORE_FOLDER);
-        PendingIntent piIgnore = PendingIntent.getService(mContext, 0, intentIgnore, 0);
+        PendingIntent piIgnore = PendingIntent.getService(mContext, 0,
+            intentIgnore, PendingIntent.FLAG_UPDATE_CURRENT);
 
         // Show notification.
         mNotificationHandler.showConsentNotification(notificationId, title, piAccept, piIgnore);
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
@@ -94,12 +94,6 @@ public class EventProcessor implements  Runnable, RestApi.OnReceiveEventListener
      */
     @Override
     public void onEvent(Event event) {
-        String deviceId;
-        String deviceName;
-        String folderId;
-        String title;
-        PendingIntent pi;
-
         switch (event.type) {
             case "ConfigSaved":
                 if (mApi != null) {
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
```diff
@@ -242,10 +209,80 @@ public void stop() {
...
+        // Prepare "ignore" action.
+        Intent intentIgnore = new Intent(mContext, SyncthingService.class)
+                .putExtra(SyncthingService.EXTRA_NOTIFICATION_ID, notificationId)
+                .putExtra(SyncthingService.EXTRA_DEVICE_ID, deviceId);
+        intentIgnore.setAction(SyncthingService.ACTION_IGNORE_DEVICE);
+        PendingIntent piIgnore = PendingIntent.getService(mContext, 0,
+            intentIgnore, PendingIntent.FLAG_UPDATE_CURRENT);
+
+        // Show notification.
+        mNotificationHandler.showConsentNotification(notificationId, title, piAccept, piIgnore);  // ToDo
```
AudriusButkevicius: Todo what?

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
```diff
@@ -242,10 +209,80 @@ public void stop() {
...
+        PendingIntent piAccept = PendingIntent.getActivity(mContext, requestCode,
+            intentAccept, PendingIntent.FLAG_UPDATE_CURRENT);
+
+        // Prepare "ignore" action.
+        Intent intentIgnore = new Intent(mContext, SyncthingService.class)
+                .putExtra(SyncthingService.EXTRA_NOTIFICATION_ID, notificationId)
+                .putExtra(SyncthingService.EXTRA_DEVICE_ID, deviceId);
+        intentIgnore.setAction(SyncthingService.ACTION_IGNORE_DEVICE);
+        PendingIntent piIgnore = PendingIntent.getService(mContext, 0,
+            intentIgnore, PendingIntent.FLAG_UPDATE_CURRENT);
```
AudriusButkevicius: How does this work, why do we have 3 intents, two pending, one non-pending etc? What's the difference here? I know nothing about android.

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
```diff
@@ -242,10 +209,80 @@ public void stop() {
...
+            return;
+        }
+        Log.d(TAG, "Device " + deviceId + " wants to share folder " +
+            folderLabel + " (" + folderId + ")");
+
+        // Find the deviceName corresponding to the deviceId
+        String deviceName = null;
+        for (Device d : mApi.getDevices(false)) {
+            if (d.deviceID.equals(deviceId))
+                deviceName = d.getDisplayName();
```
AudriusButkevicius: break?

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
Catfriend1: It's okay like this because the intents are nothing that already "lives" at that time. You can see them as "prepared" for later use. That's why the two intents (accept, ignore) are put into a pending intent for each (accept, ignore) and then passed on to our NotificationHandler class (passing it on to android's notification service). Android docs say, that a PendingIntent in a notification is a set of "what should be executed" (e.g. activity or service action like it is here) and which the target object needs to know (that's why we put extras in every intent). Android also has a built-in mechanism to take care, intents that represent the same (told by the requestCode) are only kept one-time in memory and are internally referenced. As you asked in your other question, the requestCode should stay like this (compare to the "old code" that used the same approach for producing the same requestCode when the folder/device combination we talk about is the same per triggered event). I only added the PendingIntent.FLAG_UPDATE_CURRENT as the extra's are not carried along to the service/activity launched when hitting the action button instead of clicking in the middle of the notification field itself. That's specific to the action button interface of android's notification service since API level 16+ and does not harm our compatibility level.

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
Catfriend1: It's already done and I forgot to remove the comment. Sorry.

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
```diff
@@ -242,10 +209,80 @@ public void stop() {
...
-        mNotificationHandler.showEventNotification(text, pi, notificationId);
+    private void onDeviceRejected(String deviceId, String deviceName) {
+        if (deviceId == null) {
+            return;
+        }
+        Log.d(TAG, "Unknown device " + deviceName + "(" + deviceId + ") wants to connect");
+
+        String title = mContext.getString(R.string.device_rejected,
+                deviceName.isEmpty() ? deviceId.substring(0, 7) : deviceName);
+        int notificationId = mNotificationHandler.getNotificationIdFromText(title);
```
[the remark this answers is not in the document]
Catfriend1: The ID should be in a deterministic range, "Nutomic" chose between 1000 and 2000. We could do 1000 + deviceId.hashCode % 1000 but that makes the code harder to be consistent if we have another deviceId related notification in the future. I suggest sticking to like it already was before this PR where the whole notification is hashed and an ID betweeen 1000 and 2000 is generated. We will then generate the same IDs for the same notification texts when asking the same question to a user. Keeping the ID generation consistently done one way in one place (getNotificationIdFromText) sounds best to me.

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
Catfriend1: Agreed. I've overseen this taking over the already existing code.

Catfriend1:
Ok, I fixed the nitpicks :). Agree with me on my other comments?

app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
```diff
@@ -242,10 +209,82 @@ public void stop() {
...
+
+        // Prepare "accept" action.
+        Intent intentAccept = new Intent(mContext, DeviceActivity.class)
+                .putExtra(DeviceActivity.EXTRA_NOTIFICATION_ID, notificationId)
+                .putExtra(DeviceActivity.EXTRA_IS_CREATE, true)
+                .putExtra(DeviceActivity.EXTRA_DEVICE_ID, deviceId)
+                .putExtra(DeviceActivity.EXTRA_DEVICE_NAME, deviceName);
+        // Use a deterministic ID to make multiple PendingIntents distinguishable.
+        int requestCode = deviceId.hashCode();
```
AudriusButkevicius: Why is this not the same as notificationId then?

Catfriend1: It was taken over from the previous code. I see no point why my comment from above would not apply here so I'll make a commit to use the notificationId (already generated above) as the requestCode. Values between 1000 and 2000 are okay here.

Catfriend1:
I verified again that everything works correctly at commit https://github.com/syncthing/syncthing-android/pull/1177/commits/0e36400f435ea0b0e1a55f2215723859599ea6b7 .
Also multiple concurrent folder share requests are correctly handled by the "unique per folder/device combo" notifications with their specific IDs.

Screenshot:
image

PR #1180, opened 2018-07-16
Remove shell boilerplate code (fixes #1181)

### LINKED ISSUE #1181: Android denies running the "ionice" command in non-root environments
Problem
Android OS denies running the "ionice" command from shell returing exit code "127" "permission denied" unless we use a root shell.

Reproducer
Use syncthing app **without** the root option checked > Log says "ionice returned with exit code 127". Temporarily attaching the log()-dumper for this subshell process reveals a "permission denied" error message.

Use syncthing app **with** the root option checked > Log says "ionice returned with exit code 0". Temporarily attaching the log()-dumper for this subshell process just doesn't bring up an error message. Exit code 0 means "successfully io'niced process".

### Version Information
    App Version: 0.10.11 (and earlier versions)
    Syncthing Version: v0.14.48 (not relevant)
    Android Version: Android 7.1.2

ghost:
Purpose
Remove shell boilerplate code by replacing relevant getRuntime().exec parts with Util.runShellCommand.

Minor bugfix
Fix issue #1181: Android OS denies running the "ionice" command from shell returing exit code "127" "permission denied" unless we use a root shell. The fix will run "ionice" only when the root option is enabled and superuser shell is available.

Minor changes
Correct log texts and levels.

Testing
Verified working at commit https://github.com/syncthing/syncthing-android/pull/1180/commits/57bc16770253a6b813f6106d15a6c67f8594cfdb on Android 7.1.2, device lg-h815.

Log checks done during testing
1) Verified the "Util.fixAppDataPermission" function to work properly when switching to root mode and back in the options and restarting the whole app after each step. The shell commands "chown" and "restorecon" executed correctly.

2) graceful syncthing process kill on exiting the syncthing app
image
To verify, filter the log on "SyncthingUtil" and "runShellCommand".

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
@@ -6,6 +6,7 @@ import android.content.Intent;
 import android.content.SharedPreferences;
 import android.os.Environment;
 import android.os.PowerManager;
+import android.os.SystemClock;
 import android.text.TextUtils;
 import android.util.Log;
 
@@ -299,15 +300,15 @@ public class SyncthingRunnable implements Runnable {
             }
 
             int exitCode;
-            String shellCmd;
             for (String syncthingPID : syncthingPIDs) {
                 if (i > 0) {
                     // Force termination of the process by sending SIGKILL.
-                    shellCmd = "sleep 3; kill -SIGKILL " + syncthingPID + "\n";
+                    SystemClock.sleep(3000);
+                    exitCode = Util.runShellCommand("kill -SIGKILL " + syncthingPID + "\n", mUseRoot);
                 } else {
-                    shellCmd = "kill -SIGINT " + syncthingPID + "; sleep 1\n";
+                    exitCode = Util.runShellCommand("kill -SIGINT " + syncthingPID + "\n", mUseRoot);
+                    SystemClock.sleep(1000);
                 }
-                exitCode = Util.runShellCommand(shellCmd, mUseRoot);
                 if (exitCode == 0) {
                     Log.d(TAG, "Killed Syncthing process " + syncthingPID);
                 } else {
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
@@ -240,7 +240,7 @@ public class SyncthingRunnable implements Runnable {
                 }
             }
         } catch (IOException | InterruptedException e) {
-            Log.w(TAG, "Failed list Syncthing processes", e);
+            Log.w(TAG, "Failed to list Syncthing processes", e);
         } finally {
             try {
                 if (br != null) {
@@ -250,7 +250,7 @@ public class SyncthingRunnable implements Runnable {
                     psOut.close();
                 }
             } catch (IOException e) {
-                Log.w(TAG, "Failed close the psOut stream", e);
+                Log.w(TAG, "Failed to close psOut stream", e);
             }
             if (ps != null) {
                 ps.destroy();
@@ -270,7 +270,7 @@ public class SyncthingRunnable implements Runnable {
 
         List<String> syncthingPIDs = getSyncthingPIDs();
         if (syncthingPIDs.isEmpty()) {
-            Log.w(TAG_NICE, "Found no running instances of " + Constants.FILENAME_SYNCTHING_BINARY);
+            Log.i(TAG_NICE, "Found no running instances of " + Constants.FILENAME_SYNCTHING_BINARY);
             return;
         }
 
@@ -311,7 +311,7 @@ public class SyncthingRunnable implements Runnable {
                 if (exitCode == 0) {
                     Log.d(TAG, "Killed Syncthing process " + syncthingPID);
                 } else {
-                    Log.d(TAG, "Failed to kill Syncthing process " + syncthingPID +
+                    Log.w(TAG, "Failed to kill Syncthing process " + syncthingPID +
                         " exit code " + Integer.toString(exitCode));
                 }
             }
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
@@ -263,6 +263,11 @@ public class SyncthingRunnable implements Runnable {
      * Look for a running libsyncthing.so process and nice its IO.
      */
     private void niceSyncthing() {
+        if (!mUseRoot || !Shell.SU.available()) {
+            Log.i(TAG_NICE, "Root is not available. Cannot nice syncthing.");
+            return;
+        }
+
         List<String> syncthingPIDs = getSyncthingPIDs();
         if (syncthingPIDs.isEmpty()) {
             Log.w(TAG_NICE, "Found no running instances of " + Constants.FILENAME_SYNCTHING_BINARY);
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
```diff
@@ -317,44 +298,22 @@ public void killSyncthing() {
...
-            } catch (IOException e) {
-                Log.w(TAG_KILL, "Failed close the killOut stream", e);}
-            if (kill != null) {
-                kill.destroy();
+                if (i > 0) {
+                    // Force termination of the process by sending SIGKILL.
+                    shellCmd = "sleep 3; kill -SIGKILL " + syncthingPID + "\n";
+                } else {
+                    shellCmd = "kill -SIGINT " + syncthingPID + "; sleep 1\n";
+                }
```
AudriusButkevicius: You should do the sleeping in java.

Also https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/os/Process.java#904

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
```diff
@@ -317,44 +298,22 @@ public void killSyncthing() {
...
+                exitCode = Util.runShellCommand(shellCmd, mUseRoot);
+                if (exitCode == 0) {
```
AudriusButkevicius: If the process no longer exists, it will not be 0

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
AudriusButkevicius: Also, sleep 1 will always return 0.

app/src/main/java/com/nutomic/syncthingandroid/util/Util.java
```diff
@@ -89,39 +89,41 @@ public static boolean fixAppDataPermissions(Context context) {
...
+        Log.d(TAG, "Running: '" + cmd);
+        int exitCode = runShellCommand(cmd, true);
+        if (exitCode == 0) {
+            Log.i(TAG, "Fixed app data permissions on '" + dir + "'.");
+        } else {
+            Log.w(TAG, "Failed to fix app data permissions on '" + dir + "'. Result: " +
+                Integer.toString(exitCode));
         }
-        return false;
+        return (exitCode == 0);
```
AudriusButkevicius: No need for the braces?

app/src/main/java/com/nutomic/syncthingandroid/util/Util.java
```diff
@@ -89,39 +89,41 @@ public static boolean fixAppDataPermissions(Context context) {
...
             // One should always be able to retrieve the application info for its own package.
-            Log.w(TAG, "This should not happen", e);
+            Log.w(TAG, "Error getting current package name", e);
+            return false;
+        }
+        Log.d(TAG, "Uid of '" + packageName + "' is " + appInfo.uid);
+
+        // Get private app's "files" dir residing in "/data/data/[packageName]".
+        String dir = context.getFilesDir().getAbsolutePath();
+        String cmd = "chown -R " + appInfo.uid + ":" + appInfo.uid + " " + dir + "; ";
```
AudriusButkevicius: This bit could probably be implemented in java.

app/src/main/java/com/nutomic/syncthingandroid/util/Util.java
```diff
@@ -89,39 +89,41 @@ public static boolean fixAppDataPermissions(Context context) {
...
+        // Running Syncthing as root might change a file's or directories type in terms of SELinux.
+        // Leaving them as they are, the Android service won't be able to access them.
+        // At least for those files residing in an application's data folder.
+        // Simply reverting the type to its default should do the trick.
+        cmd += "restorecon -R " + dir + "\n";
```
AudriusButkevicius: Seems it can be done via https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/os/SELinux.java

AudriusButkevicius (changes requested):

app/src/main/java/com/nutomic/syncthingandroid/util/Util.java
Catfriend1: Note for me: https://gitlab.tubit.tu-berlin.de/justus.beyer/streamagame_platform_frameworks_base/commit/dbea3cd118fe4eeb9b43ea50727987fa790a2edd

Catfriend1: Sounds nice, but no. I am not able to do this because I didn't find a way on my research. [ import android.os.SELinux; ] is not available except to framework components which our app isn't. Furthermore, if the import was available we would have to break compatibility to android 4.x. I can't recommend this way.

app/src/main/java/com/nutomic/syncthingandroid/util/Util.java
Catfriend1: This the "root part" of our app. We cannot do this from java and therefore need the root shell.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
Catfriend1: I've implemented sleep in java.

app/src/main/java/com/nutomic/syncthingandroid/util/Util.java
Catfriend1: Corrected.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
Catfriend1: The process stuff needs to be consistent with no-root/root execution level. In my opinion, the way we did it for the last years in code fits best, as we have the same commands running in a non-root as in the root shell. Else, we would have two code parts (even if we have no bug related to that at current time of writing) one for java and one for the root shell. Maybe let's do the Process.Kill from java in a later PR.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
Catfriend1: yes, according to my terminal emulator testing and source ( http://man7.org/linux/man-pages/man1/kill.1.html ) it will return exit code 1 then. It's called "failure" in the docs, so this code part is right logging "Failed to kill syncthing process ### exit code 1" then.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingRunnable.java
```diff
@@ -317,44 +299,22 @@ public void killSyncthing() {
...
-            try {
-                if (killOut != null)
-                    killOut.close();
-            } catch (IOException e) {
-                Log.w(TAG_KILL, "Failed close the killOut stream", e);}
-            if (kill != null) {
-                kill.destroy();
+                if (i > 0) {
+                    // Force termination of the process by sending SIGKILL.
+                    SystemClock.sleep(3000);
```
AudriusButkevicius: Why do we sleep here again?

Catfriend1: I just took over the original code, so honestly, I don't know exactly. I suspect this was put in to give the process a plus of 3 secs if previous kill attempts failed to end it before force-killing it.  One can argue if this a higher or lower risk of data corruption if something was stuck during the "normal" kill. Looking at the test logs, I always get syncthing normally killed and not force-killed. My aim by this PR was to remove duplicate code for handling the root shells returning exit codes so I intentionally didn't alter existing code more than necessary.

AudriusButkevicius (approved):

PR #1184, opened 2018-07-17
Move build script to python

AudriusButkevicius:


calmh:
It's supposed to auto build on push, and the CI server is supposed to auto pull. But the Docker hub build config was set to build the `docker` branch, I fixed that to master now and triggered a rebuild.

If you have a docker hub account let me know and I'll add to the org, I didn't find you under the usual name.

PR #1185, opened 2018-07-18
Add options button "undo ingoring devices and folders" (fixes #938)

### LINKED ISSUE #938: List of ignored devices
###    what happened,

I swiped away an "Unknown device [ID] wants to connect" and then the next day it was there again.

###    what you expected to happen instead, and

I would be added to a list of "ignored devices" under the "Devices" tab/page, which I could later add if I felt like it

###    any steps to reproduce the problem.

Add a new device to network, don't add it to your phone.

### Version Information

    App Version: 0.9.12
    Syncthing Version: v0.14.32
    Android Version: Android 4.3 on Sony LT25i

ghost:
Purpose
Fix issue #938 "List of ignored devices" by adding an options button to undo ignoring device/folder decisions the user has made. I think that's enough and we currently don't need to show a list with editing abilities for that niche use case.

Related issue
#938 

Testing
Verified working on device lg-h815 running Android 7.1.2 at commit https://github.com/syncthing/syncthing-android/commit/6008131ef30e227e7f813b3a9924ddd10f0b5366 .

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
@@ -81,6 +81,7 @@ public class SettingsActivity extends SyncthingActivity {
         private static final String TAG = "SettingsFragment";
         private static final String KEY_EXPORT_CONFIG = "export_config";
         private static final String KEY_IMPORT_CONFIG = "import_config";
+        private static final String KEY_UNDO_IGNORED_DEVICES_FOLDERS = "undo_ignored_devices_folders";
         private static final String KEY_ST_RESET_DATABASE = "st_reset_database";
         private static final String KEY_ST_RESET_DELTAS = "st_reset_deltas";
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -498,6 +501,25 @@ public boolean onPreferenceClick(Preference preference) {
...
+                            .setMessage(R.string.undo_ignored_devices_folders_question)
+                            .setPositiveButton(android.R.string.yes, (dialog, which) -> {
+                                if (mApi == null) {
+                                    Toast.makeText(getActivity(),
+                                            getString(R.string.generic_error) + getString(R.string.syncthing_disabled_title),
+                                            Toast.LENGTH_SHORT).show();
+                                    return;
+                                }
+                                mApi.undoIgnoredDevicesAndFolders();
+                                mRequireRestart = true;
```
AudriusButkevicius: Why needs a restart?

Catfriend1: It's like all the other options where "mRequireRestart = true": If this flag is true, the config is sent to syncthing on pressing home or leaving the SettingsActivity. This is our new settings apply routine we've introduced in PR #1099 . I rather wouldn't like to write extra code and try out if a just-send-the-config thing works. The user can pretty wait for a clean restart as this is rarely clicked.

Catfriend1: Btw undoIgnoredDevicesAndFolders only alters the in-memory cached config held by the wrapper not syncthing core. So the config must be sent by PostConfigRequest (see #1099) later which the flag indicates.

AudriusButkevicius: We should rename the flag, its confusing.

Catfriend1: So I'll make a follow-up PR and rename it to "mPendingConfig" as that's what it indicates the config still needs to be sent to syncthing.

AudriusButkevicius: You can just do it in this one

Catfriend1: Ok, here we go :) 34748e476d31d9bd8f874d87b70a98b6385d258f

AudriusButkevicius (approved):

Issue #1173, opened 2018-06-28
Disable Syncthing in Flightmode


Wiiilmaa:
I use syncthing with conditions "run in background" "only if wifi active".
Theese settings cause syncthing to start, when flightmode is activated.
expected behaviour: Do not run syncthing in flightmode if Wifi is deactivated or not connected.

Catfriend1:
I also experienced this problem before. I think its a question of what fits best to most users.
Personally, I disagree with the request as I'm often in flight mode when I don't need my mobile phone/data plan but having wifi enabled.  So syncthing definitely should run.

Anyways, I got your point, you want to have it off as you're not in the network you want. I think we should note this for my future-planned work on the run-conditions as I intent to suggest a refactor for that.
A suitable new option I would like to suggest is adding "Respect flight mode and put syncthing off" so users preferring your scenario can tick the box and others like me can sync when wifi-connected during flight mode. Compare it - e.g. to the app FolderSync - which has a similar option.

There are more tickets with discussion about more granular run conditions that are conceptually fitting the requirement that syncthing can sync both ways within an undefined timeframe when device come online.

Wiiilmaa:
Your point is absolutely valid. I also have a tablet, which is in Flight Mode 100% of the time with Wifi enabled.
But is there no way to determine if Wifi is enabled (or even better connected)? I don't see why "only if wifi active" option has to start syncthing in Flightmode with no networks enabled..
In my opinion it fits best to all users, if "only if wifi active" option enables synthing "only if wifi is active" instead of "only if wifi active or ( wifi disabled and flightmode activated )"

Catfriend1:
It is, user can decide by pref.

Issue #937, opened 2017-08-15
Add run condition: Avoid metered networks


HaleTom:
To save money and avoid syncthing using a mobile phone tethered network, (what Android calls a "metered" network), can the following `Run Condition` be added:

* `Avoid metered networks`

Nutomic:
This is already covered by the "run only on wifi" condition. If that condition is checked, Syncthing will be disabled on metered networks (on Android 4.1+). Maybe we should put that info in the description?

HaleTom:
I'd suggest put the control in the hands of the user: If I am running Syncthing on my tablet, tethered to my phone, I am on wifi but _may or may not_ wish syncing to occur on this tethered network, separate from other un-tethered networks.

carstenhag:
@Nutomic To enable "run only in wifi" I have to enable "always execute in background", which should not be a dependency in my opinion.

I should be able to only let syncthing sync via wifi without the background thing.

ndnihil:
I agree that always run in background should be separate from run on wifi. I'd even go as far as to say that each sync'd folder should have it's own criteria where you can choose what content syncs based on what connections are available. My use case aims for a scenario like:

Syncthing Servers:
  * server-home (internal only, full sync/backup of devices)
  * server-work (internal only, sync of work folder on phone/tablet/laptop)
  * server-pub ('Share via Syncthing' dir for placing content on public web repo)

Devices:
* myphone:
    * /sdext - HomeSSID/HomeVPN - server-home
    * /sdext/work - WorkSSID/WorkVPN - server-work
    * /sdext/pub - AnySSID/LTE/4G/3G - server-pub via 'Share via Syncthing'

* mytablet:
    * /sdext - HomeSSID/HomeVPN - server-home
    * /sdext/work - WorkSSID/WorkVPN - server-work
    * /sdext/pub - AnySSID/LTE/4G/3G - server-pub via 'Share via Syncthing'

* mylaptop:
    * ~/files - HomeSSID/HomeVPN - server-home
    * ~/work - WorkSSID/WorkVPN - server-work

* herphone:
    * /sdext - HomeSSID/HomeVPN - server-home
    * /sdext/pub - AnySSID/LTE/4G/3G - server-pub via 'Share via Syncthing'

* hertablet:
    * /sdext - HomeSSID/HomeVPN - server-home
    * /sdext/pub - AnySSID/LTE/4G/3G - server-pub via 'Share via Syncthing'

* herlaptop:
[...]

Issue #729, opened 2016-08-24
Show reason why Syncthing is stopped


Nutomic:
Eg because wifi is disabled, battery saver is active, etc.

The status could be shown in the notification, and in the drawer. Maybe also in the disabled dialog, but hopefully we won't need that dialog much longer.

See #727

conrad-heimbold:
The Android Material Doc's say, you should not use the notification bar for "Operations that don't require user involvement, like syncing information". 
Source: https://material.google.com/patterns/notifications.html 

Please respect these Doc's and do only show notifications, if files have been added or deleted, a new device wants to connect, or anything similar which is really a notification. 

If you want to know, what an app is doing in the background, just open that app! Syncthing should show its status (starting, running, paused, stopped) and reason for that (no wifi, not connected to charger, etc.) inside of the app, not in the notification area.

capi:
This ticket is about showing it inside the application (in the dialog that pops up at the moment), not about notifications. Please don't spam all notification-related threads now with this request.

wweich:
To wrap up #765: We should not only show this info in the notification, as it can be disabled.

Catfriend1:
I'm working on this.

Catfriend1:
Duplicate of #1263 and solved by "Show reasons for disabled service in notification dialog (#1264)"

Issue #588, opened 2016-03-20
Run condition to respect Android's "Auto-Sync" quick toggle


kluppy:
Add an option to link Syncthing to the global Auto-Sync Setting as per [Coupling syncing to Android’s sync setting](https://forum.syncthing.net/t/coupling-syncing-to-androids-sync-setting/6987/9)

Magissia:
It would be useful indeed, if you have the courage, we could have a sync per share.

licaon-kter:
@Nutomic Thanks

Nutomic:
I've reverted this commit, as #727 shows, there's more work required for this (like adding an option, and/or showing the reason syncthing is stopped).

acornsirup:
The running conditions are now managed in a separate menu item, would it be difficult to add another switch enabling this useful feature? Cheers and thanks.

Nutomic:
Not really that difficult. Contributions are welcome!

acornsirup:
Awesome! Unfortunately I can't code. I wish I were able to contribute better.

PR #1187, opened 2018-07-24
Refactor DeviceStateHolder and its receivers into RunConditionMonitor

ghost:
Purpose
Run conditions were detected async by listening to "incremental" information retrieved by broadcast receivers and "statically" by reading current device state or config. With that mixture, a "memory" of the current run/run-not state cannot be held plus "race" timing issues occur if the user changes preferences resulting in the "syncthing is disabled" dialog when it should run according to the newly set options.

The refactored code does exactly the same as the original code. From my point of view, the new implementation of the RunConditionMonitor fits better for future expansion and reflects the need of re-thinking the run condition preferences. I already prepared functions to properly detect mobile data and flight mode which will then be used in a follow-up PR.

As a plus, we'll also be able to make a nicer screen explaining the user "why syncthing is disabled" in MainActivity as all asynchronous information collection that previously would have made it hard to determine which circumstance was the cause for enabling/disabling syncthing was refactored into synchronous linear condition checks. To save phone's battery, the check of which conditions apply is still triggered by broadcasts received from the android system.

When completed, this will be the base to fix the issues:
- #1173 "Disable Syncthing in Flightmode"
- #937 "Add run condition: Avoid metered networks"
- #729 "Show reason why Syncthing is stopped"
- #588 "Option to link Syncthing to Global Android "Auto-sync Data""
- #549 ""WiFi only" should not depend on "background/foreground", Wifi only should be default"

Testing
Verified working on device lg-h815 running Android 7.1.2 at commithttps://github.com/syncthing/syncthing-android/pull/1187/commits/b2ea83d47b4ae2a01f5ce96b5fd76a1215f35844  .


app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java
```diff
@@ -0,0 +1,310 @@
...
+            return false;
+        }
+        if (!ni.isConnected()) {
+            // No network connection.
+            return false;
+        }
+        return cm.isActiveNetworkMetered();
+    }
+
+    private boolean isMobileDataConnection() {
```
AudriusButkevicius: Not used?

AudriusButkevicius (commented):
So this looks all good to me apart from a few unused functions, which I am not sure I understand the purpose off.

app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java
Catfriend1: Currently yes. Is it okay if I'll need that function later as I intend to make a new rjn condition gui proposal in a follow-up pr?

AudriusButkevicius: I think that's when you should add these functions, in a PR where they are actually used.

Catfriend1: Ok, removed them. Ready?

AudriusButkevicius (approved):

AudriusButkevicius:
If you plan some features shortly, let me know, and I'll hold back the release.

Catfriend1:
Thanks for leaving a note on this. From my side, it's okay to make a release. The next enhancements I've planned regarding the run conditions aren't urgent so they can wait to be released :).

Catfriend1:
I guess the release will contain 14.49? Then everything is okay as I already did my testing on 14.49-rc.4 with the current master. Thanks for your work on "syncthing core", thumbs up for 14.49 to be finally there :)

PR #1189, opened 2018-07-25
Better run conditions UI

### LINKED ISSUE #549: "WiFi only" should not depend on "background/foreground", Wifi only should be default
've been an early adopter of btsync and used it for a long time. When things became sour there I started investigating syncthing and watched if I could use it as an alternative.

After successfully running it for a while on both Linux (in datacentre) and Windows I started my first experiment on Android.

I had some trouble loading the sync as I expected "the sync" to be pushed to my Android just like on the other devices. When I did get it to work it immediately started sending all the data that was in the sync I was testing with.

I wasn't prepared that this could happen as all syncing software on mobile devices I've used thus far are made in a way that syncs are by default "on demand".
Now I know this hasn't been implemented (yet??).
This makes it useless for all my real world scenarios I have currently working with other software.

But my bad luck didn't stop there.
Whilst syncing I got a message that my dataplan was running out and before I could stop syncthing it burned 2.6 GB of the 3 GB dataplan.
I had my WiFi turned off because I travelled that morning by train which always gives me a bad WiFi-connection.

Some said I should have turned "WiFi only" off, but at this stage I was still running it as a foreground application. That option is not selectable then,,,,

IMHO the WiFi-only should be the initial setting and this should not depend on ""foreground/background"

I've removed the application for the time being, but I will install it immediately whenever an "on demand" is implemented.

Remark:

If "on demand" would be possible (and turned on by default) it would make sense to couple it with foreground/background. Now it doesn't (as I found out).
I would have rather paid the € 10,- to you instead of T-Mobile.

### LINKED ISSUE #588: Run condition to respect Android's "Auto-Sync" quick toggle

### LINKED ISSUE #937: Add run condition: Avoid metered networks

### LINKED ISSUE #1173: Disable Syncthing in Flightmode

ghost:
Purpose
Make the run conditions clearer to understand for the user.
Provide the ability to have ticked run condition options effective when the app runs in foreground or background. Previously, run conditions only applied to the "run in background" option which lead to an accidentially burnt data plan for some users (see issue #549).

Screenshots
image
image

Fixed issues:
- "WiFi only" should not depend on "background/foreground", Wifi only should be default" (fixes #549)
- "Disable Syncthing in Flightmode" (fixes #1173) ... and enable it on phones that have problems detecting wifi in flight mode the user can manually connect.
- "Add run condition: Avoid metered networks" (fixes #937)
- "Option to link Syncthing to Global Android "Auto-sync Data"" (fixes #588)

Related PR:
This is a follow-up to PR #1187.

Status
Most parts have been tested, no problems discovered until now. ( device lg-h815, Android 7.1.2, commit  https://github.com/syncthing/syncthing-android/pull/1189/commits/095a98ed3f4fd797dc925667e8016d69ed8f1e11 )
I offer doing some more intense testing when the review is done.

Forum discussion
https://forum.syncthing.net/t/new-and-improved-run-conditions-for-syncthing-android/11882


Catfriend1:
@AudriusButkevicius @capi @Nutomic 
Hi together,
I've made a proposal of a new run conditions UI and verified everything working correctly at commit https://github.com/syncthing/syncthing-android/pull/1189/commits/50ea1e47de609de6f00c02ab1c2bdb2e379a3eb5 . May I have your honest feedback on this. Please tell me if you like/don't like it. I would love to see the long outstanding issues fixed by this.

Although this PR's changes completely work built on top of the current master, some parts of it should get a follow-up by introducing the most important three options like "mobile data, wifi, metered wifi" via the first start welcome wizard (see PR #1176). It's technically okay but not very nice to greet a new user that has just installed syncthing app while on mobile data connection with "syncthing is disabled" (as "Run on wifi" is the default). This should be explained by the new welcome wizard slides.

AudriusButkevicius:
So the wording can be improved, and I haven't looked at the code, but I guess it's a sensible improvement.

What happens if you uncheck run on wifi and run on mobile data?
Does it even make sense not to run on wifi ever?

One of the options is in german so I have no idea what it means.

Catfriend1:
@AudriusButkevicius I think you meant this part:
image
This option wasn't modified it works like it has been before. It is the "Respect android battery saving" run condition. If checked, it ensures syncthing gets disabled as soon as the phone gets in the "save energy" state (during the last percent when discharging or when the quick toggle is hit).

Catfriend1:
@AudriusButkevicius ( Does it even make sense not to run on wifi ever? )
Yes, looking through the issues I recall some users told that they use syncthing in "default configured manner" over mobile data connection syncing little amounts of data. I wouldn't cut down this possibility by the run conditions but syncing on wifi being the default (which is proposed here already) definitely makes sense. ( That's why we also need the welcome wizard to explain this and put a toggle on it. )

Catfriend1:
@AudriusButkevicius ( What happens if you uncheck run on wifi and run on mobile data? )
Syncthing gets into the "disabled" state as expected by logical means. We would have the later ability to add WiMax and so on as separate check boxes when users request it with a valid use case.
If you uncheck wifi+mobile data, you can for example add "Run when device in flight mode" and syncthing gets started when totally offline. On older phones, the flight mode introduces wifi detection problems (already described) and the box can help to combine with "run on wifi".

AudriusButkevicius:
If anything I think we should invert the Wifi condition and have it "Disabled on Wifi" or something like that, as it does not make sense not to want to have that box checked in most cases.

Catfriend1:
@AudriusButkevicius You're right about the wifi check box. We always want it. Except on tethering. That's why the wifi checkbox enables the "Run when wifi tethered" checkbox subsequently. The reason why I used the "positive naming" (instead of the old "blacklist" naming with the "only" wording) was it's easier to understand for newbies as they go round ticking "what they want to happen" and not what they don't want. From my usage perspective, it's no problem to invert this as I know what's meant. What I still consider a little bit unlucky are the "blacklist" conditions like "charging", "power saving" or "auto-sync data". Those act as blockers disregarding all other circumstances like mobile data, wifi, etc. when one of them is ticked and the condition is not met. I hope we get more opinion on this as I feel like I should not decide this alone. I think that was the reason for Nutomic originally wording the conditions "stop on mobile data" and "stop on wifi". Btw the code is easier to read (and handle) if we do "positive" cases where we can. If you like, I'll put a commit so you can see the "Disable Wifi" variant and test it how the UI feels like then?!

Catfriend1:
@AudriusButkevicius Okay, looks better now :)

app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java
```diff
@@ -134,64 +156,107 @@ public void updateShouldRunDecision() {
...
 
-        // Run only on whitelisted wifi ssids.
-        if (prefRunOnlyOnWifi && prefWifiWhitelistEnabled) {
-            // Wifi connection detected. Wifi ssid whitelist enabled.
-            Log.v(TAG, "decideShouldRun: prefRunOnlyOnWifi && prefWifiWhitelistEnabled");
-            return isWifiConnectionWhitelisted(whitelistedWifiSsids);
+        // Run in flight mode.
+        if (prefRunInFlightMode && isFlightMode()) {
+            Log.v(TAG, "decideShouldRun: prefRunInFlightMode && isFlightMode");
+            return true;
```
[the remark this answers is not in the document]
AudriusButkevicius: After second look, flight mode doesn't actually check for flight mode, but rather for active connection, in which case this function `isFlightMode` should be renamed to represent what it actually does, ala `isConnectedToAnyNetwork`

app/src/main/res/values/strings.xml
```diff
@@ -285,19 +285,29 @@ Please report any problems you encounter via Github.</string>
...
+    <string name="always_run_in_background_enabled">to_be_deleted</string>
+    <string name="always_run_in_background_disabled">to_be_deleted</string>
+    <string name="sync_only_charging">to_be_deleted</string>
+    <string name="sync_only_wifi">to_be_deleted</string>
+    <string name="sync_only_wifi_ssids">to_be_deleted</string>
 
-    <!-- Preference summary in case it is enabled -->
-    <string name="always_run_in_background_enabled">Syncthing always runs in the background, according to preferences below.</string>
+    <string name="run_conditions_title">Run Conditions</string>
+    <string name="run_conditions_summary">Use the following options to decide when the syncthing UI will be available and synchronization takes place.</string>
```
AudriusButkevicius: Use the following options to decide when Syncthing will run

app/src/main/res/values/strings.xml
```diff
@@ -285,19 +285,29 @@ Please report any problems you encounter via Github.</string>
...
 
-    <!-- Preference summary in case it is disabled -->
-    <string name="always_run_in_background_disabled">Syncthing only runs when explicitly started, and can be stopped by menu button.</string>
+    <string name="run_on_mobile_data_title">Run on mobile data</string>
+    <string name="run_on_mobile_data_summary">Run when device is connected to the mobile data network. Attention: This can fastly consume your data plan if you sync a large amount of data.</string>
```
AudriusButkevicius: Run when device is connected via the mobile data network. Warning: This can consume a lot of data from your mobile operator data plan if you sync large amounts of data.

app/src/main/res/values/strings.xml
```diff
@@ -285,19 +285,29 @@ Please report any problems you encounter via Github.</string>
...
 
-    <string name="sync_only_charging">Run only when charging</string>
+    <string name="run_on_wifi_title">Run on wifi</string>
+    <string name="run_on_wifi_summary">Run when device is connected to a wifi network.</string>
 
-    <string name="sync_only_wifi">Run only on wifi</string>
+    <string name="run_on_metered_wifi_title">Run on metered wifi</string>
+    <string name="run_on_metered_wifi_summary">Run when device is connected to a metered wifi network e.g. a hotspot or tethered network. Attention: This can fastly consume your data plan if you sync a large amount of data.</string>
```
AudriusButkevicius: Attention: This can consume large portion of your data plan if you sync a lot of data.

app/src/main/res/values/strings.xml
```diff
@@ -306,9 +316,29 @@ Please report any problems you encounter via Github.</string>
...
+    </string-array>
+
     <string name="respect_battery_saving_title">Respect Android battery saving setting</string>
     <string name="respect_battery_saving_summary">Disable Syncthing if battery saving is active</string>
 
+    <string name="respect_master_sync_title">Respect Android \'Auto-sync data\' setting</string>
+    <string name="respect_master_sync_summary">Disable Syncthing when the quick settings tile \'Auto-sync data\' is toggled off.</string>
+
+    <string name="run_in_flight_mode_title">Run when device is in flight mode</string>
+    <string name="run_in_flight_mode_summary">Enable if your phone has problems detecting manually connected wifi during flight mode.</string>
```
AudriusButkevicius: `Enable if your phone has problems detecting manual wifi connections during flight mode.`

Also, it seems some sentences end with a dot, some don't. We should normalise this.

app/src/main/res/values/strings.xml
```diff
@@ -306,9 +316,29 @@ Please report any problems you encounter via Github.</string>
...
+
+    <string name="service_settings_title">Service setting</string>
+    <string name="service_settings_summary">Running Syncthing as a service adds a persistent notification to prevent it from being ended by android. Checking this option starts syncthing automatically when the phone boots.</string>
```
AudriusButkevicius: Running Syncthing as a service. Adds a persistent notification to prevent it from being terminated by Android. Checking this option starts Syncthing on operating system startup.

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -253,6 +276,30 @@ public void onActivityCreated(Bundle savedInstanceState) {
...
+                PreferenceScreen categoryRunConditions = (PreferenceScreen) findPreference(openSubPrefScreen);
+                final ListAdapter listAdapter = prefScreen.getRootAdapter();
+                final int itemsCount = listAdapter.getCount();
+                for (int itemNumber = 0; itemNumber < itemsCount; ++itemNumber) {
+                    if (listAdapter.getItem(itemNumber).equals(categoryRunConditions)) {
+                        //simulates click on the sub-preference
+                        prefScreen.onItemClick(null, null, itemNumber, 0);
+                        break;
+                    }
+                }
```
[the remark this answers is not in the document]
Catfriend1: We open the sub preferences screen by triggering the onItemClick event. That's the same what happens when the user clicks on "run conditions" to open the submenu.

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -317,22 +364,20 @@ private void setPreferenceCategoryChangeListener(
...
-                    mSyncOnlyCharging.setEnabled(value);
-                    mSyncOnlyWifi.setEnabled(value);
-                    mSyncOnlyOnSSIDs.setEnabled(false);
-                    // Uncheck items when disabled, so it is clear they have no effect.
-                    if (!value) {
-                        mSyncOnlyCharging.setChecked(false);
-                        mSyncOnlyWifi.setChecked(false);
-                    }
+                case Constants.PREF_POWER_SOURCE:
+                    mPowerSource.setValue(o.toString());
```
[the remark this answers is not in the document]
Catfriend1: It's correct. Did not crash during test.

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -253,6 +276,30 @@ public void onActivityCreated(Bundle savedInstanceState) {
...
+        private void openSubPrefScreen(PreferenceScreen prefScreen) {
+            Bundle bundle = getArguments();
+            if (bundle == null) {
+                return;
+            }
+            String openSubPrefScreen = bundle.getString(EXTRA_OPEN_SUB_PREF_SCREEN, "");
+            // Open sub preferences screen if EXTRA_OPEN_SUB_PREF_SCREEN was passed in bundle.
+            if (openSubPrefScreen != null && !TextUtils.isEmpty(openSubPrefScreen)) {
+                Log.v(TAG, "Transitioning to pref screen " + openSubPrefScreen);
+                PreferenceScreen categoryRunConditions = (PreferenceScreen) findPreference(openSubPrefScreen);
```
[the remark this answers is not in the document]
Catfriend1: It could be null as all the other stuff having a constant referring to the xml layout. As I already looked for a solution to directly refer to R.layout resources by names I discovered android java doesn't have a function/class for doing it. If one xml part is renamed and the constant in code not, the app would in this case crash when trying to open the main pref screen.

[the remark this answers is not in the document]
Catfriend1: So answer to the question: Currently never null, if someone does not code correct, could get null immediately noticed when clicking through the UI during test.

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -253,6 +276,30 @@ public void onActivityCreated(Bundle savedInstanceState) {
...
+                final ListAdapter listAdapter = prefScreen.getRootAdapter();
+                final int itemsCount = listAdapter.getCount();
+                for (int itemNumber = 0; itemNumber < itemsCount; ++itemNumber) {
+                    if (listAdapter.getItem(itemNumber).equals(categoryRunConditions)) {
+                        //simulates click on the sub-preference
```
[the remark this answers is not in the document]
Catfriend1: ok

app/src/main/res/values/strings.xml
Catfriend1: I put your suggestion in hoping no one gets confused that syncthing binary is not syncthing android app.

app/src/main/res/values/strings.xml
Catfriend1: ok

app/src/main/res/values/strings.xml
Catfriend1: ok

app/src/main/res/values/strings.xml
Catfriend1: ok for the new string. Dotted the battery saving string.

app/src/main/res/values/strings.xml
Catfriend1: ok

app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java
```diff
@@ -134,64 +156,107 @@ public void updateShouldRunDecision() {
...
-            Log.v(TAG, "decideShouldRun: prefRunOnlyOnWifi && !isWifiOrEthernetConnection");
-            return false;
+        // Run on wifi.
+        if (prefRunOnWifi && isWifiOrEthernetConnection()) {
+            if (prefRunOnMeteredWifi) {
+                // Check if we are on metered wifi and if wifi whitelist run condition is met.
+                if (isMeteredNetworkConnection() && wifiWhitelistConditionMet(prefWifiWhitelistEnabled, whitelistedWifiSsids)) {
+                    Log.v(TAG, "decideShouldRun: prefRunOnWifi && isWifiOrEthernetConnection && prefRunOnMeteredWifi && isMeteredNetworkConnection && wifiWhitelistConditionMet");
+                    return true;
+                }
```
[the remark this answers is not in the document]
Catfriend1: You're right about this. I've corrected this.

app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java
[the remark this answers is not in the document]
Catfriend1: isFlightMode is a function that works differently on Android 4.x than on 5+ phones. It's a little bit tricky on Android 5+, for example. To explain it on your example.
( If I am connected to a metered wifi, and prefer not to run over metered wifi, yet happen to be in flight mode ) ... you'll get an isFlightMode == false so syncthing will not run.

This is why the name is correct - even if the content of the function (as you've observed right) does something looking completely different from flight mode detection. We have this ni == null "hacky" part in since "@Nutomic" found it as its the only one compatible down until Android 4.x In newer android versions, the ConnectivityManager got expanded in android and can return a connection type flight mode. I see no reason to change the function name as it only will return true if we (really) are in flight mode. I've also tested this. What buggy phones will do is written on the paper of future issues.

Catfriend1:
Is this okay for everybody with the recent commits (up to https://github.com/syncthing/syncthing-android/pull/1189/commits/b456249f6258e9337a19cc5d6a98334219e16640 = Review 2/2) ? If yes, I'd like to go on testing it and then base the welcome wizard on top of it to do the above mentioned explanations before the user gets a "syncthing is disabled" on frist app launch when installing "from" mobile data.

imsodin:
@Catfriend1 It looks to me like the strings changed since the screenshot in your first comment. Can you verify whether it is still up to date or add a new screenshot if not please - just so I dont review outdated stuff.

app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java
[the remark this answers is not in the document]
AudriusButkevicius: To be honest, this would be eaiser to read if it was:
```
if (runOnWifi && connectedToWifi) {
   if (!preferRunOnMetered && onMetered()) {
       return false;
   }
   if (wifiSsidWhitelisted) {
       return true
   }
}
```

Catfriend1:
@imsodin 
image
image
image
image

AudriusButkevicius:
Is it possible to make the section separators look different than just the rows that have the checkboxes?
Or put a separator after them or something?

AudriusButkevicius:
I also made a comment about simplifying the logic.

Catfriend1:
@AudriusButkevicius The preferences screen unfortunately hasn't got separators. I already tried to solve this requirement but as this is not possible used the text elements in between.

AudriusButkevicius:
https://stackoverflow.com/a/35675937/1508606 ?

Catfriend1:
@AudriusButkevicius I got your comment via email but couldn't find it in here somehow. As I recall you'd like to switch from "negative"-if blocks on top to "positive" if blocks one negative in between and then positive. From my point of view, that's not easier to maintain the code as the wild "positive", "negative" path switching with suddenly "return-out" was the way the code looked before the PR and already had a bunch of bug fixes necessary in the past. To elaborate on the code strategy I used:

a) Some conditions are negative paths in the decision taking. So if we want to respect battery saving for example and this conditions applies (phone is in saving mode) we don't need to check any further and return false (back out of the whole checking routine).
b) Some conditions are positive and allow to be combined. (Run on mobile, run on wifi, run on metered wifi, later: maybe run on roaming) I agree you found some combinations don't make sense in real life and your point is valid (the if's would never go through to check if you're on mobile and metered ...) BUT, on my opinion those positive traits are better to read and understand in code (from human perspective that we initially think of TRUE and not of FALSE things > "don't think of a purple elefant" "what did you just think of in the first place"). That's why I'd suggest to stick with the positive way, even if the is one more line in the code. As the decision "if clauses" are logged down for better issue analysis we'll need later, it's also easier to go through positively thinking instead of negatively. Users report issues like "I'd ticket wifi and mobile data and restrict ssid, so it didn't work as I was on wifi with the ticked ssid...".

If it later turns out I'm not right on this, I'll voluntarily push up a PR following your advice immediately.

Catfriend1:
@AudriusButkevicius I'll try that, but according the the user feedback on SO it doesn't work.
image

AudriusButkevicius:
Perhaps using categories will give some distinction between the different groups?

I also found this: https://chromium.googlesource.com/chromium/src/+/master/chrome/android/java/res/layout/divider_preference.xml

which might or might not be useful I guess.

imsodin (commented):
Is there a main toggle/option between manually running and running automatically by these run conditions? In my opinion the top level option should be a simple two way choice. All of the specific run conditions, should only be available if the user chose to run automatically.

Also I would group the run conditions in wifi, mobile data and "energy", e.g. metered and specific wifi should be grayed out if wifi isnt checked. The idea is to have as little options available at one time as possible to make it less confusing to the user.

I also have some comments on individual strings, but I will write them up once the general structure is agreed upon.

Catfriend1:
(https://chromium.googlesource.com/chromium/src/+/master/chrome/android/java/res/layout/divider_preference.xml )  style/Divider is not found when compiling. I couldn't find the styles.xml on the chromium src.

Catfriend1:
Futhermore, adding a View (as mentioned in the chromium xml) causes the app to crash as the tag is not allowed in the PreferencesScreen section.

Catfriend1:
Tried to add the following xml but crashes out.
`        <item
            android:left="6dp"
            android:right="6dp"
            android:drawable="@drawable/list_divider" />`

Catfriend1:
Putting this doesn't crash, but gets ignored by the PrefsFragment.
`        <Preference
            android:key="static_service_settings"
            android:persistent="false"
            android:selectable="false"
            **android:drawable="@drawable/list_divider"**
            android:title="@string/service_settings_title"
            android:summary="@string/service_settings_summary"/>`

Catfriend1:
@imsodin ( Is there a main toggle/option between manually running and running automatically by these run conditions? In my opinion the top level option should be a simple two way choice. All of the specific run conditions, should only be available if the user chose to run automatically. )
What you suggest was the case before this PR. I intentionally changed this behaviour with the "always run in background (as a service) option making the individual run conditions available to the user" as there were many complaints on the issue tracker like:
- "I didn't want to set it always running on boot, was in wifi, couldn't restrict to run on wifi (as it was greyed out), left my home and accidentially burnt through my data plan while being on mobile data connection."
- As @AudriusButkevicius said, running on wifi should be the default. I second this. Syncthing should (as requested by an issue on the tracker) default to run on wifi regardless if the user runs it permenantly in background and on boot or not.

Catfriend1:
@imsodin ( Also I would group the run conditions in wifi, mobile data and "energy", e.g. metered and specific wifi should be grayed out if wifi isnt checked. The idea is to have as little options available at one time as possible to make it less confusing to the user. )
The greying out is already in here. We can make sub pref screens I think, but I would suggest doing this in a follow-up PR together with my planned UI improvement showing the user "why syncthing is disabled = which run conditions matched/mismatched".

imsodin:
Default on running on wifi is fine for me. I would add that to the welcome wizard then, because anything that runs automatically should in my opinion announce that. This doesnt really conflict with what I proposed.

I dont understand the relation of those quotes to my proposal. I am not saying to remove the options, I would just hide them between a top level option of running manually vs running automatically with run conditions. I dont understand the drawbacks of that approach and I am pretty sure that many users will be overwhelmed when confronted with so many possibilites up front. However again, that is just my opinion.

If you already do the greying out, then that is fine for me. It already directs the focus, no need for general separation in my opinion. My point is just to visually convey which options are currently relevant and which arent.

Catfriend1:
Adding code doesn't work to get at least the checkbox dividers out, they are still drawn.
`// Remove dividers
            if (Build.VERSION.SDK_INT > Build.VERSION_CODES.LOLLIPOP) {
                // For SDK version 21 and below, the code has to be within on CreateView.
                View rootView = getView();
                ListView list = (ListView) rootView.findViewById(android.R.id.list);
                list.setDivider(null);
            }`
I've already spent hours on the divider problem, can we do this LATER please?!

Catfriend1:
( I dont understand the relation of those quotes to my proposal. I am not saying to remove the options, I would just hide them between a top level option of running manually vs running automatically with run conditions. ) @imsodin 
I hope I haven't misunderstood that part. Running the app manually is what most people I've syncthing android recommended to do. I like you idea offering that without any need for the user to care about the run conditions prefs. But the drawback I've seen from it is, the user may start the app and forget it still runs. If you uncheck "run as a background service", the app still would continue running in the background if you switch to another app meanwhilst. What I thought was protecting the user from forgetting to hit "Exit Syncthing" in the drawer pane. This "protection" by making the run conditions always effective with a default (so the user still doesn't need to care adjusting them in the menu) should help to follow the general attitude on android "apps mustn't be exited manually".

I completely agree to your suggestion adding a slide about mentioning the default (run on wifi) and that the user can change it any time. Syncthing Android has - on my personal opinion - "suffered for years" from users not understanding that running the app manually causes syncing on all networks available.

imsodin:
Ah, I might have misunderstood the intention behind run conditions: Do the apply regardless of how the app is started? If that is the case, then you could just have (maybe that is already the case) an option to start syncthing automatically and an option to apply run conditions, where all these options are offered.

Catfriend1:
@imsodin ( Do the apply regardless of how the app is started? )
Before PR: No. / With PR: Yes.

I think we should do this as you suggested: ( then you could just have an option to apply run conditions, where all these options are offered. ) Is it okay making the option default?

AudriusButkevicius:
Why would you not want the run conditions to apply always?
How did the run conditions get applied before, if not always?

Catfriend1:
@AudriusButkevicius Before the PR, run conditions only applied if the user decided to always run syncthing android in the background as a service when the phone booted. Accidents syncing on mobile data connection happened because most users manually started the app and forgot to close it manually through the drawer as switching over to another app didn't suffice.

AudriusButkevicius:
A right, this is a great improvement in that case.

AudriusButkevicius (dismissed):
In principle this looks good to me, but I'll leave for you two guys to decide once this is ready.

Catfriend1:
Still ToDo:
- @imsodin ( I also have some comments on individual strings, but I will write them up once the general structure is agreed upon. )
- Improve the welcome wizard and "synching is disabled" UI to avoid greeting users like @AudriusButkevicius commented above. I would do them in follow-up PR's.

Catfriend1:
Found a minor bug, will fix it shortly.
We should run "evaluateRunConditions" onResume() of the MainActivity. Otherwise, if the user changes the metered flag for a wifi in Android 7, the app will not notice it as there is no broadcast.

app/src/main/res/values/strings.xml
```diff
@@ -306,9 +316,29 @@ Please report any problems you encounter via Github.</string>
     <string name="sync_only_wifi_ssids_location_permission_rejected_dialog_title">Permission required</string>
     <string name="sync_only_wifi_ssids_location_permission_rejected_dialog_content">Starting with Android 8.1, location access is required to be able to read the WiFi\'s name. You can use this feature only if you grant this permission.</string>
 
+    <string name="power_source_title">Run when device powered by</string>
```
imsodin: Run when device **is** powered by

app/src/main/res/values/strings.xml
```diff
@@ -306,9 +316,29 @@ Please report any problems you encounter via Github.</string>
...
+
+    <string-array name="power_source_entries">
+        <item>AC and battery power</item>
+        <item>AC power</item>
+        <item>Battery power</item>
```
imsodin: Consequently you need to add a final dot here as well, if you do it on line 328 below.

imsodin (commented):
Now that I actually understand what this PR does (as in apply run conditions unconditionally, which is great!), I have other questions. I'll try to check in more often, I am just pretty busy away from the computer atm.

Before there was the always run in background option - that seems to be gone. Is that correct? If so, will Syncthing now always start automatically or never (I believe neither possibility would be good). I consider the old name confusion anyway, but an option like "Start Syncthing automatically" should still be there, right? Looking at the current settings structure, that option could be part of Behaviour?

The flight mode option isn't in the screenshot? What is its purpose anyway? I mean afaik that's just a quick toggle to disable both cellular and wifi, i.e. the wifi and mobile data run conditions should be enough?  
Same for `Service setting`.

The rest are just string nit-picks.

app/src/main/res/values/strings.xml
```diff
@@ -285,19 +285,29 @@ Please report any problems you encounter via Github.</string>
...
+    <string name="run_on_wifi_title">Run on wifi</string>
+    <string name="run_on_wifi_summary">Run when device is connected to a wifi network.</string>
 
-    <string name="sync_only_wifi">Run only on wifi</string>
+    <string name="run_on_metered_wifi_title">Run on metered wifi</string>
+    <string name="run_on_metered_wifi_summary">Run when device is connected to a metered wifi network e.g. a hotspot or tethered network. Attention: This can consume large portion of your data plan if you sync a lot of data.</string>
 
-    <string name="sync_only_wifi_ssids">Restrict to certain wifi networks</string>
+    <string name="run_on_whitelisted_wifi_title">Run on specified wifi networks</string>
+    <string name="run_on_whitelisted_wifi_networks">Run on selected wifi networks: %1$s</string>
```
imsodin: Maybe "Run on selected wifi networks **only**" or "Run **only** on selected wifi networks".  
(Bold is just for purposes of review, not to be done in the actual UI).

Catfriend1:
@imsodin The "start syncthing automatically option" equals to the previous UI option "always run in background" and is here:
image
The flight mode option is on the screenshot:
image
You are right that logically flight mode is a combination out of wifi and mobile data connection, but this option, as the summary below it says, is there for a workaround as some phones have problems detecting the flight mode correctly, plus, the user can make exceptions from it by enabling wifi manually when in flight mode (Android shows indeed two icons then, flight mode and wifi connection). That's why I made this option as there are tickets reported about it. I also have this problem on my phone. To better understand the flight mode issue it needs testing on different phones running different android versions. If you look at other apps like e.g. Folder Sync or CSipSimple that also have network-related run conditions you'll also find this "workaround option" for flight mode.

Catfriend1:
@imsodin The "Run as background service" option could be moved to another sub pref screen (which has to be available instead of greyed out when syncthing is not running). From code perspective, it is a run condition (see BootReceiver, NotificationManager and AppConfigReceiver) which applies to running the android app automatically (not syncthing core). From user perspective, it's indeed behaviour.
Hmm... you're right. If this will never lead to the screen "Syncthing is disabled" we can move it. I'll check if behaviour is always available and not greyed out, then move it.

imsodin:
Must have been blind regarding those options.  
I think it's a bit hidden at the bottom of run conditions for a pretty important setting, but thematically I don't think it's far off there. Also the small text below Service does explain what it means, but people are lazy. "Run as background service" is probably the technically correct term, but I believe something like "Start service automatically on boot" would be easier to understand.

Catfriend1:
@imsodin Are we ready now or can I still improve it?

imsodin (dismissed):

Catfriend1:
Ok, testing completed, no crashes or problems so far found.

Catfriend1:
@imsodin Would you please be so kind and review the "resolve merge conflicts" commit https://github.com/syncthing/syncthing-android/pull/1189/commits/a022ea54149c5191dfbb2f32fa8343059c4c4f13 ? I think its okay as those are two constants taken over from the welcome wizard PR manually as github couldn't merge automatically.

imsodin (approved):

PR #1195, opened 2018-07-26
Fix race during config reload and missing null check (fixes #1194)

### LINKED ISSUE #1194: Race in DeviceListFragment#updateList during config reload
Report from GPlay Console:

### Version Information
    App Version: 0.10.12 (and earlier)
    Syncthing Version: v0.14.49
    Android Version: Android 6.0

>
`Heute, 00:01 in der App-Version 4145 Huawei Honor 7 (HWPLK), Android 6.0
Bericht 1 von 1
java.lang.NullPointerException: 
  at com.nutomic.syncthingandroid.fragments.DeviceListFragment.lambda$static$0$DeviceListFragment (DeviceListFragment.java:33)
  at com.nutomic.syncthingandroid.fragments.DeviceListFragment$$Lambda$0.compare (Unknown Source)
  at java.util.TimSort.countRunAndMakeAscending (TimSort.java:320)
  at java.util.TimSort.sort (TimSort.java:185)
  at java.util.Arrays.sort (Arrays.java:1998)
  at java.util.Collections.sort (Collections.java:1900)
  at com.nutomic.syncthingandroid.fragments.DeviceListFragment.updateList (DeviceListFragment.java:94)
  at com.nutomic.syncthingandroid.fragments.DeviceListFragment.access$000 (DeviceListFragment.java:30)
  at com.nutomic.syncthingandroid.fragments.DeviceListFragment$1.lambda$run$0$DeviceListFragment$1 (DeviceListFragment.java:59)
  at com.nutomic.syncthingandroid.fragments.DeviceListFragment$1$$Lambda$0.run (Unknown Source)
  at android.os.Handler.handleCallback (Handler.java:743)
  at android.os.Handler.dispatchMessage (Handler.java:95)
  at android.os.Looper.loop (Looper.java:150)
  at android.app.ActivityThread.main (ActivityThread.java:5621)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:794)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:684)`

ghost:
Purpose
Fix race during config reload and missing null check in FolderListFragment and DeviceListFragment
See related issue: #1194 

Testing
Verified working at commit https://github.com/syncthing/syncthing-android/pull/1195/commits/6a8cc7a3afb19ef4f21ed9ae4cb78f1bebdbd545  (device lg-h815, Android 7.1.2)

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
@@ -121,6 +121,11 @@ public class RestApi {
     private final Object mAsyncQueryCompleteLock = new Object();
 
     /**
+     * Object that must be locked upon accessing mConfig
+     */
+    private final Object mConfigLock = new Object();
+
+    /**
      * Stores the latest result of {@link #getFolderStatus} for each folder
      */
     private HashMap<String, FolderStatus> mCachedFolderStatuses = new HashMap<>();
@@ -199,7 +204,9 @@ public class RestApi {
     }
 
     private void onReloadConfigComplete(String result) {
-        mConfig = new Gson().fromJson(result, Config.class);
+        synchronized(mConfigLock) {
+            mConfig = new Gson().fromJson(result, Config.class);
+        }
         if (mConfig == null) {
             throw new RuntimeException("config is null: " + result);
         }
@@ -456,7 +463,11 @@ public class RestApi {
     }
 
     public boolean isConfigLoaded() {
-        return mConfig != null;
+        Boolean configLoaded;
+        synchronized(mConfigLock) {
+            configLoaded = mConfig != null;
+        }
+        return configLoaded;
     }
 
     /**
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
@@ -392,6 +392,7 @@ public class RestApi {
                 Folder f = it.next();
                 if (f.id.equals(id)) {
                     it.remove();
+                    break;
                 }
             }
         }
@@ -464,6 +465,7 @@ public class RestApi {
                 Device d = it.next();
                 if (d.deviceID.equals(deviceId)) {
                     it.remove();
+                    break;
                 }
             }
         }
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
@@ -204,10 +204,12 @@ public class RestApi {
     }
 
     private void onReloadConfigComplete(String result) {
+        Boolean configParseSuccess;
         synchronized(mConfigLock) {
             mConfig = new Gson().fromJson(result, Config.class);
+            configParseSuccess = mConfig != null;
         }
-        if (mConfig == null) {
+        if (!configParseSuccess) {
             throw new RuntimeException("config is null: " + result);
         }
         Log.v(TAG, "onReloadConfigComplete: Successfully parsed configuration.");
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
```diff
@@ -199,7 +204,9 @@ public void reloadConfig() {
     }
 
     private void onReloadConfigComplete(String result) {
-        mConfig = new Gson().fromJson(result, Config.class);
+        synchronized(mConfigLock) {
+            mConfig = new Gson().fromJson(result, Config.class);
+        }
```
AudriusButkevicius: you should just use `private synchronized void` on all functions that touch the config, as you added a lock in one place but not others.

AudriusButkevicius: Synchronized functions is how you deal with concurrency and race conditions. It's not bad practice, the only other practice is using atomics (which I think in Java is close to impossible), or immutable objects which you always copy.

If you don't want to use synchronized, you need to synchronize around your lock every time you access mConfig in all other functions, as now it's a half way house, we lock in some cases but not the other.
If we lock in all cases this pretty much ends up being equivalent of just using `public synchronized` as that gets translated into `synchronized(restApi) { restApi.getDevices() }`. These locks are reentrant, and even if I didn't look at the code much, I can't imagine that any of these functions call out to a separate thread, and wait for it to respond, and the other thread tries to access the restApi object at the same time.

Catfriend1: Yeah, I'll put the synchronized in tomorrow, then let's run the app for test and see if most things are running correctly as far as it can be seen.

Catfriend1: Hmm... I've looked at the code again. To be "hands on" I'd suggest looking at RestApi#ignoreDevice. Just putting synchronized(mConfigLock) { ... } around the code in this function seems wrong to me as it calls RestApi#sendConfig conditionally. There, mConfig is read-accessed and sent via post (...). That means we would also have to add synchronized(mConfigLock) { ... } in sendConfig. This is the point where it is too complicated for me to do it now as "just adding it" would deadlock if the body of ignoreDevice holds the lock and sendConfig gets called while "locked".

Putting public synchronized void ignoreDevice() { ... } instead of the mConfigLock and doing this also on sendConfig() will not solve our problem. Every thread calling this RestApi#public functions will have its own lock, so one lock per function.

Can you give me a "hands on" example where to put a working lock, let's say, on two functions like the ignoreDevice and sendConfig (both called externally and one internally from the same class)? Seems I didn't get the point what exactly to do here.

AudriusButkevicius: Synchronized locks are reentrant, so you can do `synchronized(foo) { synchronized(foo) {} }` as long as it happens on the same thread. Synchronized methods lock around the object, not the method.

You have plenty of other sections that access mConfigvariabke which also need to be under a lock.

Catfriend1: Somehow I don't get rid of the bad feeling doing too much synchronize() adds in the code could break it and we don't see it coming. You might be technically right, but I feel better with granular improvements to the threading stuff I have to admin I don't have a complete oversight how "Nutomic" put all those threads and same-thread as well as foreign-thread handler, callback listener stuff. Problem is, I can put the improvement you require to merge this, but I cannot assure it won't negative impact. That's why I only started with one spot that got fixed.
@imsodin Maybe you can tell me if I'm right or wrong here. I **can** put the code Audrius' requested but I also don't want to break things going too far.

AudriusButkevicius (dismissed):
Looks good to me, feel free to merge.

AudriusButkevicius (approved):

app/src/main/java/com/nutomic/syncthingandroid/service/RestApi.java
imsodin: Yeah no...  
I quickly skimmed the conversation and I definitely can't give any relevant input. The first time I came in contact with parallel computing was on Syncthing in Go - I have no education on it, nor do I know java or the syncthing-android codebase. That's what I meant with high-level review in another place: I am happy to comment on concepts or interface (though I am no expert there either), but I can't in the technical division (and I don't see myself going there even long-term, so don't hold your breath ;) ).

I am aware this is disappointing, because I had very productive disagreements with Audrius (the amount of bulls*** that I would have inserted into Syncthing would have been astounding otherwise), but sometimes we did get stuck disagreeing - and then it's nice to get a tie-breaker from a third party. I just really can't be that party.

PR #1199, opened 2018-07-27
Fix sending wrongly encoded commands into the unix subshell (fixes #1198)

### LINKED ISSUE #1198: Incorrect read and write permissions detection
Incorrect read and write permissions detection for folders on the device (Nexus 5X, Android 8.1.0). The application says that the folder is read-only. At the same time, if you enable the option in the web interface, then the synchronization for write starts.

### Version Information

    App Version: 0.10.12
    Syncthing Version: v0.14.49
    Android Version: Android 8.1.0

ghost:
EDIT:
Purpose
This PR provides a fix for issue #1198 . The problem is that the "canWriteToFolder" check fails when the folder path to test contains cyrillic characters. Doing it from java would make the test inaccurate, so no help by implementing more code. Root cause of the problem: Wrongly encoded commands get sent into the unix subshell. Resolved by using BufferedWriter as it's already doing the proper encoding work for the subshell correctly and is a built-In java class.

Testing
Verified working on device lg-h815 running Android 7.1.2 at commit https://github.com/syncthing/syncthing-android/pull/1199/commits/7541fa978e6c7f71544c4fefc8bcf025d85fc6d4 on base of release tag 0.10.12

Screenshots
a) Russian folder name gets accepted by the shell and folder permission check in "canWriteToFolder" return the correct result (in this case on my phone read-writeable).
image
b) Log output
image

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
@@ -88,6 +88,7 @@ public class SettingsActivity extends SyncthingActivity {
         @Inject NotificationHandler mNotificationHandler;
         @Inject SharedPreferences mPreferences;
 
+        private Preference         mCategoryRunConditions;
         private CheckBoxPreference mAlwaysRunInBackground;
         private CheckBoxPreference mSyncOnlyCharging;
         private CheckBoxPreference mSyncOnlyWifi;
@@ -125,6 +126,12 @@ public class SettingsActivity extends SyncthingActivity {
         private Config.Gui mGui;
 
         private Boolean mPendingConfig = false;
+
+        /**
+         * Indicates if run conditions were changed and need to be
+         * re-evaluated when the user leaves the preferences screen.
+         */
+        private Boolean mPendingRunConditions = false;
 
         @Override
         public void onCreate(@Nullable Bundle savedInstanceState) {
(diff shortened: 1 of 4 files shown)


AudriusButkevicius:
This is a poor solution. There are plenty of other encodings that will have the same problem, Chinese, Japanese, Arabic, Greek etc.

Catfriend1:
( This is a poor solution. ) It is. Please merge it to help users out first ( to prevent user loss ) and I'll keep that on my todo list as we keep the original issue open. I'll do more research on it and propose a better solution as soon as I found one. Thanks for your understanding.

AudriusButkevicius:
No, sorry, but we either fix it properly or we don't, as we'll soon have another poor solution for a different encoding. 

If you want to make a "hotfix" for this, then disable the checking all together until there is a better way.

Catfriend1:
Yes, I'm on it.

AudriusButkevicius:
This PR should probably revert my changes.

AudriusButkevicius:
To be honest, we should revert this whole subshell nonsense, as there is no reasonable escaping.
User can create a folder with name ` foo; rm -rf /; echo` which will then be interpreted as:
`rm foo; rm -rf /; echo /.stwritetest`

AudriusButkevicius:
As I said, I think this whole "user input into a subshell" should be backed out.

Catfriend1:
@AudriusButkevicius I disagree. User cannot enter paths manually causing this as the canWrite... function is only called with parameter content that came from android's folder picker returning VALID paths.

AudriusButkevicius:
And what's preventing me from creating a directory with that name?

Catfriend1:
You can do so, but that is not suspect to this PR as it's expecting to have a "security team" that regulates this. It's totally clear to me I can offer you a text filter, e.g. RegEx to prevent this one case, but then someone could make up another, probably weird non-daily use case, security infringement use. Anyway that's not the main topic we must care about NOW. The user can also open a terminal emulator and enter the bad command by hand. He doesn't need the syncthing folder creation to do this as both have the same unix permissions on android.

AudriusButkevicius:
Sure, but a user can be tricked into accepting a folder offered by a remote peer that has the folder name with shell commands in it. I've posted a link on the forum for a utility function which should help work around this.

Catfriend1:
Agree to your forum post. Disagree with the folder share scenario, we hand in the path from the picker not the announced folder label.

AudriusButkevicius:
Sure. Syncthing prepopulates the path, perhaps android doesn't, hence this is not directly a vulnerability, but it's still an attack vector. There is nothing worse than to have a remote execution attack vector in a piece of software that claims to be secure, encrypted and all about privacy.

Catfriend1:
What about the license of StringEscapeUtils.java, its under apache license. Is this "compatible" and allowed? I think it's now your turn to escape the following I found out that would makes sense:
shellEscape(absoluteFolderPath) and then verifies it doesn't break anything else with the encoding that just got fixed.

PR #1202, opened 2018-07-28
Updated de (German) translation

wischi-chr:
Commit made via [Stringlate](https://lonamiwebs.github.io/stringlate/)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -33,7 +33,7 @@ android {
     }
     defaultConfig {
         applicationId "com.nutomic.syncthingandroid"
-        minSdkVersion 14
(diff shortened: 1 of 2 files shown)


AudriusButkevicius:
Translations are handled via transifex.

Catfriend1:
@wischi-chr Thanks for offering your help to the project. Unfortunately this is unreviewable due to reordered string positions in the xml all over. Please head to the syncthing forum and ping calmh to apply for being added to the transifex team.

AudriusButkevicius:
I don't think you need to be added to a team to update transaltions. I think you update translations and might just need approval from another translator.

Issue #1192, opened 2018-07-25
Context.startForegroundService() did not then call Service.startForeground()


Catfriend1:
From GPlay Console
![image](https://user-images.githubusercontent.com/16361913/43201310-67d6b43e-9018-11e8-89fa-f909cd0b5e59.png)

Impact
Android 8+ users cannot run syncthing as a background service without getting a hang if the service startup takes longer than 5 seconds. This is due to a recently introduced android restriction. There were also some users leaving feedback on google play related to this problem that "Syncthing would hang a lot on Android 8+".

### Version Information
    App Version: 0.10.12
    Syncthing Version: v0.14.49
    Android Version: Android 8+

Catfriend1:
> Heute, 12:07 in der App-Version 4145
Samsung Galaxy Note8 (greatqlteue), Android 8.0
Bericht 1 von 2

Der Thread ""main" (tid=1)" befindet sich im Ruhezustand.

ALLE MAXIMIEREN
"main" tid=1 Sleeping 
"main" prio=5 tid=1 Sleeping
  | group="main" sCount=1 dsCount=0 flags=1 obj=0x74e5cd70 self=0x7ec16c3a00
  | sysTid=8899 nice=0 cgrp=default sched=0/0 handle=0x7ec6b8b9c8
  | state=S schedstat=( 1676864215 465920588 1917 ) utm=143 stm=24 core=1 HZ=100
  | stack=0x7fef98b000-0x7fef98d000 stackSize=8MB
  | held mutexes=
  at java.lang.Thread.sleep (Native method)
- sleeping on <0x0658c8e4> (a java.lang.Object)
  at java.lang.Thread.sleep (Thread.java:373)
- locked <0x0658c8e4> (a java.lang.Object)
  at java.lang.Thread.sleep (Thread.java:314)
  at android.os.SystemClock.sleep (SystemClock.java:126)
  at com.nutomic.syncthingandroid.service.SyncthingRunnable.killSyncthing (SyncthingRunnable.java:310)
  at com.nutomic.syncthingandroid.service.SyncthingService.shutdown (SyncthingService.java:502)
  at com.nutomic.syncthingandroid.service.SyncthingService.onDestroy (SyncthingService.java:459)
- locked <0x055b314d> (a java.lang.Object)
  at android.app.ActivityThread.handleStopService (ActivityThread.java:3709)
  at android.app.ActivityThread.-wrap27 (ActivityThread.java)
[...]

PR #1205, opened 2018-08-05
Improve UI when syncthing is not running (fixes #729)

### LINKED ISSUE #729: Show reason why Syncthing is stopped

Catfriend1:
Purpose:
Show a better UI to the user and explain why syncthing is not running instead of the popup dialog "syncthing is disabled" blocking access to the UI and showing old "syncthing running state information" behind that popup dialog which is outdated because it's no longer running.

Screenshots:
a) Main UI - Syncthing is running - a new tab "Status" has been added, the button top right leads to the options.
image
image
image

b) Main UI - Syncthing is not running - tabs "folders" and "devices" previously showed outdated information, they are hidden now
image
image

c) Drawer - Syncthing is running - Menu is shown as "before this PR" - no changes
image

d) Drawer - Syncthing is not running - the drawer can now be opened but runtime stats and some action buttons are not available (as they only apply if syncthing was running)
image

e) After an (environmentally) run condition change - syncthing is starting up again.
image

Related issues:
#729 "Show reason why syncthing is stopped"

Testing:
Verified working on device lg-h815 running Android 7.1.2, last test was at commit https://github.com/syncthing/syncthing-android/pull/1205/commits/a202a06e4192d07ef30f4dbebdf8715d0195e64c .


Catfriend1:
@AudriusButkevicius I've planned the status tab to show an explanation "why syncthing is running", so both groups of users can get an answer - especially if they misconfigured a run condition pref or android "misbehaved in our eyes" (I suspect some newer Android's don't send the broadcasts every time but this needs for investigation to prove/disprove...) We would then get better bug reports if the users just tell us what's on the status tab.

AudriusButkevicius:
I don't think in the running case this os worthy its own tab. Perhaps the run condition window could indicate which ones are active which ones are not.

Catfriend1:
@AudriusButkevicius It's not only the run conditions... The drawer is currently cluttered with less important side information (CPU usage, RAM usage, ...) which can also go into the status tab in a follow-up PR.

AudriusButkevicius:
So drawer is ok, but I don't think its new tab worthy.

Catfriend1:
@AudriusButkevicius We cannot show zero tabs with a blank rectangle in the main UI while status indication is vital to the wrapper as it was with the old popup dialog showing centered. The drawer has to be full of actions (later).

Catfriend1:
What do the others think? @Nutomic @capi
The tab doesn't hurt anyone if it cleans our drawer now and later as no one has to look at it. The default tab is still folders.

AudriusButkevicius:
I don't think anything moves from the drawer while its running? So it would literally be a tab with a single string which I think its pointless. Perhaps we could have a read only/grayed out section in the run conditions (or condition status view). Run conditions are not pivotal to what the app does... Its like putting a tab for logs, it's useful, sure, but its only accessed when shit doesn't work and should not get a prime ui element.

AudriusButkevicius:
Yes, I don't use the app, but that doesn't mean I can't say something when I see a UX car crash in the works.

Catfriend1:
Ux car crash... Come on. This is an enhancement, if not you can politely ask if I would make you a follow up PR according to user voice.

calmh:
I don't use the app either, but I too have an opinion. In general, major UI changes require major justification.

That said, when starting the app, if syncing is not happening for whatever reason, it's totally appropriate to display the why and how front and center. A tab or other central UI element to explain why Syncthing is not running seems the obvious choice.

If I start the app and syncthing is running, the app should be about what's happening. Devices, folders, sync status, etc. There is apparently a drawer with meta information - CPU usage and so on. Adding a note here as to what exact conditions result in things currently being up and running seems appropriate. It's a minor detail, after all.

In the future, if there is lots of runtime status information that is relevant to day to day operations, I'm sure a new tab is the right place for it.

Catfriend1:
@calmh Thanks for your opinion on this.
( Devices, folders, sync status, etc. )
This part is exactly matching my intention making up this PR. You've mentioned "three tabs" indirectly. I agree that "syncthing is running" is a minor thing. If it was a simple (and less backed up by frequent user bug reports) task to implement run conditions properly on android, I'd agree putting a note in the drawer and say it's running but we currently often experience problems with this (run conditions do not fire as android does not deliver broadcasts or deliver differently in different os versions). Most of them I've already fixed but Android 8 introduces more service problems. e.g. #1192 

( In the future, if there is lots of runtime status information that is relevant to day to day operations, I'm sure a new tab is the right place for it. )
Can we please call out the future already at this point? My reason is I'm a daily sync user and finding myself often expanding the drawer and just looking at the updates of download/upload rate until they calm down to zero so I know syncing is complete for all devices and folders. A summary would be nice to have in the future like "all online devices synced 100%". This was just one idea to use the new status tab in the near future when it is visible while syncthing is running.

Plus, the drawer is so cluttered with less relevant information it from my perspective doesn't justify its main use case - the action buttons. If you have a small screen or rotate the phone to landscape, it looks like this:
image
On my opinion, we don't see the major UI elements like action buttons anymore (and have to scroll) but some static information on how many discovery options are ticked stay visible (index server - that textview doesn't contain any live status of connected index server nodes).

My intention having the tab is mainly aiming at:
- Having more space to display runtime information on syncthing and overall sync status
- Having the third area in original syncthing's web UI as a third tab, so "every area of the original web UI will be shown in  a tab" (folders, devices, info)
- Actions are on the web UI "gear menu" > that will be our drawer. Plus, it allows us cleaning up the misuse of the preferences screen that has already taken place. (We have a lot of actions like data export and import in our preferences, I think they belong into the drawer as a head word, e.g. "Backup & Restore" using new activities and UI launched by them to guide the user through the process.)

Currently, the Status tab is multiline. So CPU, RAM, Download, Upload, Indexserver can go to it. This would also avoid writing interfaces code (Fragment<->Service) two times, as the Drawer is a separate class than the status tab the code would partly double up. At the moment, this PR has it clearly and completely in the StatusFragment handling the new tab.

So, if you still insist on showing the "syncthing is running" in the drawer, please leave me on word and I'll take it as CEO decision and do it. Even being a little bit disappointed that the tab won't be our win to be able to display more useful information in the future. The drawer is already too "full" from my perspective and I expect future development in syncthing core will need more space on the wrapper's UI to keep it valuable to the user.

AudriusButkevicius:
I am all down for a grand vision of important things being in it's own tab, but let's add the tab when we have all the things that will go into it.

I don't mind a tab as long as it's not a single string of text like it is now.

Also, I am not sure I buy the "all online devices are 100% in sync" story going into this new tab as we already have devices tab which should be the right place for this info to be. Perhaps android does not show device completion, but there is no reason why it shouldn't be able to do that which then you can use to see if your devices are in sync or not.

imsodin:
I agree with both side's arguments: I see the merits of a dedicated status tab and I also think that the information that is in it in this PR does not justify it yet. So why not leave the status tab away when Syncthing is running (and display it when it isn't running as proposed) and add it generally when the drawer reorganisation (more menu options, no more status info) happens?

Catfriend1:
@AudriusButkevicius @imsodin I've moved the "live status" from the drawer to the status tab. The drawer is now looking cleaned up as only the syncthing (core) version is still displayed on top as it was before. All action buttons on the drawer can now be reached in portrait and landscape mode.

Catfriend1:
Hmm... please wait a little bit before reviewing this again. I see something is wrong with the handler callbacks that I need to fix first.

Catfriend1:
Gnarf, rotation changes at runtime make it crash ... Investigating ...

Catfriend1:
Yes, I'll do so when the last bug is fixed. Currently onResume() doesn't restore the active tab (minor glitch).

Catfriend1:
New screenshots related to the last changes:
image
image

New things sum'd up:
+ Drawer has more space for actions but still has the syncthing version label there unchanged.
+ Status shows the information about syncthing (core's) runtime status that was on the drawer before, refreshing in the same interval as before (5 secs). I've made that timer stop if the user doesn't look at it it will not poll for new status data. It immediately refreshes when the user switches to the "Status" tab.

Ready to see it in beta ;-) If it's not good enough, I can still improve it or we revert the whole thing.

AudriusButkevicius:
I'd like to see what others have to say, but I prefer the old way. Now the drawer is empty, and the status tab just looks a large text box.

As I said, when syncthing is not running, I think it's ok to show the tab and the reason.
When it's running, you rarely care, and if you DO care to know why it's running when you don't expect it to, you can just open the logs and read them.

Catfriend1:
@AudriusButkevicius There would be no problem to spice the FragmentAdapter a little bit up with a fresh design on the status tab but that'll be a follow-up if it is accepted. I've already put the text via an ArrayAdapter so it's easy to extend that class adding styling information for each independent line (= status) on it. If someone wishes for, I can also move actions from prefs into the drawer, but remember it's not "almost empty" on a landscape screen layout. Let's wait what the other say.

calmh:
I don't care, do either.

AudriusButkevicius:
@imsodin, the deciding vote?

Catfriend1:
Resolved merge conflicts caused by Transifex imports. (https://github.com/syncthing/syncthing-android/pull/1205/commits/5722d5a2a421bbd58601b857356e39686ad5dc37 and https://github.com/syncthing/syncthing-android/pull/1205/commits/d39437fb98c319e282f8a88b48ade2dbcea707ec)

Nutomic:
I definitely like the idea! I'm not quite sure if the status tab should be shown while Syncthing is running, but it makes sense now that you moved sync status etc over there.

However, I don't like the layout of the status tab itself. RIght now it looks really messy, and hard to read. Some suggestions:
- display a big :heavy_check_mark:  or :x: (or similar) to indicate if Syncthing is running
- run conditions should be way shorter, eg "wifi connected :heavy_check_mark: :x: (arranged in a nice grid)
- put a horizontal line before the other stats, and also put them in a grid

Catfriend1:
@Nutomic Thanks for your feedback on this. The graphical improvement would be the next thing on my list todo, So two "modules" still need a little bit more code. If this has a future I'd do the work.

vincentardern:
I like the idea in general, sorry I don't have time to critique it further. The drawer has always seemed to me like a weird place to get such detailed stats.

AudriusButkevicius:
Ok, I'll try to go through the code in the evening.

app/src/main/java/com/nutomic/syncthingandroid/fragments/StatusFragment.java
```diff
@@ -0,0 +1,260 @@
...
+                    "- " + syncthingService.getRunDecisionExplanation().trim().replace("\n", "\n- "));
+            default:
+                break;
+        }
+
+        // Add status holders refreshed by callbacks to the list.
+        if (mServiceState == SyncthingService.State.ACTIVE) {
+            synchronized (mStatusHolderLock) {
+                if (!TextUtils.isEmpty(mCpuUsage)) {
+                    statusItems.add(getString(R.string.cpu_usage) + ": " + mCpuUsage);
```
Nutomic: You shouldn't do `+` on strings because that won't work well with translations in various languages. Use string parameters instead.

app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java
```diff
@@ -58,6 +60,7 @@ public void onStatusChanged(int which) {
     private final Context mContext;
     @Inject SharedPreferences mPreferences;
     private ReceiverManager mReceiverManager;
+    private String mRunDecisionExplanation = "";
```
Nutomic: It would be cleaner if you just store a pointer to the string resource here, and only do `getString()` once you display it to the user. Use `@StringRes int mRunDecisionExplanation` for proper typing.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingService.java
```diff
@@ -587,6 +587,13 @@ public NotificationHandler getNotificationHandler() {
         return mNotificationHandler;
     }
 
+    public String getRunDecisionExplanation() {
+        if (mRunConditionMonitor == null) {
+            return "This should not happen: mRunConditionMonitor is not instantiated.";
```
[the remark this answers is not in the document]
Nutomic: Also don't hardcode messages, this should be in strings.xml.

app/src/main/java/com/nutomic/syncthingandroid/fragments/StatusFragment.java
Nutomic: Or handle it in the layout with seperate TextViews.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingService.java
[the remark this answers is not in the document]
Catfriend1: This is an exception because with the current PR's code,  this line is ensured to be never reached. It only fulfills the purpose of helping a later developer touching the code to notify him of a bad mistake.

PR #1209, opened 2018-08-06
Root only - Temporarily increase fs.inotify.max_user_watches to 128K (fixes #1208)

### LINKED ISSUE #1208: Syncthing is running out of inotify watches
See forum report at: https://forum.syncthing.net/t/android-failed-to-set-up-inotify-handler-please-increase-inotify-limits/11924/15

### Version Information
    App Version: 0.10.13
    Syncthing Version: v0.14.49

Quote:
> [Android] “failed to set up inotify handler. please increase inotify limits” iusearchbtw
Background
I use Syncthing on my phone, Arch laptop, and Ubuntu desktop. These devices all share the same exact folders with the same exact options (permissions, rescan interval, etc.). Up until now, syncing files between all my devices has worked properly (albeit a bit too slow and battery-consuming for my taste, especially for folders with many files).

Issue
Recently, two main folders on my Android client have appeared as “Out-of-sync”, despite removing these folders on all my devices and then propagating a fresh copy of them again to them. Only my desktop and my laptop are completely in sync, but my phone’s versions of the 2 folders in question are out of sync, no matter how much I wait. Moreover, I can no longer share new folders with my phone (only between my desktop and laptop); I wasn’t receiving the notifications to share the folders on my phone.

Then, I looked at the Android web GUI and it said this: “Filesystem Watcher Errors — failed to setup inotify handler. Please increase inotify limits. See …”. I went to the specified website and I executed the commands therein on my desktop and even my laptop for good measure (I ran the appropriate Arch command for my laptop), and rebooted, to no avail; I was receiving the same error on my Android client and the folders were still out of sync.

Then I wondered, “Wait, why did I have to run Linux commands to increase the inotify limit when the issue was happening on my phone? Is there an equivalent set of commands for Android?”.

Other Stuff I’ve Tried
I looked up this issue pretty much everywhere: the Github issues page, this forum, and the Arch Wiki. The only advice that I’ve seen for fixing this issue is on the FAQ page, and clearly those commands didn’t work at all. I’m guessing they only apply to Linux machines, not to Android phones.

Screenshot of my phone’s web GUI
20180804_18512520180804_185125.jpg1080x5928 1020 KB
Final Thoughts
Please help. Also, please include error messages in the official app GUI, not just the web GUI. I didn’t even know the error message would be in the web GUI until I randomly decided to look at it because I assumed the app GUI would tell you about something as important as that.

ghost:
Purpose
Fix the issue for root users by temporarily increasing "fs.inotify.max_user_watches" to 128K (instead of 8K).

Related issue:
#1208 "Syncthing is running out of inotify watches"
Forum discussion: https://forum.syncthing.net/t/android-failed-to-set-up-inotify-handler-please-increase-inotify-limits/11924/15

Testing
Verified working running syncthing rooted on Android 7.1.2 on device lg-h815 at commit https://github.com/syncthing/syncthing-android/commit/ef5729d2c8f27d31182f7d61e0666797a3a22870 .

Tests
a) Running non-rooted
image

b) Running rooted
image

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -125,6 +125,14 @@
                 <action android:name="com.nutomic.syncthingandroid.action.STOP" />
             </intent-filter>
         </receiver>
+        <activity android:name=".activities.FolderTypeDialogActivity"
+            android:label="@string/folder_type"
+            android:theme="@style/Theme.Syncthing.Dialog"
+            android:parentActivityName=".activities.FolderActivity">
+            <meta-data
+                android:name="android.support.PARENT_ACTIVITY"
+                android:value="com.nutomic.syncthingandroid.activities.FolderActivity" />
+        </activity>
         <activity android:name=".activities.PullOrderDialogActivity"
             android:label="@string/pull_order"
             android:theme="@style/Theme.Syncthing.Dialog"
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
@@ -77,6 +77,7 @@ public class FolderActivity extends SyncthingActivity
 
     private static final int FILE_VERSIONING_DIALOG_REQUEST = 3454;
     private static final int PULL_ORDER_DIALOG_REQUEST = 3455;
+    private static final int FOLDER_TYPE_DIALOG_REQUEST =3456;
     private static final int CHOOSE_FOLDER_REQUEST = 3459;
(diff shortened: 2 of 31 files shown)


imsodin (commented):
Looks reasonable, can't say whether it's the right place/way to do this.

Catfriend1:
At least it doesn't break anything for the "normal user". I even can't imagine this to be bad for root users as if the "sysctl" command may be unsupported in the future, it just will return an error exit code on the shell, not getting executed and the code being logged to logcat.

AudriusButkevicius (approved):

PR #1213, opened 2018-08-08
Keep Wi-Fi naming consistent

licaon-kter:


Catfriend1:
Approved. Good point, I also noticed that. Thanks for helping.

PR #1243, opened 2018-11-19
build: add gradle.properties to select java version 8

imsodin:
Apparently the current default java version on my system is 10, and the resulting error from gradle is totally clear:

```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring project ':app'.
> java.lang.NullPointerException (no error message)

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 1s
```

In order to spare other new (and not java-versed) contributors this headache, lets just specify this version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ The project is translated on [Transifex](https://www.transifex.com/projects/p/sy
 - Android SDK (you can skip this if you are using Android Studio)
 - Android NDK (`$ANDROID_NDK_HOME` should point at the root directory of your NDK)
 - Go (see [here](https://docs.syncthing.net/dev/building.html#prerequisites) for the required version)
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
And what if I am on Windows or my java is in a different location?

imsodin:
... and there really seems to be no way (except parsing `java -version` or some similar BS) to give a meaningful error - every journey into java/android land is such a dream. I'll settle for a readme note then.

Nutomic:
Looks like we just have to update gradle. Will do this later.

https://www.infoq.com/news/2018/05/gradle-releases-version-4.7

Nutomic:
I guess Java 10 is not supported by the Android build tools, I was still getting errors after upgrading everything.

PR #1244, opened 2018-11-20
DO NOT MERGE Release 0.10.15

imsodin:
This PR just exists solely to check that I didn't mess up, it should not be squash merged (I'll probably have to push directly to master for a clean log). Someone please just acknowledge that this is fine, then I'll do that and also add the appropriate tag.

The only change compared to 0.10.15-beta1 are imported translations.


AudriusButkevicius:
Looks ok, even though it says it's a branch not off of master.

PR #1264, opened 2019-01-09
Show reasons for disabled service in notification dialog (fixes #1263)

### LINKED ISSUE #1263: Run conditions: Tell user which applies/inhibits the startup [FEATURE REQUEST]
When the `"Syncthing is disabled - Would you like to change the run conditions?" [translated]`  popup appears, it would be really nice to see *which* condition(s) inhibit(s) the start.

capi:
This implementation aims to change the actual run-condition check as little as possible, but returns a set of reasons instead of a single boolean. Since the original implementation returns early in several cases, this early returns are simulated but still attempt to collect all applying reasons.

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/model/RunConditionCheckResult.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/model/RunConditionCheckResult.java
@@ -1,0 +1,78 @@
+package com.nutomic.syncthingandroid.model;
+
+import com.nutomic.syncthingandroid.R;
+
+import java.util.Collections;
+import java.util.List;
+
+public class RunConditionCheckResult {
+
+    public enum BlockerReason {
+        ON_BATTERY(R.string.syncthing_disabled_reason_on_battery),
+        ON_CHARGER(R.string.syncthing_disabled_reason_on_charger),
+        POWERSAVING_ENABLED(R.string.syncthing_disabled_reason_powersaving),
+        GLOBAL_SYNC_DISABLED(R.string.syncthing_disabled_reason_android_sync_disabled),
+        WIFI_SSID_NOT_WHITELISTED(R.string.syncthing_disabled_reason_wifi_ssid_not_whitelisted),
+        WIFI_WIFI_IS_METERED(R.string.syncthing_disabled_reason_wifi_is_metered),
+        NO_NETWORK_OR_FLIGHTMODE(R.string.syncthing_disabled_reason_no_network_or_flightmode),
+        NO_MOBILE_CONNECTION(R.string.syncthing_disabled_reason_no_mobile_connection),
+        NO_WIFI_CONNECTION(R.string.syncthing_disabled_reason_no_wifi_connection),
+        NO_ALLOWED_NETWORK(R.string.syncthing_disabled_reason_no_allowed_method);
+
+        private final int resId;
+
+        BlockerReason(int resId) {
+            this.resId = resId;
+        }
+
+        public int getResId() {
+            return resId;
+        }
+    }
+
+    public static final RunConditionCheckResult SHOULD_RUN = new RunConditionCheckResult();
+
+    private final boolean mShouldRun;
+    private final List<BlockerReason> mBlockReasons;
+
+    /**
+     * Use SHOULD_RUN instead.
+     * Note: of course anybody could still construct it by providing an empty list to the other
+     * constructor.
+     */
+    private RunConditionCheckResult() {
+        this(Collections.emptyList());
+    }
+
+    public RunConditionCheckResult(List<BlockerReason> blockReasons) {
+        mBlockReasons = Collections.unmodifiableList(blockReasons);
+        mShouldRun = blockReasons.isEmpty();
+    }
+
+
+    public List<BlockerReason> getBlockReasons() {
+        return mBlockReasons;
+    }
+
+    public boolean isShouldRun() {
+        return mShouldRun;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        RunConditionCheckResult that = (RunConditionCheckResult) o;
+
+        if (mShouldRun != that.mShouldRun) return false;
+        return mBlockReasons.equals(that.mBlockReasons);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = (mShouldRun ? 1 : 0);
+        result = 31 * result + mBlockReasons.hashCode();
+        return result;
+    }
+}
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java
@@ -16,18 +16,20 @@ import android.os.BatteryManager;
 import android.os.Build;
 import android.os.PowerManager;
 import android.support.annotation.Nullable;
-import android.support.v4.content.LocalBroadcastManager;
 import android.util.Log;
 
-import com.google.common.collect.Lists;
 import com.nutomic.syncthingandroid.SyncthingApp;
-import com.nutomic.syncthingandroid.service.ReceiverManager;
+import com.nutomic.syncthingandroid.model.RunConditionCheckResult;
 
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
 import javax.inject.Inject;
+
+import static com.nutomic.syncthingandroid.model.RunConditionCheckResult.*;
+import static com.nutomic.syncthingandroid.model.RunConditionCheckResult.BlockerReason.*;
 
 /**
  * Holds information about the current wifi and charging state of the device.
@@ -52,7 +54,7 @@ public class RunConditionMonitor {
     };
 
     public interface OnRunConditionChangedListener {
-        void onRunConditionChanged(boolean shouldRun);
+        void onRunConditionChanged(RunConditionCheckResult result);
     }
 
     private final Context mContext;
@@ -60,14 +62,14 @@ public class RunConditionMonitor {
     private ReceiverManager mReceiverManager;
 
     /**
-     * Sending callback notifications through {@link OnDeviceStateChangedListener} is enabled if not null.
+     * Sending callback notifications through {@link OnRunConditionChangedListener} is enabled if not null.
      */
     private @Nullable OnRunConditionChangedListener mOnRunConditionChangedListener = null;
 
     /**
-     * Stores the result of the last call to {@link decideShouldRun}.
+     * Stores the result of the last call to {@link #decideShouldRun()}.
      */
-    private boolean lastDeterminedShouldRun = false;
+    private RunConditionCheckResult lastRunConditionCheckResult;
 
     public RunConditionMonitor(Context context, OnRunConditionChangedListener listener) {
         Log.v(TAG, "Created new instance");
@@ -140,21 +142,25 @@ public class RunConditionMonitor {
     }
 
     public void updateShouldRunDecision() {
-        // Check if the current conditions changed the result of decideShouldRun()
+        // Reason if the current conditions changed the result of decideShouldRun()
         // compared to the last determined result.
-        boolean newShouldRun = decideShouldRun();
-        if (newShouldRun != lastDeterminedShouldRun) {
+        RunConditionCheckResult result = decideShouldRun();
+        boolean change;
+        synchronized (this) {
+            change = lastRunConditionCheckResult == null || !lastRunConditionCheckResult.equals(result);
+            lastRunConditionCheckResult = result;
+        }
+        if (change) {
             if (mOnRunConditionChangedListener != null) {
-                mOnRunConditionChangedListener.onRunConditionChanged(newShouldRun);
+                mOnRunConditionChangedListener.onRunConditionChanged(result);
             }
-            lastDeterminedShouldRun = newShouldRun;
         }
     }
 
     /**
      * Determines if Syncthing should currently run.
      */
-    private boolean decideShouldRun() {
+    private RunConditionCheckResult decideShouldRun() {
         // Get run conditions preferences.
         boolean prefRunOnMobileData= mPreferences.getBoolean(Constants.PREF_RUN_ON_MOBILE_DATA, false);
         boolean prefRunOnWifi= mPreferences.getBoolean(Constants.PREF_RUN_ON_WIFI, true);
@@ -166,18 +172,20 @@ public class RunConditionMonitor {
         boolean prefRespectPowerSaving = mPreferences.getBoolean(Constants.PREF_RESPECT_BATTERY_SAVING, true);
         boolean prefRespectMasterSync = mPreferences.getBoolean(Constants.PREF_RESPECT_MASTER_SYNC, false);
 
+        List<BlockerReason> blockerReasons = new ArrayList<>();
+
         // PREF_POWER_SOURCE
         switch (prefPowerSource) {
             case POWER_SOURCE_CHARGER:
                 if (!isCharging()) {
                     Log.v(TAG, "decideShouldRun: POWER_SOURCE_AC && !isCharging");
-                    return false;
+                    blockerReasons.add(ON_BATTERY);
                 }
                 break;
             case POWER_SOURCE_BATTERY:
                 if (isCharging()) {
                     Log.v(TAG, "decideShouldRun: POWER_SOURCE_BATTERY && isCharging");
-                    return false;
+                    blockerReasons.add(ON_CHARGER);
                 }
                 break;
             case POWER_SOURCE_CHARGER_BATTERY:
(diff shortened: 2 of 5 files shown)


app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java
```diff
@@ -142,19 +140,27 @@ public void onReceive(Context context, Intent intent) {
...
             }
-            lastDeterminedShouldRun = newShouldRun;
         }
     }
 
     /**
      * Determines if Syncthing should currently run.
      */
-    private boolean decideShouldRun() {
+    private Set<String> decideShouldRun() {
```
AudriusButkevicius: Let's create a auxilary class / DTO / POJO whatever the hell, and potentially add reasons for running and a boolean?
I guess reasons can be both positive and negative, so we could store both and imply that the top one is the reason for our current state, which then allows people to understand the order of positives and negatives.

capi: I considered this the first time, but since the run logic had some issues previously I just didn't want to change too much to it. At the moment it aims to just be the simplest solution to the request without actually changing the logic. Future improvements like you said would be sensible, I just didn't want to do it as first step.

AudriusButkevicius: Well the DTO would only have the stack of messages, that doesn't change the logic, and I think that's still valuable to understand where the issues are (when reading the stack of messages people post)

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingService.java
```diff
@@ -258,7 +263,20 @@ public int onStartCommand(Intent intent, int flags, int startId) {
      * function is called to notify this class to run/terminate the syncthing binary.
      * {@link #onServiceStateChange} is called while applying the decision change.
      */
-    private void onUpdatedShouldRunDecision(boolean newShouldRunDecision) {
+    private void onUpdatedShouldRunDecision(Set<String> reasons) {
+        boolean newShouldRunDecision = reasons.isEmpty();
+        boolean reasonsChanged;
+        synchronized (mStateLock) {
```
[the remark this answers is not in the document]
capi: True, it should be `mCurrentState` thanks for catching that.

app/src/main/res/values-de/strings.xml
```diff
@@ -404,6 +404,18 @@ Bitte melden Sie auftretende Probleme via GitHub.</string>
   <string name="syncthing_disabled_change_settings">Einstellungen ändern</string>
   <!--Button text on the "syncthing disabled" dialog, used as menu item to stop syncthing service if "always_run_in_background" is true-->
   <string name="exit">Beenden</string>
+  <!-- Reasons for "syncthing disabled" dialog -->
```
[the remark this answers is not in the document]
capi: Yeah, I was afraid of something like that. I checked the history of the file and saw manual commits outside of transifex so I hoped it would be OK. I'll simply remove the file from the push and see that it'll be translated on transifex instead.

capi:
I've changed the changes requested by @Catfriend1. @AudriusButkevicius is right about the DTO, but I'll need some more time for this, hopefully later tonight. Using an immutable DTO would also remove the need for all the synchronization and make the code clearer, so yes, I'll change it to that.

app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java
```diff
@@ -166,18 +172,23 @@ private boolean decideShouldRun() {
         boolean prefRespectPowerSaving = mPreferences.getBoolean(Constants.PREF_RESPECT_BATTERY_SAVING, true);
         boolean prefRespectMasterSync = mPreferences.getBoolean(Constants.PREF_RESPECT_MASTER_SYNC, false);
 
+        Set<String> blockerReasons = new TreeSet<>(); // TreeSet so that orders are sorted
```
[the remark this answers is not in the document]
capi: Yes, in the DTO that was requested I'll do exactly this.

app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java
```diff
@@ -166,18 +172,23 @@ private boolean decideShouldRun() {
...
+        Boolean resultForced = null; // need to remember if we encountered a return in previous version
```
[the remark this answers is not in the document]
capi: The final result is a `boolean`, the problem is, that the initial code returns at a few stages, so I can set the final result only, if no other return would have returned previously. Boolean allows me to check if there would have been already a return, and only set the final result if the existing code wouldn't have already returned.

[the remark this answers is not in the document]
Nutomic: Yeah but putting three values in a boolean like this seems really confusing for me.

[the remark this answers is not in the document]
capi: I hope you agree with my new solution that avoids using this at all, together with a DTO.

capi:
So, I now force-pushed the draft that uses an enum and a DTO for the reasons. I'd appreciate feedback.

app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java
```diff
@@ -16,19 +16,31 @@
...
+import static com.nutomic.syncthingandroid.model.RunConditionCheckResult.Reason.GLOBAL_SYNC_DISABLED;
+import static com.nutomic.syncthingandroid.model.RunConditionCheckResult.Reason.NO_ALLOWED_NETWORK;
+import static com.nutomic.syncthingandroid.model.RunConditionCheckResult.Reason.NO_MOBILE_CONNECTION;
+import static com.nutomic.syncthingandroid.model.RunConditionCheckResult.Reason.NO_NETWORK_OR_FLIGHTMODE;
+import static com.nutomic.syncthingandroid.model.RunConditionCheckResult.Reason.NO_WIFI_CONNECTION;
+import static com.nutomic.syncthingandroid.model.RunConditionCheckResult.Reason.ON_BATTERY;
+import static com.nutomic.syncthingandroid.model.RunConditionCheckResult.Reason.ON_CHARGER;
+import static com.nutomic.syncthingandroid.model.RunConditionCheckResult.Reason.POWERSAVING_ENABLED;
+import static com.nutomic.syncthingandroid.model.RunConditionCheckResult.Reason.WIFI_SSID_NOT_WHITELISTED;
+import static com.nutomic.syncthingandroid.model.RunConditionCheckResult.Reason.WIFI_WIFI_IS_METERED;
```
Nutomic: I think you can import the enum as a whole, and not list every item separately? Not sure, my Java is a bit rusty.

Otherwise looks good 👍

capi: Could be, it was generated that way by Android Studio. I'll check.

capi:
I suppose the failed TeamCity build is due to the changes of the Go version in the docker container?

app/src/main/java/com/nutomic/syncthingandroid/model/RunConditionCheckResult.java
```diff
@@ -0,0 +1,69 @@
+package com.nutomic.syncthingandroid.model;
+
+import com.google.common.base.Objects;
+import com.nutomic.syncthingandroid.R;
+
+import java.util.Collections;
+import java.util.List;
+
+public class RunConditionCheckResult {
```
Catfriend1: This new DTO is very nice, good work :-)

app/src/main/java/com/nutomic/syncthingandroid/model/RunConditionCheckResult.java
```diff
@@ -0,0 +1,69 @@
...
+            this.resId = resId;
+        }
+
+        public int getResId() {
+            return resId;
+        }
+    }
+
+    private final boolean mShouldRun;
+    private List<Reason> mBlockReasons;
```
capi: This should be `final` as well.

capi:
I've added the final remarks of @Catfriend1 and @Nutomic.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingService.java
```diff
@@ -578,6 +595,30 @@ private void onServiceStateChange(State newState) {
...
+
+    private void onRunConditionCheckResultChange(RunConditionCheckResult result) {
+        mHandler.post(() -> {
+            for (Iterator<OnRunConditionCheckResultListener> i = mOnRunConditionCheckResultListeners.iterator();
+                 i.hasNext(); ) {
+                OnRunConditionCheckResultListener listener = i.next();
+                if (listener != null) {
+                    listener.onRunConditionCheckResultChanged(result);
+                } else {
+                    i.remove();
```
AudriusButkevicius: Why remove here?

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingService.java
```diff
@@ -262,7 +269,17 @@ public int onStartCommand(Intent intent, int flags, int startId) {
      * function is called to notify this class to run/terminate the syncthing binary.
      * {@link #onServiceStateChange} is called while applying the decision change.
      */
-    private void onUpdatedShouldRunDecision(boolean newShouldRunDecision) {
+    private void onUpdatedShouldRunDecision(RunConditionCheckResult result) {
+        boolean newShouldRunDecision = result.isShouldRun();
+        boolean reasonsChanged;
+        synchronized (mCurrentCheckResult) {
```
AudriusButkevicius: Why do you need to synchronize on an atomic? That makes no sense.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingService.java
```diff
@@ -262,7 +269,17 @@ public int onStartCommand(Intent intent, int flags, int startId) {
...
+            reasonsChanged = !mCurrentCheckResult.get().equals(result);
+            mCurrentCheckResult.set(result);
```
AudriusButkevicius:
```
changed = !mCurrentCheckResult.getAndSet(result).equals(result)
```

app/src/main/java/com/nutomic/syncthingandroid/model/RunConditionCheckResult.java
```diff
@@ -0,0 +1,68 @@
...
+            return resId;
+        }
+    }
+
+    private final boolean mShouldRun;
+    private final List<BlockerReason> mBlockReasons;
+
+    public RunConditionCheckResult() {
+        mBlockReasons = Collections.emptyList();
+        mShouldRun = true;
```
AudriusButkevicius: `this(Collections.emptyList())`?

app/src/main/java/com/nutomic/syncthingandroid/model/RunConditionCheckResult.java
```diff
@@ -0,0 +1,68 @@
...
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        RunConditionCheckResult that = (RunConditionCheckResult) o;
+        return mShouldRun == that.mShouldRun && mBlockReasons.size() == that.mBlockReasons.size()
+                && mBlockReasons.containsAll(that.mBlockReasons);
+    }
+
+    @Override
+    public int hashCode() {
+        return mShouldRun ? 17 : 42;
```
AudriusButkevicius: why not Object.hashCode(mShouldRun) or something like that?

app/src/main/java/com/nutomic/syncthingandroid/model/RunConditionCheckResult.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        RunConditionCheckResult that = (RunConditionCheckResult) o;
+        return mShouldRun == that.mShouldRun && mBlockReasons.size() == that.mBlockReasons.size()
+                && mBlockReasons.containsAll(that.mBlockReasons);
+    }
```
AudriusButkevicius: IntelliJ (and potentially Android studio) can generate both hashcode and equals whcih has a better implementation.
It seems strange that equality takes reasons into account, and hash code does not.

app/src/main/java/com/nutomic/syncthingandroid/model/RunConditionCheckResult.java
capi: Will do.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingService.java
capi: Because I want consistency over the compare and assignment. I was thinking the same, but otherwise the assignment will race.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingService.java
capi: Yes, but in this case we assign the result even if not changed. On the other hand since it is immutable, it doesn't hurt.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingService.java
capi: To be honest, I just copied the existing method and did not think too much about it.

capi:
I'm skiing for a week, once I'm back I'll address the remaining points. Thanks for taking all the time for the review.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingService.java
Catfriend1: If an activity that registered a listener is unavailable being set to null, it's automatically cleant up. Nutomic's code has this all over the place.

capi:
I've rebased to latest master and I've addressed the reviews of @AudriusButkevicius. I reworked the `hashCode()` implementation (the last one was fulfilling the contract, but I agree that it looked strange). I still need to keep my own implementations since the generated versions don't do a deep compare of the list.
Also I made the empty constructor private and introduced a `SHOULD_RUN` static final field, since using this makes it more readable in my opinion.
Also I removed the synchronization of the `AtomicReference` by using the suggested `getAndSet()` approach.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingService.java
capi: I've not addressed this one, since it's consistent with the other code. If it really is a blocker, I can of course adjust.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingService.java
capi: Fixed by your suggestion of using `getAndSet()` instead.

AudriusButkevicius:
I am still not sure I understand why we need magical hash and equals, why we can't use IDE generated ones.

capi:
The one it did for me, it did not. And I also just noticed that I broke it (while me last version was not optimal but technically correct). So **PLEASE DO NOT MERGE** until I fixed it.

I'll double-check tonight, but last time I let it generate equals and hash-code it used equals on the list, which does not do a deep comparison of the list content. And honestly I think it can't, because in this case the hash would not be stable as soon as the list changed. You can only do a deep compare and use it for hash if the list itself is immutable.

AudriusButkevicius:
Right, I guess if the order should be ignored, the equals should do a set comparison?

At which point I guess you do need a custom equals and hashcode.

capi:
@AudriusButkevicius: I have thought about it, you are right. Since I explicitly use a `List` and not a `Set`, I should not suddenly use `Set` semantics for `equals()` and `hashCode()`, therefore I now use the InteliJ-generated implementation.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingService.java
AudriusButkevicius: I understand the value gets gc'ed and becomes null, but how given there is a pointer in this set, which should prevent collection?

If we had a weakref, I'd understand it, but we don't have a weakref, we have a strong ref.

app/src/main/java/com/nutomic/syncthingandroid/model/RunConditionCheckResult.java
capi: You are looking at the old version. This is already outdated by my latest push. This version has been replaced by the InteliJ version.

AudriusButkevicius:
```
    @Override
    public boolean equals(Object o) {
        if (this == o) return true;
        if (o == null || getClass() != o.getClass()) return false;

        RunConditionCheckResult that = (RunConditionCheckResult) o;

        if (mShouldRun != that.mShouldRun) return false;
        return mBlockReasons.equals(that.mBlockReasons);
    }

    @Override
    public int hashCode() {
        int result = (mShouldRun ? 1 : 0);
        result = 31 * result + mBlockReasons.hashCode();
        return result;
    }
```

is what I see. Usually IntelliJ generates `Object.equals(x, y) && ...` etc, and `Object.hash(o) + Object.hash(y)`

capi: I needed to the revert to the default generated version (you can change that in the generate dialog), since our target API level does not have `Object.hash(o)` available. This is available from API 19, we are still targeting API 16.

PR #1270, opened 2019-02-11
Support Variable Size Blocks (VSB)

### LINKED ISSUE #153: Add-Repo bug
When using the android-GUI I cannot klick on CREATE. I can give the repo a name, chose a folder on my Device (except external SD, but I know that you work on that) and all that stuff. But klicking on CREATE does nothing at all

Version 0.5.1
Android 4.4.4

ghost:
Purpose:
* Update folder model defaults
* UI for Variable Size Blocks (VBS) (fixes #153)

Testing:
Verified working on AVD 9.x at https://github.com/syncthing/syncthing-android/commit/efef5d04cbb18c858d9bec1e9d8b18feaa9080c3 .

Backport from https://github.com/Catfriend1/syncthing-android/pull/156

Screenshot:
image


app/src/main/java/com/nutomic/syncthingandroid/model/Folder.java
```diff
@@ -37,7 +37,7 @@
     public boolean disableSparseFiles;
     public boolean disableTempIndexes;
     public boolean paused;
-    public boolean useLargeBlocks;
+    public boolean useLargeBlocks = false;
     public int weakHashThresholdPct = 25;
```
AudriusButkevicius: Why default false? I think go now has defaults set to true?

Catfriend1: I recall this is still unreleased and planned for 1.1.0, so I set it to false until we need it.

AudriusButkevicius (dismissed):

app/src/main/java/com/nutomic/syncthingandroid/model/Folder.java
imsodin: AFAIK the most recent released beta of 0.10.17 contains 1.0.1, so this change will make it into 0.10.18, which will be released with 1.1.0 - so defaulting to true as well should be good.

AudriusButkevicius: True

AudriusButkevicius (approved):

PR #1271, opened 2019-02-12
Preserve xml encoding on read, parse and save (fixes #1234)

### LINKED ISSUE #1234: Emoji replaced by replacement characters (�) in machine name
Some of my Syncthing instances have emoji in their machine names (e.g. "🦑", v0.14.50 macOS; "🐛", v0.14.49, FreeBSD).

When adding these instances to the Syncthing Android app (app version 0.10.14, syncthing version v0.14.50, LineageOS 14.1) the machine names are at first correctly detected and shown with emoji:

before

However, when the app is closed and subsequently re-opened, the emoji are replaced by UTF-8 replacement characters:

after

The emoji work on all my other platforms (macOS, Windows, FreeBSD, Linux) but aren't shown properly in the Android app.

### Version Information

    App Version: 0.10.14
    Syncthing Version: v0.14.50
    Android Version: LineageOS 14.1

ghost:
Purpose:
* Preserve xml encoding on read, parse and save

Backport of https://github.com/Catfriend1/syncthing-android/pull/89

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
@@ -18,11 +18,19 @@ import org.w3c.dom.Element;
 import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
 import org.xml.sax.SAXException;
+import org.xml.sax.InputSource;
 
 import java.io.BufferedReader;
+import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.InputStreamReader;
 import java.io.IOException;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.OutputStreamWriter;
+import java.io.UnsupportedEncodingException;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.Locale;
@@ -34,6 +42,7 @@ import javax.inject.Inject;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
+import javax.xml.transform.OutputKeys;
 import javax.xml.transform.Transformer;
 import javax.xml.transform.TransformerException;
 import javax.xml.transform.TransformerFactory;
@@ -95,14 +104,19 @@ public class ConfigXml {
             throw new OpenConfigException();
         }
         try {
+            FileInputStream inputStream = new FileInputStream(mConfigFile);
+            InputStreamReader inputStreamReader = new InputStreamReader(inputStream, "UTF-8");
+            InputSource inputSource = new InputSource(inputStreamReader);
+            inputSource.setEncoding("UTF-8");
             DocumentBuilder db = DocumentBuilderFactory.newInstance().newDocumentBuilder();
-            Log.d(TAG, "Trying to read '" + mConfigFile + "'");
-            mConfig = db.parse(mConfigFile);
+            Log.d(TAG, "Parsing config file '" + mConfigFile + "'");
+            mConfig = db.parse(inputSource);
+            inputStream.close();
+            Log.i(TAG, "Successfully parsed config file");
         } catch (SAXException | ParserConfigurationException | IOException e) {
-            Log.w(TAG, "Cannot read '" + mConfigFile + "'", e);
+            Log.w(TAG, "Failed to parse config file '" + mConfigFile + "'", e);
             throw new OpenConfigException();
         }
-        Log.i(TAG, "Loaded Syncthing config file");
     }
 
     public URL getWebGuiUrl() {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -342,14 +356,34 @@ private void saveChanges() {
...
+
+            // Prepare Object-to-XML transform.
             TransformerFactory transformerFactory = TransformerFactory.newInstance();
             Transformer transformer = transformerFactory.newTransformer();
-            DOMSource domSource = new DOMSource(mConfig);
-            StreamResult streamResult = new StreamResult(mConfigTempFile);
-            transformer.transform(domSource, streamResult);
+            transformer.setOutputProperty(OutputKeys.METHOD, "xml");
+            transformer.setOutputProperty(OutputKeys.ENCODING, "UTF-16");
+            transformer.setOutputProperty(OutputKeys.OMIT_XML_DECLARATION, "yes");
```
AudriusButkevicius: Does thos option behave incorrectly hence you need to write the header manually?

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -342,14 +356,34 @@ private void saveChanges() {
...
+            transformer.setOutputProperty(OutputKeys.INDENT, "no");
+
```
AudriusButkevicius: Is there a reason this is needed? It's nice to have the config readable.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: Yes, that's correct.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: I preferred to have the config readable as I often use adb pull config.xml during debug sessions.

AudriusButkevicius: But this says 'no' to indent, which I assume means not to indent it, which makes it a single long line?

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
AudriusButkevicius: What part is missing in the header with that option?

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: When I made a load and save test the indent was preserved like it was before.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: If you don't write the header manually and change the encoding syncthing cannot start. Seems to be a bug in the xml library.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
AudriusButkevicius: And if you remove this line all together, does it still have indentation? 

Reason I am asking is because for the next reader it will be completely not obvious why this line is here when it's not respected and the original intent is not clear either.

If it still does what it needs without this line, let's just remove it.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
AudriusButkevicius: Sure. I am just asking how the header differs between what we are writing and what the library writes, so I could understand what the problem is.

Potentially it could also be explained in a comment above for any future reader.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: I did this months ago and tested as long until I found a way where the transformation is still working when the encoding was changed. I'll check this when I have to update the ConfigXml next time in the fork and report back. Maybe I can do the work once instead of twice like this to find the optimal way.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: The library still would write UTF-8 so parsing UTF-16 stuff would throw an error.

AudriusButkevicius: I will defer to @imsodin, @nutomic and @imsodin for approval, as I simply don't understand how:

1. Reading as UTF-8
2. Writing a header as UTF-8
3. Writing data as supposedly UTF-16 (yet lying in the header of the enconding being utf-8 which I presume is default anyway)
4. Setting indentation to no which however supposedly formats the data correctly

fixes this issue. 

I don't dispute it works, but I want to understand why settings are what they are and why this works because of them.

If others feel they understand this as it is, or if they don't think answers to this are necessery they can feel free to approve and merge.

As a minimum I feel this could be simplified (using built in header writer as I think it can be mended via options to set the right output, setting the encoding output to utf-8, setting indentation to true)

Nutomic: I agree with Audrius, this code looks weird and overly complicated. But I havent done Android development in over a year, and I dont have the time to dig into this myself. 

That said, I wouldnt be completely opposed to merging it, if it works and you can add some comments explaining why you do this. Maybe with links to stackoverflow or wherever you found the solution.

Perhaps there is also a library that handles this better?

PR #1281, opened 2019-02-26
Fallback to http if https with TLS 1.2 is unavailable

ghost:
Purpose / Discussion:
https://forum.syncthing.net/t/android-and-tls1-2/12931/16

Testing:
- Very "quickly" tested on AVD 9.x and 7.0 and verified working.

Note:
This code was backported from the Fork. It solves the problem, but in a future contribution, we should try getSupportedProtocols() as suggested by user Bt90 (see forum).

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/http/ApiRequest.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/http/ApiRequest.java
@@ -157,9 +157,12 @@ public abstract class ApiRequest {
         }
         @Override
         protected HttpURLConnection createConnection(URL url) throws IOException {
-            HttpsURLConnection connection = (HttpsURLConnection) super.createConnection(url);
-            connection.setHostnameVerifier((hostname, session) -> true);
-            return connection;
+            if (mUrl.toString().startsWith("https://")) {
+                HttpsURLConnection connection = (HttpsURLConnection) super.createConnection(url);
+                connection.setHostnameVerifier((hostname, session) -> true);
+                return connection;
+            }
+            return super.createConnection(url);
         }
     }
 
--- a/app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
@@ -106,8 +106,9 @@ public class ConfigXml {
     }
 
     public URL getWebGuiUrl() {
+        String urlProtocol = Constants.osSupportsTLS12() ? "https" : "http";
         try {
-            return new URL("https://" + getGuiElement().getElementsByTagName("address").item(0).getTextContent());
+            return new URL(urlProtocol + "://" + getGuiElement().getElementsByTagName("address").item(0).getTextContent());
         } catch (MalformedURLException e) {
             throw new RuntimeException("Failed to parse web interface URL", e);
         }
@@ -154,9 +155,15 @@ public class ConfigXml {
         }
 
         /* Section - GUI */
-        // Enforce TLS.
         Element gui = getGuiElement();
-        changed = setConfigElement(gui, "tls", "true") || changed;
+
+        // Platform-specific: Force REST API and Web UI access to use TLS 1.2 or not.
+        Boolean forceHttps = Constants.osSupportsTLS12();
+        if (!gui.hasAttribute("tls") ||
+                Boolean.parseBoolean(gui.getAttribute("tls")) != forceHttps) {
+            gui.setAttribute("tls", forceHttps ? "true" : "false");
+            changed = true;
+        }
 
         // Set user to "syncthing"
         changed = setConfigElement(gui, "user", "syncthing") || changed;
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/Constants.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/Constants.java
@@ -1,6 +1,7 @@
 package com.nutomic.syncthingandroid.service;
 
 import android.content.Context;
+import android.os.Build;
 import android.os.Environment;
 
 import java.io.File;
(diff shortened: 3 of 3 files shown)


imsodin:
You should base your changes on master, not on 0.10.18-beta2-release, i.e. not include Audrius commits in this PR. I can take care of merging the changes from that branch.

Catfriend1:
Ahmmm...yes. I've already resolved the conflicts so it needs to be rebased on his branch. I'd suggest we merge his branch. There just one single thing I don't understand: the go111enabled var.

imsodin:
Let me sort the master / release branch business out. I'll try to adopt the strategy used on syncthing/syncthing, described by Jakob like this (think beta instead of rc for now):

> In the repo, a new release branch will be created at the time the rc is cut, and the commit is tagged with the rc version tag. If bugfixes need to happen they are cherry-picked from master to release and a new rc tag created. Finally, whatever is the last commit on release is tagged with the real version tag, merge into master, and the release branch is deleted. This is mostly bureaucracy for the release maintainer, i.e., me at the moment.

See #1282

imsodin:
@Catfriend1 Are you ok with me rebasing your branch to clean up the commit/branch mess (in case you checked that tickbox so I have permissions to do so)?

Catfriend1:
Sure, go on :-). Let me know if sth is unclear to resolve.

app/src/main/java/com/nutomic/syncthingandroid/http/ApiRequest.java
```diff
@@ -175,4 +175,37 @@ private SSLSocketFactory getSslSocketFactory() {
             return null;
         }
     }
+
+    /**
+     * Extends {@link HurlStack}, uses {@link #getSslSocketFactory()} and disables hostname
+     * verification.
+     */
+    private class NetworkStack extends HurlStack {
+
```
imsodin: spurious empty line

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
```diff
@@ -154,9 +155,15 @@ public void updateIfNeeded() {
...
         /* Section - GUI */
-        // Enforce TLS.
         Element gui = getGuiElement();
-        changed = setConfigElement(gui, "tls", "true") || changed;
+
+        // Platform-specific: Force REST API and Web UI access to use TLS 1.2 or not.
+        Boolean forceHttps = Constants.osSupportsTLS12();
+        if (!gui.hasAttribute("tls") ||
+                Boolean.parseBoolean(gui.getAttribute("tls")) != forceHttps) {
+            gui.setAttribute("tls", forceHttps ? "true" : "false");
```
imsodin: Just ` gui.setAttribute("tls", forceHttps)`

imsodin (commented):
<strike>What are the changes to ApiRequest.java for? Naively it looks like the other changes are enough to enable/disablt TLS depending on android version to me.</strike> Never mind, I f***ed up merging.

imsodin:
@Catfriend1 I might have finally got it right, but please check that I didn't do any obvious mistakes.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: Would work, but we miss detecting the change then and according to "changed" we need to save the config to disk.

Catfriend1:
Thanks for merging it with the beta branch. Had a look on github mobile and looks fine so far now. To make sure, someone or me should run the generated APK on an Android 7 and non Anddoid 7 phone , checking the logs of ApiRequest if http / https appears.

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
imsodin: I don't mean the entire block, just that `forceHttps ? "true" : "false"`. However I didn't notice the quotes around true/false: Seems like `setAttribute` takes a string not a bool. That seems quite bad to me, but unrelated to this PR.

Catfriend1: Yes, In wonder if we should have constants for that as it hasn't to match Java but what SyncthingNative expects when reading the config.

imsodin: Are `setConfigElement(gui, ...` (before) and `gui.setAttribute(...` (now) equivalent?

imsodin:
I rebased the changes just to make sure github doesn't think I had any part in authoring this due to my merge commits.

In general looks good - a confirmation would be nice indeed. I asked Nummer378 on the forum, they seemed to have a test setup ready: https://forum.syncthing.net/t/android-and-tls1-2/12931/25

app/src/main/java/com/nutomic/syncthingandroid/util/ConfigXml.java
Catfriend1: No because during fork dev I discovered a problem here as in very early syncthing's it was an xml node and now is an Attribute of the xml tag. So setAttribute is right.

hawson:
I have an Android 4.4.4 device.  I believe that I am affected by the TLS issue using the mainline Syncthing app (0.10.18-beta2).  The Syncthing-Fork App (1.0.1.8) appears to work.

Catfriend1:
Should I still adjust sth @imsodin ?

imsodin:
From my side not, just waiting for confirmation. And a review by someone knowing java wouldn't hurt, though I guess the change is simple enough for my "hoping-for-wysiwyg-style" review.

GermanCoding:
Having a look at this right now, just a few minutes.

imsodin (approved):

PR #1282, opened 2019-02-26
Fast forward master to current release branch

imsodin:
@AudriusButkevicius Could you confirm that I don't break anything by merging this to master, then merging #1281 (once ready, not part of the breaking question ;) ) and proceeding from there for v0.10.18-beta3.


imsodin:
Uh ah sorry, I should have been more clear: I intended to merge the branch/fast forward, not squash merge to keep the git history as clean as possible. Well I guess it doesn't make much difference.

PR #1284, opened 2019-02-28
0.10.18-beta4

imsodin:
For those who haven't noticed: I just messed up my first try of a release, that's why this is beta4.


Catfriend1:
Thanks for your help on the release.

PR #1287, opened 2019-03-08
Update .gitignore to ease work on both projects

ghost:
Purpose:
- Ease between switching from forked branches to official branches

@imsodin Would you be okay to merge the .gitignore even though this is just a benefit for me at the moment? It wouldn't harm anything as the files that are additionally ignored don't exist / have no relevance in Syncthing Official Repo. We also could revert this any time easily.

Do you have another idea how I can .gitignore files only relevant for my local git when the .gitignore of the repo changes between branches? Currently I do a xcopy every time I need to switch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,5 +29,30 @@ project.properties
 gradle/wrapper/gradle/
 gradle/wrapper/gradlew*
 
+# Prebuilt-go
+syncthing/go
+syncthing/go.tgz
+syncthing/go.zip
+
+# Android NDK
+syncthing/android-ndk-r*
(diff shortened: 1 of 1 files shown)


imsodin:
Are you aware of `.git/info/exclude`? That allows you to exclude project specific stuff locally. So if these ignores aren't relevant to anyone but you, I'd put them there. If for some reasons you need them checked in, I wouldn't be opposed, I mean we also have intellij specific stuff in there.

Catfriend1:
That's okay for me :-). Thanks for the info.

PR #1290, opened 2019-03-13
Bumped version to 1.1.1-rc.1

imsodin:
This prepares the release of the current syncthing RC as 1.1.1-rc.1 (to beta track) as discussed in https://forum.syncthing.net/t/android-new-versioning-scheme/12929  
Should there be a note in whatsnew regarding the new versioning policy?

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -35,8 +35,8 @@ android {
         applicationId "com.nutomic.syncthingandroid"
         minSdkVersion 16
         targetSdkVersion 26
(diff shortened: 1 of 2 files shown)


imsodin:
They (docker container) just run `./gradlew --no-daemon buildNative lint assembleRelease publishRelease` and fdroid checks the version against `^[0-9.]+$` to build releases only, which still works.

Catfriend1:
Great! Good finding.:-)

PR #1295, opened 2019-03-25
Bumped version to 1.1.1-rc.2

imsodin:
I also imported translations, even though it's an RC update. I think adding/updating translations should on RCs is a good idea, as it means the final release has a chance of containing new translations.


imsodin:
Well, that would be an argument against doing it, as I don't want to introduce breakages during the beta phase, only fix them. However I hope such glitches are rare enough and update them anyway for reasons stated above :)

PR #1301, opened 2019-04-13
Adding dark theme (closes #653)

### LINKED ISSUE #653: FR: Dark theme
As it says in the title. A dark theme is veyr benecial to our eyes and to those with amoled screen.

camingus:
closes #653

Adds a dark theme option in addition to the light theme. No support for a black theme yet, but I'm sure that issue will be raised after this change gets merged.

Known issues:
* Icon colors don't change on API <20. Proposed solution: add a dependency on Android Support Libraries v23.2 or greater. This issue came up because I replaced all icons with vector graphics that use a color that automatically changes based on theme or whether or not the resource is enabled. Previously the project handled logic for enable/disabled explicitly. Reference: https://developer.android.com/studio/write/vector-asset-studio

* ~~Popups from the settings menu aren't readable in dark theme. This is a weird issue where the theme appears to partially apply to the settings menu. Everything works great until a popup is generated. When the dark theme is enabled, the theme changes the foreground color to white, but leaves the background white too. Android doesn't support explicitly theming the settings XML. Not sure how to fix this yet. There might be something in the Android Q (JetPack) library. Or we could use a third party settings solution.~~

Light Theme Screenshots:
Screenshot_20190413-100338
Screenshot_20190413-100348
Screenshot_20190413-100355

Dark Theme Screenshots
Screenshot_20190413-100408
Screenshot_20190413-100156
Screenshot_20190413-100206

Known issue 1 - Icon colors don't change automatically in API level <20:
Screenshot_1555175592

Known issue 2 - Settings pop ups use white foreground and background in dark theme:
Screenshot_20190413-100324


app/src/main/res/values/colors.xml
```diff
@@ -3,6 +3,7 @@
     <color name="primary">#03A9F4</color>
     <color name="primary_dark">#0288D1</color>
     <color name="accent">#FFC107</color>
+    <color name="icon_light">#ccffffff</color>
```
camingus: I didn't end up using this. I can remove it.

Catfriend1:
Could you please add some screen shots? Thanks for your contribution 🙂, I definitely have to try this out on my dev machine when coding next time. I've already done a lot of improvement work to fix glitches in the sub prefs screen in syncthing fork, it might be worth to have a look at it. Will post a link later... Sub prefs are generating a new view group that does not inherit things like toolbar from SyncthingActivity automatically.

Catfriend1:
At a first glance, clicking on settings/Theme shows the two options light and dark with none of both selected. We should have the default white theme checked here.

Catfriend1:
Sub-pref screens play nicely here on AVD 9.x:
image

Catfriend1:
Optional suggestion for improvement:
When on dark theme, the Web UI is still (Default) white themed. It would be convenient to the user, if we'd also change SyncthingNative's config.xml to the dark theme to avoid user questions like "BUG web UI remains white" with the usual answer "you've got to go into the web ui advanced settings to change that too".
image

camingus:
I've added screenshots to the original post. I'll look into the default on the theme setting next. And for setting the web UI, I think we can open a new issue for that one. I'm also not sure if there should be 1 app settings that controls both themes, or 2 app settings so users can choose different ones. I could see a user wanting a black web UI theme with a dark app theme. But also I would guess users either use one or the other. Is there any data on that?

camingus:
I've set the default check box to light theme.

Catfriend1:
Great :-). I've solved the riddles about the white popups. You could try adding /res/values-night instead of overriding AlertDialogBuilder. It also affects the popup dialogs from Android.

camingus:
I'm sorry. I don't quite follow the suggestion. I understand that I can make a themes.xml in values-night, but I'm not sure what the suggestion is for the contents of that file.

Catfriend1:
@camingus I've experimented a little bit, so I'm happy to assist you with this example: https://github.com/Catfriend1/syncthing-android/commit/10fd391f0851d400f08034348a7af7a598f6dad6

camingus:
Omg I spent so much time trying to figure out how to do this. But it was so easy! Thank you (:

camingus:
Hmmm. I broke the accent color on the button text... I need to figure out how to add that back.

camingus:
Okay my change above works on API 16, but not everything...

camingus:
Okay. At this point I think all that's left is the icon color on API level <20. Should we add a dependency on 'com.android.support:appcompat-v7:23.2.0' as part of this change?

app/src/main/java/com/nutomic/syncthingandroid/activities/DeviceActivity.java
```diff
@@ -378,7 +378,7 @@ private void showDeleteDialog(){
     }
 
     private Dialog createDeleteDialog(){
-        return new android.app.AlertDialog.Builder(this)
+        return new AlertDialog.Builder(this, R.style.Theme_Syncthing_Dialog)
```
Catfriend1: According to my own tests, using values-night allows you to omit this (in multiple places)

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -311,8 +315,8 @@ public void onServiceConnected() {
         public void onServiceStateChange(SyncthingService.State currentState) {
             mApi = mSyncthingService.getApi();
             boolean isSyncthingRunning = (mApi != null) &&
-                        mApi.isConfigLoaded() &&
-                        (currentState == SyncthingService.State.ACTIVE);
+                    mApi.isConfigLoaded() &&
+                    (currentState == SyncthingService.State.ACTIVE);
```
Catfriend1: I'd suggest keeping the diff as small as possible here as there's no change.

app/src/main/res/values/styles.xml
```diff
@@ -64,4 +64,5 @@
         <item name="android:paddingTop">5dp</item>
         <item name="android:paddingBottom">5dp</item>
     </style>
+
```
Catfriend1: Blank line:)

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
camingus: I honestly don't even know what happened here. This change doesn't even look better. I'll revert.

Catfriend1: If you're using android studio it may be auto reformat code being turned on.

app/src/main/java/com/nutomic/syncthingandroid/activities/DeviceActivity.java
camingus: Is that better? If we make a values-night we have to pick the colors. But with the DayNight theme we can let Android pick the colors. I agree that it can be easy to accidentally forget to include this when creating dialogs in the future, but I like the idea of Android choosing how dark the dark theme is.

Catfriend1: Both solutions seem valid to me. If you'd like to keep the parameter, I'd suggest wrapping it into one maintainable place like https://github.com/Catfriend1/syncthing-android/commit/89b6bba293a30a360bdb6947d30e257a8862d1f5

camingus: Okay it looks like we can compromise here. I don't think values-night is the magic here. It's just using the android:property in the xml. The fix I did for the settings menu pop ups seems to also apply to regular pop ups.

I think it still needs some tweaking though. Or we *should* still use the parameter. On my personal phone (API level 27) it works just fine. On my emulated API level 16 device the pop ups are dark themed but button text (again!!!) is the wrong color.

Catfriend1: I need to have a look at this on my devices to get a better impression... Thanks for your detailed testing so far :-)!

camingus: I think the button text color is probably not the end of the world. It's using the accent color (yellow) instead of the primary color (blue). Lots of things in the app are already yellow. I think I'm fine with that the way it is. And I don't think there's much I can do about it. android:colorAccent is available starting with API level 21 and the minimum API level Syncthing supports is 16. So can't use that. I'll add the method to the utils thing and refactor all of the calls to use it. It'll be pretty tiny though. The DayNight theme eliminates the need to check theme before picking (although this won't be true once the black theme feature request comes in). So it'll only be 1 line.

Catfriend1:
Yehaaa, that's cool:)

Catfriend1:
Hmmm ... Would it have any negative consequences to add your suggested support library? From my side it's fine. if we don't need it urgently and could do it in the existing v4 without putting too much extra work on it i would prefer that.

Btw: how did you get those xml pictures? I'm asking as I intend to merge this into Syncthing-Fork which would need more assets adjusted. In my first step I didn't merge any assets from here at all.

Catfriend1:
I've found a nice way to have all code in one place as I see your "Load theme" code in onCreate had to be duplicated among all activities not depending on SyncthingActivity.
See https://github.com/Catfriend1/syncthing-android/pull/399/commits/ba1dbad40b6e3f26de12a9b1a970cecef87d1680, https://github.com/Catfriend1/syncthing-android/pull/399/commits/a2d47ba6ab2eb06379a66fdd1ac0b9f7451d860c, https://github.com/Catfriend1/syncthing-android/pull/399/commits/2e8a2349c3baede10a58604b2ede06a80dc96021 on how to implement this.

camingus:
For adding the library, I think it's probably fine. It's part of the support libraries which seem to get added in pieces. Syncthing already has a dependency on 'com.android.support:design:27.0.2' which I think is basically a dependency on v27.0.2 of the support libraries. I can't add a dependency for app compat of a different version without also changing this version. No idea what the support libraries themselves depend on though. I'm guessing it's probably okay, but I would need to experiment. I'm just cautious about adding dependencies because of nightmares involving 2 different dependencies that themselves depend on two different versions of the same library.

In terms of the amount of work, it's probably more than it looks on the surface. But I would need to do it to be sure. According to the page on [Support Library 23.2](https://android-developers.googleblog.com/2016/02/android-support-library-232.html) we need to add the dependency, add a configuration setting, and update every place where reference a drawable in XML. And maybe Java too. I'm completely fine with tackling this as a separate Git issue.

For the vector graphics (XML pictures) I created them in 3 different ways. For the QR code I went to a [QR code generating website](https://www.qrcode-monkey.com/) and it created the SVG for me. And then I imported the SVG with Android Studio. New -> Vector Asset -> Local File. I also ended up manually tweaking the generated XML file. For the majority of the rest of the icons I used the vector asset thing in Android Studio as well, but instead of browsing from a local file it has a "Clip Art" option with common Android icons. I tried to replace all of the icons from there, but I couldn't find one for the tag text outline or the zip box. For those I made my own SVG files using Inkscape. I'm not a graphic designer so if they look underwhelming, that's why.

camingus:
I can create the issue (and the fix) for icon colors on <20 after we merge. Is there anything else to do here?

Catfriend1:
Looking at the AndroidManifest from your last commit, there were a lot of formatting changes. Could we please focus on the relevant lines of code to be on the diff.

camingus:
My apologies. Looks like Android Studio was helping again. Will fix.

camingus:
Sorry again. I didn't even know you had to update that file when adding an activity (so I'm kinda glad that Android Studio did it for me). I've reduced it back to the important changes which you can view here:

https://github.com/syncthing/syncthing-android/pull/1301/files#diff-c8cc2dd42271f2bf39c6aa81eb6a5529

Catfriend1:
Ok, thanks for your work and resolving the reviews. I think that's enough from my side, sure, I'd somehow like to see for example things that SettingsActivity restarts the whole app cleanly after a theme change or reflecting the theme change in config.xml (like it's already on the forks code) but.. eh.. you already put a lot of work in and did a good job which should first be honored. :-)
I'm personally a bit scared this merge would deviate official much from the fork - there would be large differences in resources as I already had problems with the two code bases and different translation string sets, now the assets would also make porting code forth and back harder.
To be honest: this has nothing to do with your work and I would agree merging this. Just wanted to point out what I think about getting the fork and official "together".
What do the others say? Should we improve this even more, merge it now or plan the time to merge this at a future point?

camingus:
For the extra changes that you mentioned above, please feel free to file issues if we decide to merge this without those and I'll word towards fixing it in future pull requests. I'm not an Android developer so I'm pretty naive about the right way to do things. So thanks for pointing that out.

If we want to tackle those as part of this change I'm fine with that too, but I just need to do some research on what it means to restart the whole app cleanly (every example I saw online used this code to restart the activity stack) and why that's better than the current implementation. I'm also not very familiar with Syncthing itself so I'd need to find the config file and what exactly needs to change.

Anyways, I'll be monitoring the thread for the final decision. Hopefully we get there before I forget everything I just learned about Android.

imsodin:
I have some questions only based on the comment discussion:

> SettingsActivity restarts the whole app cleanly after a theme change

Does the theme change work without a restart? If no, we need it, if yes, why do we need it?

> reflecting the theme change in config.xml

How is the chosen theme persistant over restart now then?

> I'm personally a bit scared this merge would deviate official much from the fork - there would be large differences in resources

What additional divergence does this introduce? I assume you needed to include the "dark theme read drawables" into your fork as well, or didn't you? What could be done (in this PR or as an immediate followup) to limit the divergence?

Catfriend1:
@imsodin:
Just a quick answer as I'm rare on time...
> Does the theme change work without a restart? If no, we need it, if yes, why do we need it?

Android needs to restart the activity to apply theme changes. Syncthing does not require to restart.

> How is the chosen theme persistant over restart now then?

camingus changes the Android theme, not the SyncthingNative theme in this PR, so this is persisted by Android SharedPreferences.xml and not config.xml.

> I assume you needed to include the "dark theme read drawables" into your fork as well, or didn't you?

I already merged "a did myself finish and polished version of this PR" to the fork's repo. (https://github.com/Catfriend1/syncthing-android/pull/401) This didn't include any drawable resources to keep the differences low between official and fork. Dark theme works but we have sometimes  "black icons on grey background" where camingus suggests "white icons on grey background" to have better contrast.

> What could be done (in this PR or as an immediate followup) to limit the divergence?

It could be done at the first stage only touching the code which would be quite "compatible" with the fork's code. Then, the assets could be changed with the outstanding ToDo from camingus to add the support library for the icon's color change by the OS. I currently don't know a good solution on how the resource differences could be kept low in the long run. Syncthing Fork has more/other resource images than official already which keeps me away from merging camingus' assets there as this would be a "80% complete" work. (The assets are well done, though.)

AudriusButkevicius:
Sorry, but I don't think I am useful here.
There are 3 of you, I am sure you are cumulively smart and careful among yourselfs to work this out :)

camingus:
So the biggest blocker is just the icons? If I make new icons for the fork can we merge this? I'm not a graphic designer, but Inkscape has a [cool feature](https://www.wikihow.com/Trace-an-Image-Using-Inkscape) where you can give it a png and it'll draw an SVG for you.

Catfriend1:
Let's say I'd appreciate that if you like to :). That would mean I could also ditch my black and white variants of the same asset. Very nice ;)

app/src/main/res/values/strings.xml
```diff
@@ -812,5 +814,6 @@ Please report any problems you encounter via Github.</string>
     <string name="browse">Browse</string>
     <string name="no_sub_folder_is_selected">No sub folder is selected</string>
     <string name="share_device_id_chooser">Share device ID with</string>
+    <string name="title_activity_themed_app_compat">ThemedAppCompatActivity</string>
```
AudriusButkevicius: Seems like a weird string to translate?

app/src/main/res/values/arrays.xml
```diff
@@ -15,4 +15,14 @@
...
 
+    <string-array name="theme_names">
+        <item>Light</item>
+        <item>Dark</item>
+    </string-array>
+
+    <string-array name="theme_values">
+        <item>1</item>
+        <item>2</item>
+    </string-array>
```
AudriusButkevicius: Why the numeric values? Can't we just use text? Or is this for the purpose of translation vs value in storage?

If yes, I don't see the preference names being in translation or is that automatic?

AudriusButkevicius:
Also, are the known issues addressed?

camingus:
Ah sorry I missed the email notifications on this. The known issues with popup colors were addressed. The known issue with the icon colors was not. The reason it wasn't addressed is because it looks like it would add quite a bit more to this review. So it may make more sense in a separate pull request. According to the release notes for Android [support library 23.2](https://android-developers.googleblog.com/2016/02/android-support-library-232.html) the data types of all of the icons would need to change. And the project will need to take a dependency on com.android.support:appcompat-v7:27.0.2. If it doesn't seem like too much of a deal to include that in this pull request it can be included. The icon colors being black with the dark theme on also isn't a total deal breaker to have for a bit. You can still see them. And this issue only impacts older devices.

app/src/main/res/values/arrays.xml
camingus: The reason why these are numbers is so that they can be used with android's DayNight theme. They're based on the [constant values on AppCompatDelegate](https://developer.android.com/reference/android/support/v7/app/AppCompatDelegate.html#mode_night_auto).

app/src/main/res/values/strings.xml
camingus: That is a really good point. It looks like this is another one of those things that was added automatically by android studio when the ThemedAppCompatActivity class was created. It can be removed.

app/src/main/res/values/arrays.xml
camingus: I should note that they don't show up as numbers in the UI. They're numbers so that this value can be passed to `AppCompatDelegate.setDefaultNightMode` without having to convert it first.

AudriusButkevicius: Question about translations still stands

AudriusButkevicius:
Lets rebase, remove the ackward string and merge. We've bumped a bunch of deps while this was brewing.

imsodin:
@camingus Any chance you can fix the string today (nowish)? I'd like to have this in the overdue release :)  
Or anyone else knows if its usage in the manifest can just be removed, then I can do it myself:  
https://github.com/syncthing/syncthing-android/pull/1301/files#diff-c8cc2dd42271f2bf39c6aa81eb6a5529R33

camingus:
Sorry I don't have as much free time these days as I did 2 months ago. I'm stuck working my day job for the next 8 hours. I can start looking into the left over issues after that. But I'm also completely fine with you making the changes yourself. I guess I'll get around to it when I get home if hasn't been done already.

Unfortunately I don't know if that class can be removed from the manifest. I would remove it see what happens.

camingus:
I removed the class from the manifest and it seemed fine. This answer on [stack overflow](https://stackoverflow.com/a/5844616) seems to also support that it's fine to remove.

app/src/main/res/values/arrays.xml
camingus: You're right. I've moved the strings to strings.xml

app/src/main/res/values/strings.xml
camingus: This has been removed.

imsodin:
Thanks @camingus for taking care of this. As the release deployment is failing due to the dependency upgrades and that's not yet resolved, we may as well add this to the RC. However somehow Github is got completely confused, the diff makes no sense whatsover. Can you rebase your changes on master to hopefully get github back on track - thanks.

Catfriend1:
Thanks for camingus' contribution so far. I'd follow Simon's suggestion to add this into the RC so we can get user's feedback on it and hopefully release it in july with 1.2.0 :-)).

PR #1307, opened 2019-04-20
Disable badge for persistent notification (fixes #1305)

### LINKED ISSUE #1305: the dot is always shown
image

Please don't tell this is due a permanent notification. as there are many other apps, like linphone or battery charge lime, that also have a permanent notification but without a dot on the icon.

GermanCoding:
**Purpose:**

As mentioned in #1305, some users don't like the "dot" (badge) displayed when there is not really new info available.

**Changes:**

The current behavior is that the dot is always shown when syncthing has a notification open - usually the persistent `Syncthing is running` message used for the background service.

This PR changes this behavior and disables the dot/badge when only the `Syncthing is running` notification is displayed. This does not affect the operation of the notification or the service.

_Please note that this **only** affects the `Syncthing is running` notification._ The `Syncthing is disabled` message or other info notifications will still get a badge. This is intentional, because I think these notifications deserve attention from the user.

**Tests:**

I verified this briefly on Android Emulators running Android 8.1, 7.0 and 4.4. On API level >= 26 (Oreo) this works as intended, on lower Androids this has no effect because this feature does not exist.

**Compatibility note:**
`NotificationChannel.setShowBadge()` was added in API level 26. The buildscript currently targets/compiles this API level, therefore I concluded that usage of this method is unproblematic. The method call is guarded by a version comparison, to avoid calling this on anything lower than Oreo.

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/NotificationHandler.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/NotificationHandler.java
@@ -57,6 +57,7 @@ public class NotificationHandler {
             mPersistentChannel.enableLights(false);
             mPersistentChannel.enableVibration(false);
             mPersistentChannel.setSound(null, null);
+            mPersistentChannel.setShowBadge(false); // Disables the dot next to the app icon
             mNotificationManager.createNotificationChannel(mPersistentChannel);
 
             mPersistentChannelWaiting = new NotificationChannel(


Catfriend1:
Hi,

I've tested this von AVD 8.0 (API level 26) and the badge still appears. I've also tried a different commit for the other channels, setting two of them to false and setting three of them to false - no luck.
https://github.com/Catfriend1/syncthing-android/pull/408/commits/dd05263627b6d634960388bf401fa0e100652383

Can anyone please confirm this is working?

GermanCoding:
I'm repeating myself, but it worked in my tests on 8.1. I just performed an additional test with Android 8.0/API level 26 and it works there too. Screenshot below.

It's possible that you have uncleared notifications which prevent the badge from disappearing. Have you tried a clean install?

Android_8_Test

Catfriend1:
I'll try a fresh install later and post the screenshot from the emulator if the issue persists.

capi:
It seems to work on my Android 9, that's why I approved it. The thing is, that the handling of the notification dot can vary a lot between the ROMs and also can be forced and prohibited by users via the notification settings. That's how I got rid of the dot for this notification until now.

I think that the change is small and doesn't hurt, even if it might not fix it for everyone. I'm quite sure that it shouldn't cause problems for anyone.

capi:
Also, changes to the notification channel do not affect existing notifications, only newly created ones. And if you modified any of the settings for the channel at any time, it's possible that Android does not pick up the new settings (intentionally). That all could come into play here.

Catfriend1:
Thanks for the details on it, good work! Sure, it couldn't break anything so it's good :).

PR #1310, opened 2019-04-27
Show notification badge if user interaction is required

ghost:
Purpose:
- Also setShowBadge for the other two notification channels.

Testing:
- I've verified this working on Android AVD 8.0 emulator by starting Syncthing and letting it sit active in the background (no badge dot shown) and sharing a folder to it triggering a user interaction required-notification (badge dot down).

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/NotificationHandler.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/NotificationHandler.java
@@ -57,7 +57,7 @@ public class NotificationHandler {
             mPersistentChannel.enableLights(false);
             mPersistentChannel.enableVibration(false);
             mPersistentChannel.setSound(null, null);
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/nutomic/syncthingandroid/service/NotificationHandler.java
```diff
@@ -57,7 +57,7 @@ public NotificationHandler(Context context) {
             mPersistentChannel.enableLights(false);
             mPersistentChannel.enableVibration(false);
             mPersistentChannel.setSound(null, null);
-            mPersistentChannel.setShowBadge(false); // Disables the dot next to the app icon
```
Catfriend1: I've removed the comment as it doesn't exist in STF and "setShowBadge" is pretty clear what it does from my point of view.

AudriusButkevicius (approved):

PR #1343, opened 2019-06-05
Update go version

imsodin:
This will require an update of the docker image used on the build server I think, but I don't know how/what exactly (@AudriusButkevicius).


AudriusButkevicius:
There is a build job on TC that builds and pushes the image

PR #1344, opened 2019-06-07
Revert "Support Variable Size Blocks (VSB)"

AudriusButkevicius:
Reverts syncthing/syncthing-android#1270


AudriusButkevicius:
I am puzzled why it cries about removed strings.

Catfriend1:
prepare_release.bash does it automatically, here is an excerpt of the relevant script part:

>Updating Translations
-----------------------------
"
tx push -s
# Force push/pull to make sure this is executed. Apparently tx only compares timestamps, not file
# contents. So if a file was `touch`ed, it won't be updated by default.
# Use multiple transifex instances for pulling to speed things up.
tx pull -a -f -r "syncthing-android.stringsxml" &
tx pull -a -f -r "syncthing-android.description_fulltxt" &
tx pull -a -f -r "syncthing-android.description_shorttxt" &
tx pull -a -f -r "syncthing-android.titletxt" &
wait
./gradlew deleteUnsupportedPlayTranslations


It uploads the translations from the github repo (lang en) which will result in the deletion of removed strings in all other languages. Then it pulls translations from all languages other then the primary lang en to the git repo. Now, a removed string is completely gone and lint no longer complains there are string translations present not being there in the primary lang en.

An alternative would be to suppress the lint warnings of this specific check for extra translations.

AudriusButkevicius:
Ran the push and pull (which is probably not a brilliant idea) and that cleaned stuff up. Let's see what happens.

Catfriend1:
I've also ran into this problem before, I decided only to run push pull and commit the result if I intend to merge this soon into the master. In clear words: you shouldn't run it again in another PR when different strings were addressed until this PR got merged. The result would be transifex rebirth-ing strings missing here and deleting from another thread as it doesn't know about the github code traits.

AudriusButkevicius:
cAN YOU REAPPROVE AND MERGE?

app/src/main/res/layout/fragment_folder.xml
```diff
@@ -221,38 +221,6 @@
...
-                android:layout_height="wrap_content"
-                android:layout_marginLeft="75dp"
-                android:layout_marginStart="75dp"
-                android:layout_marginTop="-20dp"
-                android:textAppearance="@style/TextAppearance.AppCompat.Caption"
-                android:text="@string/folder_variable_size_blocks_description"
-                android:focusable="false"/>
-
-            <LinearLayout
-                android:id="@+id/editIgnoresContainer"
```
imsodin: Why is this removed?

AudriusButkevicius: Huh, good point.

AudriusButkevicius: It seems it slipped in with the original PR. Need to figure out if its used.

Catfriend1: Let me check please when I'm back from my vacation. From what I recall this is an extra layout container that came from the fork but is useless in official as there isn't the in-app edit .stignore UI. So at a first glance it can be removed.

AudriusButkevicius:
It's actually not used, so this is still kosher.

imsodin:
> It's actually not used, so this is still kosher.

It still needs to go or stay together with the `android:id="@+id/edit_ignores` thing after it.

AudriusButkevicius:
edit_ignores is used, as it's a think you click on to open external text editor

app/src/main/res/layout/fragment_folder.xml
```diff
@@ -221,38 +221,6 @@
...
-                android:layout_width="match_parent"
-                android:layout_height="wrap_content"
-                android:background="?selectableItemBackground"
-                android:orientation="vertical"
-                android:gravity="center_vertical">
-
                 <TextView
```
imsodin: Continuing from https://github.com/syncthing/syncthing-android/pull/1344#issuecomment-500565690:  
What made me think this belongs together with the remove paragraph above is that this part is indented compared to the removed `LinearLayout`, i.e. before it looked like it was a child of the editIgnoresContainer, now it's a child of something much further up. If that has no particular meaning or is fine, then just ignore this.

AudriusButkevicius: I have no idea, I looked for the string, it's not there. I don't think any of this stuff is parent-childish, it's just a list of stuff.

AudriusButkevicius:
So what's happenin

imsodin:
I wanted to wait for the string fix in #1301 to merge that before the release (wanted to do this change together with bumping syncthing). I'll try to fix #1301 myself now, without changing authorship.

Catfriend1:
Well if it can't wait until I'm home from vacation and can test the APK it's okay merging this and I'll test it later. Guessing we have time until 1.2.0 gets released to make it clean.

PR #1346, opened 2019-06-13
Upgrade gradle play publisher to 2.2.1

imsodin:
Release deployment is broken, most likely as gradle play publisher 1.x is not compatible with gradle 5.x. Upgrading to 2.x requires running `./gradlew bootstrap` due to extensive changes to how play metadata is handled and that requires having the key. @calmh Is it possible for you to run that command on teamcity or manually and commit the results to this branch?


calmh:
I probably can, but I'm not sure where. The build runs in a fairly short lived docker container, and the agent container itself has no gradle stuff. Is this maybe something that should go in the docker/prebuild.sh or somewhere instead? I have zero clue how the gradle stuff works.

imsodin:
If you have "manual" access with the keys in place, then literally just checkout this branch and run `./gradlew bootstrap` - no docker container required.

Otherwise maybe you can duplicate the "Release Syncthing-Android" build job and modify the custom script in step one to run `./gradlew --no-daemon bootstrap` instead of `./gradlew --no-daemon buildNative lint ...`. No idea how you then get access to the generated files in `app/src/main/play/...`, maybe export the used docker image as an artifact and extract it from there.

AudriusButkevicius:
What does bootstrap do? Is it some local changes or remote changes?

calmh:
Yeah, exactly. :) If this is just that it needs to have the keys in order to be able to set something up in the repo I think Audrius has those and could do it on some computer with the relevant software installed... (Doing this in the build container is probably cumbersome.)

imsodin:
It does change the directory/file structure of all the listings (descriptions and such) locally, but apparently it pulls the information from google play instead of using what's in the repo.

calmh:
Maybe I can build the docker image locally and run it there. I have no interest in trying to satisfy all the SDK/NDK/Java/Studio/whatever requirements on my work mac :)

calmh:
Or if you just lack the signing keys and have everything else in place, we can solve that.

imsodin:
Docker should work, just takes a moment as it needs to build various ndk/go things.

And it's just the signing keys (I hope, I'd lie if I'd say I am sure about anything regarding android o.O ).

calmh:
I'm experimenting.

imsodin:
This upgrade is a really nice experience.

/sarcasm off  
Thanks for all the help with this!

Catfriend1:
@imsodin: 
I'm trying to follow this PR since several hours but having no luck with the gradlew publishListingRelease command. It halts with a http-500 status code error and no error message points at the reason. This is happening in the lang "ar" after it tries to upload images. In the play/listings/ar folder there are only .txt files (like in the official repo) and no PNG's , no "graphics"folder". 
I saw "uploadImages = false" was removed as it is no longer available in the play-publisher plugin.
Do you have any clue why I cannot upload?
```
 com.google.api.client.googleapis.json.GoogleJsonResponseException: 500 Internal Server Error
  {
    "code" : 500,
    "message" : null
  }

```

imsodin:
Lets see whether something comes out of the upstream bug report. A workaround if it keeps failing would be just moving the play listings graphics to another location and updating them manually in the console when they change.

PR #1348, opened 2019-06-14
Update build scripts and .gitignore and disable travis emails

imsodin:
This has been lying around a while and I think it would be beneficial for others compared to the existing build scripts. Essentially it just splits the different tasks up, such that you can e.g. only bump the version, or only push translations or whatever. The only new thing is, that I added `check-version.bash` that exits with an error if the syncthing-android version is "not compatible" with the syncthing version. In quotation mark because the check is primitive, it just makes sure that the Syncthing version is a prefix of the android version (e.g. android 1.1.1-rc.1 and syncthing 1.1.1 would pass the test, but that's fine, the important thing is that it doesn't pass the other way: stable syncthing-android version with an unstable syncthing version).

If anyone knowing gradle could hook up `check-version.bash` with the release stage of the build, such that it fails to release if that script fails, that would be awesome.


scripts/push-translations.bash
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -e
+
+printf "Pushing Translations
+-----------------------------
+"
+tx push -s
```
AudriusButkevicius: This is kinda useless?

scripts/check-version.bash
```diff
@@ -0,0 +1,26 @@
...
+
+PROJECT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && cd .. && pwd )"
+cd "syncthing/src/github.com/syncthing/syncthing/"
+CURRENT_TAG=$(git describe)
+if [ "v${NEW_VERSION_NAME#$CURRENT_TAG}" = "v${NEW_VERSION_NAME}" ]; then
+    printf "New version name %s is not compatible with Syncthing version %s\n" \
+           "$NEW_VERSION_NAME" "$CURRENT_TAG"
+    exit 1
+fi
+cd ${PROJECT_DIR}
```
AudriusButkevicius: This doesn't do anything?

scripts/push-translations.bash
imsodin: Sure, it just serves as a kind of bookmark, as in I don't kneed to know what program and/or options to invoke.

Catfriend1: @imsodin :
tx push -s uploads the english strings (for all resources, the so called default locale) to  transifex so other's can start translating new or changed strings. Maybe exit code should be checked if the cmd succeeded.

imsodin: I know what it does, my point is that if I want to get strings to transifex I don't necessarily remember `tx push -s` but looking at tab completion I would consider `push-translations.bash` as a tool that likely does what I want to :)

Catfriend1: Ahh now I understand :-). Yes I agree this easier to remember and also better for coding newbies because they won't need to dig into transifex docs first to know what is what (sometime ago I felt lost at this point I remember and looking at the scripts helped a lot to learn.)

AudriusButkevicius:
I think that this shouldn't work as a lot of that script depended in being in certain dirs which is now lost?

imsodin:
What depends on being in certain dirs? I checked if anything (eg dicker) uses the script at the repo root, but didn't thing anything. and the scripts do work

imsodin:
I accidentally committed part of these changes when preparing the 1.2.0-rc.2 release, and now committed the rest to get into a consistent state again. If there are problems/concerns with it, I am obviously happy to change/revert it.

PR #1400, opened 2019-09-09
FOREGROUND_SERVICE is required if targetSdkVersion is set to 28

sungamagnus:
Reference: https://developer.android.com/about/versions/pie/android-9.0-migration#tya and https://stackoverflow.com/a/52382711


AudriusButkevicius:
Someone elses PR landed first.

PR #1405, opened 2019-09-10
Update gradle

imsodin:
gradle-play-publisher 2.4.0 has a fix for our new build problems and requires a newer gradle - lets hope this works out of the box (https://github.com/Triple-T/gradle-play-publisher/issues/629).


AudriusButkevicius:
My money is that this will cause new issues when releasing, but lets see.

imsodin:
> My money is that this will cause new issues when releasing, but lets see.

I am not gonna disagree with that, but at least it fixes one issue that makes release currently nearly impossible.

imsodin:
@AudriusButkevicius Are you ok with merging this and the publisher and seeing how it goes?

AudriusButkevicius:
You have my blessing

PR #1479, opened 2020-02-24
removed menu decorations, decreased file sizes with optipng

PanderMusubi:


Catfriend1:
Hi,

Thank you. Can I have some Screenshots please?

PanderMusubi:
These are the screenshots. There were already in the repo, I only improved them slightly. Not sure what screenshots you are asking for.

Catfriend1:
I just see your text, sorry, might be because I'm on the phone.

PanderMusubi:
No worries.

Catfriend1:
Looking at the PR on a computer, it says "binary file not shown" so one cannot see the images in the GitHub Web Diff but checking out the PR. Hm... I wonder if this would decrease the APK file size of the app. I see you're changing the png's residing in the play/listing subfolders which are only relevant to the triple-T-play-publisher plugin which uploads them to the play store article when a new release is published. As I understood how APK packaging works, I bet those PNGs are not included in the built APK file so saving file size won't apply here.

Anyone likes to chime in and say yes/no to merging this? I'm on a neutral opinion.

Catfriend1:
@PanderMusubi  Can you please explain why you converted the following file from JPG to PNG?

> app/src/main/play/listings/en-GB/graphics/feature-graphic/feature_graphic.png

I would not recommend converting the file format just because of file size because Google Play Store has policies in place which file formats and pixel widths they accept on upload. If we send sth incompatible, GPlay API might throw an API error 500 at us when doing the release.

imsodin:
These are just listings, i.e. playstore stuff. I don't see the motivation of shaving <1MB from that or the git repo.

PanderMusubi:
@imsodin These images are also used by F-Droid. Optimizing the images without degradation of the quality helps save bandwidth usage on the user side for the website and the client app via both Google Play and F-Droid.

@Catfriend1 The JPG has non-photographic elements and had a quality of 100 (no compression). So PNG with some compression is smaller and better in quality. Feature graphic in Google Play etc. also supports PNG. I was wondering why it was JPG in the first place.

IzzySoft:
@Catfriend1 @imsodin as you can see from the path shown in "files changed", this only affects the screenshots in your Triple-T structure. What Pander is trying to achieve? Look at the file sizes. Those are the screenshots shown along your app description within the F-Droid app (and Play Store app if you published there as well). Consider users with a small data budget: the less file size (while still having sufficient details), the better.

That's what this PR is about, if I understood Pander correctly. Assuming it is, I'd suggest to reopen and merge :wink:

And no, app size is not affected in any way. This is just for the "stores".

imsodin:
My (potentially misguided) rationale is that the costs outvalue the benefits: Someone trusted by the project needs to check the changes for correctness, which then save a tiny amount of bandwidth (I consider that point valid anyway, but syncthing being a data-transferring app makes it even more clear).

IzzySoft:
@imsodin what correctness do you expect to be checked for in screenshots provided by the project via Fastlane? Or did you mean one of you would need to check that Pander didn't replace your screenshots with pr0n? :speak_no_evil:

As for the "potentially misguided": I didn't check your images. But I've seen repos where each screenshot was ~2..5M in size. There were usually 5+ screenshots. Makes altogether 10..25M, just for the screenshots of one app. On my personal device, that's 10% of my monthly data plan. I know folks where it would be a quarter, just for checking one app. Yes, there are areas where 100M per month is "normal" (or even "quite a lot")…

@PanderMusubi maybe you could give some numbers (old size, new size, amount saved), to illustrate my words? If it's just "50k saved" I'll immediately shut up and run away ashamed :scream:

IzzySoft:
> Looking at the PR on a computer, it says "binary file not shown" so one cannot see the images in the GitHub Web Diff but checking out the PR.

Click the 3 dots at the upper-right corner, select "view file". Hm, 885k -> 667k (featureGraphic), still quite big. 1024x500 – yes, that's what Triple-T recommends. If I download that PNG, convert it to JPG and set quality to 75%, it goes down to ~100k – and I don't see a difference in quality (we're not in an art gallery – featureGraphic is for illustration only).

For the screenshots, it indeed rather looks minimal – but sums up to another 100..150k maybe. So altogether, about 1M saved – for each user taking a look at your app's listing (assuming you'd take my 75% jpg).

IzzySoft:
Thanks @Catfriend1 – and new screenshots, when "optimized", would of course "supersede" this :wink:

PanderMusubi:
Thanks all. Note that GitHub also offers diffing images in three ways, please try 2-up, Swipe and Onion Skin at https://github.com/syncthing/syncthing-android/pull/1479/commits/241c464c6a26b9ef6fe912b508896b85f215728a#diff-bb2f31e4883e2869be4790427b9a58a2

PR #1490, opened 2020-04-01
Create empty file in marker directory (fixes #1249)

### LINKED ISSUE #1249: [FEATURE] Create dummy, non-empty file in .stfolder to prevent automatic cleanup of empty directories
Some Android distributions, like MIUI, have built-in filesystem cleaner that clean up trash from various locations. Common feature of it is to remove empty directories. I've run into issue where Xiaomi MIUI's Cleaner removed .stfolder rendering syncthing not working. I had to manually add some random, non-empty file there to prevent it from happening again.

It would be nice to have an option in Syncthing for Android to create any non-empty file in .stfolder for the sake of cleaners.

calmh:
HTH


calmh:
I didn't test any of this.

Catfriend1:
According to @hever s feedback it works fine. Lgtm.

AudriusButkevicius:
I guess we'll know if it does not work...

Issue #1515, opened 2020-05-30
Theme: follow system default


igalic:
Rather than having to switch between light and dark myself, in each app, like some caveman, android 10 has APIs which allow apps to read out what the current system preference is, and follow that. 

please consider implementing this as well!

### Version Information

    App Version: 1.5.0
    Syncthing Version: 1.5.0
    Android Version: Android 10 / OxygenOS 10

dicky-cat:
[http://docs.syncthing.net/]() #1355

praveenrajput:
If no one is working on this issue, can I fix this?

AudriusButkevicius:
I don't think anyone is working on this.

praveenrajput:
What should I do for localization of "FOLLOW SYSTEM" string, as I don't know exact translation.

AudriusButkevicius:
Translations are handled in transifex. It should be fine as long as the string is exposed as a resource.

vhouriet:
I think we can close this issue now. The app defaulted to dark theme for me on multiple devices.

PR #1555, opened 2020-09-28
Theme: follow system default #1515

praveenrajput:
- Now app by default follows system theme for api level 28 and higher, for api level below 28, it fall backs to light theme.


AudriusButkevicius (approved):
Lgtm

PR #1556, opened 2020-10-06
Add Internal Activity for QR Code Scanner

### LINKED ISSUE #1533: Switch to a less invasive QR scanner app
The Syncthing app requires "Barcode Scanner" by ZXing Team when scanning host IDs.
This app has started requiring more and more permissions that are invasive to users' privacy.
As demonstrated by the screenshot below, it is requesting/using permissions that aren't required to function with Syncthing and that are invasive to privacy; including:
* Device & app history
* Contacts
* Wi-fi

Screenshot

sumit-anantwar:
Resolves #1533 

Add Internal Activity for QR Code Scanner.
Please check the below screen recording for the demo.
https://www.dropbox.com/s/h507363wsmj338q/Screen_Recording_20201006-221716_Syncthing.mp4?dl=0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ before_install:
   - eval "$(gimme stable)"
   # Hack to accept Android licenses
   - yes | sdkmanager "platforms;android-28"
+  - yes | sdkmanager "platforms;android-29"
 
 # Cache gradle dependencies
 # https://docs.travis-ci.com/user/languages/android/#Caching
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -7,12 +7,12 @@ plugins {
 dependencies {
     implementation 'eu.chainfire:libsuperuser:1.1.0.202004101746'
     implementation 'com.google.android.material:material:1.0.0'
-    implementation 'com.google.zxing:android-integration:3.3.0'
     implementation 'com.google.code.gson:gson:2.8.6'
     implementation 'org.mindrot:jbcrypt:0.4'
     implementation 'com.google.guava:guava:29.0-android'
     implementation 'com.annimon:stream:1.2.1'
     implementation 'com.android.volley:volley:1.1.1'
+    implementation 'com.journeyapps:zxing-android-embedded:3.6.0'
     implementation 'androidx.constraintlayout:constraintlayout:1.1.3'
     implementation 'com.google.dagger:dagger:2.28.3'
     annotationProcessor "com.google.dagger:dagger-compiler:2.28.3"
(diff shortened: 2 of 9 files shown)


AudriusButkevicius:
@Catfriend1 what do you think?

Catfriend1:
Cool work:-)! Will try it out during my next dev session and report back. Generally a good idea to remove being dependent from another app. I'm currently using Binary Eye QR Scanner only for Syncthing's purpose.

tomasz1986:
Does the scanner work in the older versions of Android too? Syncthing itself, as of now, supports Android 4.1 and up.

sumit-anantwar:
> @sumit-anantwar Do we really need to bump gradle to use this?

No, we don't need to bump Gradle, I did it out of habit.
If that's an issue, I can roll it back.

sumit-anantwar:
> Does the scanner work in the older versions of Android too? Syncthing itself, as of now, supports Android 4.1 and up.

I am using Zxing itself, just an embedded version of it, published by the team at JourneyApp.
https://github.com/journeyapps/zxing-android-embedded

We are already using this library in a few of our apps. It works reliably.
By default, it supports Android API 24+. If we require support for <24, we just need to add dependency to `zxing-core:3.3.0`.
Will push that commit by EOD today.

Catfriend1:
Looks good to me now. I'm just wondering how to judge implementations/dependency correct from the point of a reviewer. Does anyone else have knowledge about this? E.g. we already have eu.chainfire and this means to me "someone in the past decided to trust it/them". Can we also accept "com.journeyapps:"? I'm a little bit unsure because I saw the discussions about the new Mozilla Firefox F-Droid release a few weeks ago and there, the F-Droid team said "some repositories of implementations are okay, others not" and it was brought up because Mozilla recently had added a new implementation in question.

AudriusButkevicius (approved):

tomasz1986:
> I'm just wondering how to judge implementations/dependency correct from the point of a reviewer. Does anyone else have knowledge about this?

This is only what I have read here and there, so the information may not be precise, but F-Droid seems to have some kind of a blacklist of disallowed components/dependencies. If they are detected, then the application is automatically rejected from the repository. They keep updating the blacklist, so even if a particular application is OK right now, the situation may change in the future. The criteria seem to be obvious, i.e. any non-free and non-open source additions are rejected.

Catfriend1:
@tomasz1986  :-).

Then I'm all good with it now. Thanks for the contribution @sumit-anantwar  🚀  🎉

PR #1565, opened 2020-10-17
Add privacy policy for play store

imsodin:
The play store rejects the update due to a missing privacy policy. Apparently needing the camera requires having one, and we do need the camera since https://github.com/syncthing/syncthing-android/pull/1556


AudriusButkevicius:
Does UR fall under user information?

imsodin:
I'd say so, that's why I mentioned it in addition to linking to the syncthing docs

AudriusButkevicius (approved):

PR #1566, opened 2020-10-18
Add "all files" management permission

wersimmon:
Android 11 (R) requires the new "ACTION_MANAGE_ALL_FILES_ACCESS_PERMISSION" permission, added in this version, to view the files of other applications.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -8,6 +8,8 @@
     <uses-feature android:name="android.hardware.location.gps" android:required="false" />
 
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+    <!-- ACTION_MANAGE_ALL_FILES_ACCESS_PERMISSION is required to access files from other apps on 11 "R" -->
+    <uses-permission android:name="android.permission.ACTION_MANAGE_ALL_FILES_ACCESS_PERMISSION" />
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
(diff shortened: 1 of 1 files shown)


imsodin:
We aren't targetting android 11 yet and won't be for a while, as google doesn't allow one to use this permission at the moment (until "early 2021"): https://support.google.com/googleplay/android-developer/answer/9956427

wersimmon:
Can we use the legacy external storage request, instead, then?

I am currently running Android 11 and this is breaking my workflow

AudriusButkevicius:
That's already asked by the app I believe.

imsodin:
I assume they mean [requestLegacyExternalStorage](https://developer.android.com/reference/android/R.attr#requestLegacyExternalStorage) which we do not use yet, as we still target android 9/sdk 28. I was under the impression targeting the old sdk is equivalent to requesting legacy behaviour on newer platforms, but apparently that's no longer the case on android 11. We anyway need to target sdk29 soon to be allowed to keep publishing to google play, and there we'll need to set `requestLegacyExternalStorage` anyway. I'll do that for the next (1.11.1 or 1.12.0) release.

wersimmon:
Yes, that's what I was referring to. Thanks!

auanasgheps:
Please note that _All File Management_ is a **big lie** from Google.

On my OnePlus 8T - Android 11 native:
- Syncthing-Fork - Target API 30 - All Files enabled - I am [unable ](https://github.com/Catfriend1/syncthing-android/issues/704#issuecomment-737522026) to select the root of my storage
- Syncthing official - Target API 28 - Standard Permissions - I can use the root storage

I know it's Google playing dirty, but please, please, please: it is a huge change which breaks "full" device backups with Syncthing, let's try to find a workaround.

PR #1569, opened 2020-10-21
Fix typo in es l10n

josesalvador:


AudriusButkevicius:
This is all managed in transifex so changes should be done there.

josesalvador:
> This is all managed in transifex so changes should be done there.

ACK

I tried to join the translation team to fix the typo but I don't have any answer yet.

PR #1575, opened 2020-11-10
Target android sdk 29

imsodin:
This PR bumps our target sdk level to 29 (android 10), requesting legacy storage. This is a requirement for publishing google play updates (caveat at the end). The behaviour *shouldn't* change, but what should and shouldn't happen on android is a mystery, so I'll add a call for testing/feedback on the forum once this is in a release candidate.

@Catfriend1 A sanity check if I missed something obvious would be appreciated. I did follow the Android 11 troubles you had, but I believe those problems should not come up as we still target android 10/sdk 29. 

Google play caveat:  
<strike>Google is blocking us at the moment anyway, as suddenly our coarse access location permission for restricting sync by wifi name seems to be a problem - I wrote to the support, I am expecting a resolution in 6-8 weeks. I intend to keep releasing as usual, users annoyed by not getting those release should switch to F-Droid (users should anyway switch to F-Droid ;) ). </strike>  
Google support has responded, details in https://forum.syncthing.net/t/how-to-react-to-google-play-restrictions/15872/26

For reference:  
 - WiFi permissions: https://developer.android.com/guide/topics/connectivity/wifi-scan
 - Deprecated method: https://developer.android.com/reference/android/net/wifi/WifiManager.html#getConfiguredNetworks()

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/views/WifiSsidPreference.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/views/WifiSsidPreference.java
@@ -5,6 +5,7 @@ import android.app.Activity;
 import android.content.Context;
 import android.content.pm.PackageManager;
 import android.net.wifi.WifiConfiguration;
+import android.net.wifi.WifiInfo;
 import android.net.wifi.WifiManager;
 import android.os.Build;
 import android.os.Bundle;
@@ -18,6 +19,7 @@ import com.nutomic.syncthingandroid.R;
 import com.nutomic.syncthingandroid.service.Constants;
 
 import java.util.Arrays;
+import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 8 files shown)


Catfriend1:
Hi, thanks for working on it @imsodin . I'll give https://build.syncthing.net/repository/download/SyncthingAndroid_Build/85323:id/apk/debug/app-debug.apk a quick shot on my Android 10 phone.

imsodin:
Fixed location permission handling to be sdk29/android 10 compatible. And the method to get configured wifis has been deprecated, apparently it's too dangerous to let the users decide if they want to share that info with an app. So now we just display already stored wifis plus the currently connected wifi.

This might not build as the docker image was changed and isn't deployed yet - a local docker build worked fine.

imsodin:
@AudriusButkevicius If you have a moment, I'd appreciate a quick review. I mostly just removed code, but there's a few changes/additions and I'd appreciate getting a check if the little code I wrote managed to violated any (java) best practices or whatever :D (no hurry though, I intend to skip a Syncthing release to have this in beta for a while).

Catfriend1:
@imsodin Could I test shot the Apk from teamcity on my phone or is it somehow non representative for testing because of the outdated docker? In case you'd like me to test it, I'd appreciate an APK :).

imsodin:
I changed the sdk tools. Weirdly enough that doesn't seem to be a requirement (CI still builds fine). I don't know how much of a difference it makes. My locally built apk has a separate id  (com.nutomic.syncthingandroid.debug) and name to make it co-installable with the normal release - if you want to test that: https://www.swisstransfer.com/d/9c6f38ca-da71-418b-ac0f-32c1df426786

Catfriend1 (approved):
Works on my Android 9 phone. If it isn't connected to any WiFi and one hits the button "choose whitelisted WiFi ssids" then it shows "<unknown ssid>". I think that is required for fail-safe... so you did it right and do not end up with an empty (null) list.

Catfriend1:
Did not do a test but looks even better now :)

Catfriend1:
@imsodin Do you need more help or testing on this? If required, I'd take another of your APK builds for a test.

imsodin:
Thanks for looking into it and testing on your device. I also did some basic testing on my android 10 device. I think we are fine on that front :)

I am waiting for a review by @AudriusButkevicius (or a "just go ahead, no time or interest" ;) ).

auanasgheps:
Sorry for spamming but please [consider this issue](https://github.com/syncthing/syncthing-android/pull/1566#issuecomment-737781698) before bumping up target SDK.

imsodin:
This is skd 29 / android 10 - it's not related to the all file permissions thing, we are still requesting legacy storage handling here.

If you want you can test an apk from CI or the link above and see if it still works using a "root path" - it should. Otherwise there will be an beta release with this to test.

auanasgheps:
@imsodin tested the debug version and it works! Thanks!

So this is going to be like this in the future? If yes, sorry for bothering you.

imsodin:
@AudriusButkevicius could you please have a look, now that v1.12.1-rc.1 is here? Thanks!

> I am waiting for a review by @AudriusButkevicius (or a "just go ahead, no time or interest" ;) ).

AudriusButkevicius:
Sorry. Completely missed this. As I though it was just chatter. I'll have a look later today

app/src/main/java/com/nutomic/syncthingandroid/views/WifiSsidPreference.java
```diff
@@ -59,99 +61,69 @@ public WifiSsidPreference(Context context) {
...
-                } else {
-                    Toast.makeText(getContext(), R.string.sync_only_wifi_ssids_need_to_grant_location_permission, Toast.LENGTH_LONG).show();
+        boolean connected = false;
+        WifiManager wifiManager = (WifiManager) context.getApplicationContext().getSystemService(Context.WIFI_SERVICE);
+        if (wifiManager != null) {
+            WifiInfo info = wifiManager.getConnectionInfo();
+            if (info != null) {
+                String ssid = info.getSSID();
+                // api lvl 30 will have WifiManager.UNKNOWN_SSID
+                if (ssid != "" && !ssid.contains("unknown ssid") && !selected.contains(ssid)) {
```
AudriusButkevicius, line 79: This lost the null check, and will blow up with NPE?

app/src/main/java/com/nutomic/syncthingandroid/views/WifiSsidPreference.java
```diff
@@ -59,99 +61,69 @@ public WifiSsidPreference(Context context) {
...
-    /**
-     * Removes any network that is no longer saved on the device. Otherwise it will never be
-     * removed from the allowed set by MultiSelectListPreference.
-     */
-    private void filterRemovedNetworks(Set<String> selected, CharSequence[] all) {
-        HashSet<CharSequence> availableNetworks = new HashSet<>(Arrays.asList(all));
-        selected.retainAll(availableNetworks);
-    }
+        if (!connected) {
+            Toast.makeText(context, R.string.sync_only_wifi_ssids_connect_to_wifi, Toast.LENGTH_LONG).show();
```
AudriusButkevicius, line 87: How do we know it's not a perms flop?
I guess we might not have asked for the permission at this point?

app/src/main/java/com/nutomic/syncthingandroid/views/WifiSsidPreference.java
imsodin: I don't see what might be null here? We create `selected` ourselves - it may be empty but not null. And `ssid` is a string.

AudriusButkevicius: Strings are nullable in java, so ssid.contains will blow up if ssid is null.

AudriusButkevicius: The comment before even talks about ithttps://github.com/syncthing/syncthing-android/issues/620 in

Catfriend1: Audrius is right, I've had tests in the past I've documented in code where ssid was returned as null by Android. That happens when location is turned off.
See https://github.com/Catfriend1/syncthing-android/blob/d38f56f6967447680020eeaa3828367a4599d653/app/src/main/java/com/nutomic/syncthingandroid/views/WifiSsidPreferenceBase.java#L131

imsodin: Right, right, `String` is a class... Missed the comment because it was part of the code I just discarded due to `getConfiguredNetworks` deprecation. At least I know now why I asked for a review :|  
Thanks!

imsodin:
Good to merge now?

AudriusButkevicius (approved):

PR #1579, opened 2020-11-18
Explicitly mention location in privacy policy

imsodin:
To comply with Google Play policies, details discussed in https://forum.syncthing.net/t/how-to-react-to-google-play-restrictions/15872


AudriusButkevicius (approved):

AudriusButkevicius:
Does this get attached to the submission? Otherwise I don't think they care about our md files. Also, we should explain that we do not collect location data in any way.

imsodin:
> Also, we should explain that we do not collect location data in any way.

I think that's covered by `but it never does look up your location`, isn't it?

> Does this get attached to the submission?

A link to this is in a privacy-policy field of the play console. I didn't check yet whether that already puts that into the store listing automatically (though that would be sensible) and we'll also have to add it to the app itself. Right now I am just looking at updating the privacy wording to hopefully be compliant.

PR #1602, opened 2021-02-03
Containers and build process improvements/fixes

imsodin:
- Use ndk 21 due to Go bug: https://github.com/golang/go/issues/42655

 - Use pre-built clang binaries instead of building ourselves  
The script building the standalone-ndk printed these messages saying you don't actually need to build it but use these pre-built ones already part of the ndk as downloaded. I assume that was just not in place at the time this step was created - or is there a catch? @AudriusButkevicius 

 - Build syncthing when creating container to prepopulate modules


AudriusButkevicius:
I just rewrote this script from bash to python, not really sure why bash did what it did.

AudriusButkevicius (approved):

PR #1608, opened 2021-02-14
Change capitalization from Github to GitHub in Welcome wizard

yegortimoshenko:
In GitHub, both Git and Hub are capitalized. Let me know if you want me to fix up locales, too!


imsodin:
This means the entire string needs to be re-translated in all locales. And you can't prevent that by making the change in all locale files in the repo it's still a new string on transifex. That's hardly any benefit for a lot of work.

yegortimoshenko:
Isn't it still a typo essentially though? :/

imsodin:
Sure, I am not saying the change is incorrect. Just that its drawbacks outweigh the benefits.

PR #1610, opened 2021-02-19
Improve privacy policy

imsodin:
I got feedback from google play support again, and now only our privacy policy isn't compliant any more (the prominent location disclosure now is - yeii xD ).

The (improved) privacy policy now lives at https://syncthing.net/android-privacy-policy  
https://github.com/syncthing/website/pull/18

I am not entirely sure what exactly was lacking before - here's the feedback:

> **Invalid privacy policy**
> 
> Based on our review, your privacy policy doesn't comply with our policy
> requirements. Please [add or update your privacy
> policy](https://support.google.com/googleplay/android-developer/answer/113469#privacy),
> and make sure it is available on an active URL (no PDFs), is
> non-editable, applies to your app, and **specifically covers user
> privacy including your app's usage of location data.** 
> 
> **Please make sure** privacy policy is *clearly labeled as a privacy
> policy in the context of the content* and specifies handling
> of location data.
> 
> Please note that the [Personal and Sensitive
> Information](https://support.google.com/googleplay/android-developer/answer/9888076#personal-sensitive) policy
> requires that: "The privacy policy must, together with any in-app
> disclosures, comprehensively disclose how your app accesses, collects,
> uses, and shares user data. Your privacy policy must disclose the types
> of personal and sensitive data your app accesses, collects, uses, and
> shares and the types of parties with which any personal or sensitive
> user data is shared."\

The part about labelling is emphasized, so I expanded the privacy policy description. I also made the privacy policy more explicit in terms of location data, as in specifically used the term "location data".  
Please share any other ideas/insights into what might be improved.

After merging I expect another round of the same game: Release, get an auto-rejection the day after and then ask support for an evaluation - lets see how that goes this time.

### DIFF
--- a/app/src/main/play/listings/en-GB/full-description.txt
+++ b/app/src/main/play/listings/en-GB/full-description.txt
@@ -8,4 +8,4 @@ Forum: https://forum.syncthing.net/
 
 Issues: https://github.com/syncthing/syncthing-android/issues
 
-Privacy-policy: https://github.com/syncthing/syncthing-android/blob/main/privacy-policy.md
+Privacy-policy: https://syncthing.net/android-privacy-policy
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -556,7 +556,8 @@ Please report any problems you encounter via Github.</string>
 
     <!-- Menu item linking privacy policy -->
     <string name="privacy_title">Privacy Policy</string>
-    <string name="privacy_summary">Open the Syncthing-Android privacy policy</string>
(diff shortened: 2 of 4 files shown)


tomasz1986:
I have given this a shot: https://github.com/syncthing/website/pull/19. Please have a look and let me know what you think. I have tried to expand the explanation a little, make the format clearer, and also change the wording to make it more formal and more "corporate", so that the content is (hopefully) easier to understand by those responsible for the review process.

Of course, I am no lawyer, but I have read tons of formal and semi-formal documents in my life, so there is that.

I did take some inspiration from Brave (https://brave.com/privacy/browser), which is also a FOSS software, although it does collect some user information, so it is not exactly the same situation, but still.

PR #1613, opened 2021-02-28
Improve and fix bugs in Slovak translation

jose1711:


AudriusButkevicius:
Translations are managed via transifex.

Issue #1626, opened 2021-03-25
Welcome wizard: Transparent overlay on location-screen


Jonta:
- Got to slide 3 on Welcome wizard
- Got: Transparency at the bottom
- Expected: Prettier, non-glitchy-looking. =)

![syncthing-transparent-overlay](https://user-images.githubusercontent.com/359397/112546617-eecfe400-8db9-11eb-8f77-811244306dda.jpg)

### Version Information

    App Version: 1.14.0
    Syncthing Version: v1.14.0
    Android Version: 10
    Android theme: Dark

PotatoGod123:
I'll Work on this! Do you have a way to contact you?

Jonta:
> I'll Work on this! Do you have a way to contact you?

@PotatoGod123 This thread is fine. =)

PR #1657, opened 2021-05-21
(WIP)#1626 welcome wizard location screen 

PotatoGod123:
Ran into building issues so I can't properly test out the changes, Using Android Studio to build. Could we get into contact for more information.

### DIFF
--- a/app/src/main/res/layout/activity_firststart_slide3.xml
+++ b/app/src/main/res/layout/activity_firststart_slide3.xml
@@ -60,7 +60,7 @@
 
             <TextView
                 android:layout_width="wrap_content"
-                android:layout_height="40dp"
+                android:layout_height="wrap_content"
                 android:layout_gravity="center_horizontal"
                 android:layout_marginTop="@dimen/desc_marginTop"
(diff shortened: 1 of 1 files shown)


imsodin:
If you do UI changes, please post screenshots. Which clearly you can only do if the build is working, so that needs to be fixed first. Likely culprit is recent android studio shipping java11 and that fails due to some deprecated classes being depended upon - you need java 8. Or figure out what is depending upon it - I think it was gradle, and updating that will likely fix this. But also means changing a bunch of deprecated things. Again, if you want to tackle this you are very welcome to do so.

And please explain why you choose an absolute height. Making the "bottom bar" not fully transparent or constraining the text view to above that bar seem like more sensible options to me (without thinking a lot about it).

st-review:
:robot: *beep boop*

I'm going to close this pull request as it has been idle for more than 90 days.

This is not a rejection, merely a removal from the list of active pull requests that is periodically reviewed by humans. The pull request can be reopened when there is new activity, and merged once any remaining issues are resolved.

PR #1674, opened 2021-07-11
Add folder-device with enc-password (fixes #1673)

### LINKED ISSUE #1673: Encryption password is overwritten for any change to any folder using the Android UI
I noticed constant crashes when I use encrypted (untrusted) folders. When I set up a folder, I use the Web UI to set the password. But later I found that the password was empty.

This crashes Syncthing and I see an error on the other end:

```
... Failure checking encryption consistency with device ... for folder ...: remote expects to exchange plain data, but is configured to be encrypted
```

Seems like if I use the Android UI (not the Web UI), open a folder than make any change to any folder, it clears the password for all folders.

Steps to reproduce:

- Open the Web UI and add a folder (Test), share it with a device and set a password
- Go back to the Android UI, open any folder and make any change (for the test I opened a different folder (Folder2) and unshared it)
- Back on the Web UI, the original folder (Test) is still shared with the device but with no password

### Version Information

    App Version: 1.18.0
    Syncthing Version: v1.18.0
    Android Version: Android 10

imsodin:
I started looking into using the new "modular" config api (just patching the parts the wrapper cares about), however that required quite a lot of changes all over the place so I didn't get anywhere in the short amount of time I had. This hopefully fixes the issues people are seeing with encryption passwords and potentially other things, as the previous code wrongly used "normal devices" instead of "folder-devices".

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/model/Folder.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/model/Folder.java
@@ -20,7 +20,7 @@ public class Folder {
     public String type = Constants.FOLDER_TYPE_SEND_RECEIVE;
     public boolean fsWatcherEnabled = true;
     public int fsWatcherDelayS = 10;
-    private List<Device> devices = new ArrayList<>();
+    private List<Folder.Device> devices = new ArrayList<>();
     public int rescanIntervalS;
     public final boolean ignorePerms = true;
     public boolean autoNormalize = true;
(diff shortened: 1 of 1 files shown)


AudriusButkevicius:
I guess this is as good as it gets, namely whack-a-mole.

This is not for now, but just a thought how this could be fixed.
We could switch Android to use a non-lossy serialisation format (for example protobuf, by adding accept-encoding: application/x-protobuf`), which then preserves fields it doesn't know how to parse in the round-trip.

AudriusButkevicius (approved):

PR #1678, opened 2021-07-25
Adjust to API change regarding pending folders/devices

imsodin:
Ignoring devices is broken since https://github.com/syncthing/syncthing/pull/6443 and causes crashes.

This removes all the state that has previously been kept on the pending devices, and instead uses the required info (labels, addresses) from the events funnelled through the activity.


app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
```diff
@@ -117,11 +118,9 @@ public void onEvent(Event event) {
                 );
                 break;
             case "FolderRejected":
-                onFolderRejected(
-                    (String) event.data.get("device"),          // deviceId
-                    (String) event.data.get("folder"),          // folderId
-                    (String) event.data.get("folderLabel")      // folderLabel
-                );
+                for (Map<String,String> m : (List<Map<String,String>>) event.data.get("added")) {
```
AudriusButkevicius, line 121: On the phone, hence hard to check, but seems strange that rejected has an added field?

AudriusButkevicius (dismissed):
In principle, lgtm, I assume you tested this.

AudriusButkevicius (approved):

PR #1684, opened 2021-08-13
Fixed README.md recognised as binary (fixes #1687)

Smithienious:

PR #1691, opened 2021-09-03
Fixed the README.md bug, which leads to README.md file not render the markdown correctly.

durwankurnaik:
I changed the starting line from 'syncthing-android' to 'Syncthing Android' and commited the changes, if this PR doesn't works for you, you may do the same thing with your main branch's README file directly.


tomasz1986:
My guess would be that the name is probably not the issue. The actual problem is likely fixed by https://github.com/syncthing/syncthing-android/pull/1684.

durwankurnaik:
Can I ask what was the issue which was fixed using #1684 ?

imsodin:
As far as I see it was the exact same change in that PR.

PR #1695, opened 2021-09-17
Update NDK

AudriusButkevicius:


AudriusButkevicius:
This needs to be force merged, and we need to rebuild the build image for the PR to pass.

AudriusButkevicius:
This needs to be force merged, and we need to rebuild the build image for the PR to pass.

Issue #1721, opened 2021-10-29
Bump android api level to 30


imsodin:
I want to collect all requirements to release with android api level 30. I half ignored half forgot that the google play deadline to do the switch is November, i.e. next week. The scary change is the scoped storage one. Which is partially good (SD card support #29), partially bad (performance) and also scary (some reports in the fork app that did the switch of disappearing data).

1. Investigate if accessing now restricted paths (e.g. `Android/data` and `Android/obb`) generates an error such that `!os.IsNotExist(err)`. If not, add some smartness or at least pause folders and make them receive-only together with a wizzard to make users aware, to make sure that previously scanned files do not get detected as deleted on scan.
https://developer.android.com/training/data-storage/manage-all-files
https://developer.android.com/training/data-storage
https://github.com/Catfriend1/syncthing-android/commit/d0af150ebaba18c999e0060e555e48d53112210c
https://forum.syncthing.net/t/android-11-all-files-access-for-the-syncthing-app/14651
https://github.com/Catfriend1/syncthing-android/issues/568
https://github.com/Catfriend1/syncthing-android/pull/618/files

2. Add wizzard page to request all file access (respectively adapt existing one).

3. Investigate and fix networking restriction issues.
https://github.com/syncthing/syncthing-android/issues/1663
https://github.com/syncthing/syncthing-android/issues/1628
https://github.com/Catfriend1/syncthing-android/pull/800
Probably more a curiosity due to lack of time: A go app running on android, which had some of this problems, and uses java bindings to fix them: https://github.com/tailscale/tailscale/issues/2293

2. Get all file access approved in google play
https://developer.android.com/training/data-storage/manage-all-files#all-files-access-google-play

5. Find out if there's more things that need fixing - @Catfriend1 ? :)

While I want to keep publishing syncthing-android, I cannot guarantee anything here due to time constraints (I can guarantee I won't be able to do it any time close to the deadline :) ).

**Please if you like syncthing-android and have a bit of time to spare, help out. I am very happy to collaborate and guide!**

tomasz1986:
I can try to help if we yet again need to write some legalese regarding all file access in the privacy policy. Others would have to join in when it comes to the other points though :pensive:.

Catfriend1:
[...]

PR #1724, opened 2021-11-01
Bump android api level to 30 (ref #1721)

imsodin:
Based on https://github.com/Catfriend1/syncthing-android/commit/545a9ffb1dd419d662ebc280eb05bd9352a56a56. Builds, but entirely untested. Also lacks necessary adjustments to wizards and networking stuff, see https://github.com/syncthing/syncthing-android/issues/1721


imsodin:
It's ready and working for me on Android 10 and 11. The intention was to spend as little time on UI as possible, thus "just add a start wizard slide". The "just" was more annoying than expected. And the android SAF URI to filesystem path conversion (broken targeting api lvl 30) and syncthing service shutdown needed fixing (broken before already, but maybe without consequences, now it broke the syncthing restart after db reset).

Screenshot_20211212-233215

Please check the message text, both whether it's generally understandable/helpful and small language issues.

tomasz1986:
I sounds understandable to me 👍.

One thing that I think should be added though is that database reset iself is not risk-free, e.g. when the devices weren't 100% in sync beforehand, there will be conflicts and/or files deleted on other devices in the meantime will be pushed there again (and there could possibly even be data loss if conflicts are disabled).

app/src/main/res/values/strings.xml
```diff
@@ -21,11 +21,19 @@ Please report any problems you encounter via Github.</string>
...
 
+    <!-- Slide 4 -->
+    <string name="api_level_30_title">Upgrade Information</string>
+    <string name="api_level_30_desc">With the latest update syncthing-android is targeting API level 30, which changes the behaviour of Android\'s filesystem. Good news: On Android 11 and newer it is now able to write to an external SD card. Bad news: It may be slower and some paths will be hidden. Due to these hidden paths, the database needs to be reset to prevent data loss. This will cause all your data to be hashed again, but it doesn\'t need to be transferred from remote devices. It also may create conflicts and re-create recently deleted files if you aren't fully in sync right now. Depending on how much data you have, this process can take quite a long time.\n\n Please press the button below to reset the database and start Syncthing.</string>
```
phanirithvij, line 32:
```suggestion
    <string name="api_level_30_desc">With the latest update syncthing-android is targeting API level 30, which changes the behaviour of Android\'s filesystem. Good news: On Android 11 and newer it is now able to write to an external SD card. Bad news: It may be slower and some paths will be hidden. Due to these hidden paths, the database needs to be reset to prevent data loss. This will cause all your data to be hashed again, but it doesn\'t need to be transferred from remote devices. It also may create conflicts and re-create recently deleted files if you aren\'t fully in sync right now. Depending on how much data you have, this process can take quite a long time.\n\n Please press the button below to reset the database and start Syncthing.</string>
```
The TeamCity build fails because of `aren't`
```sh
[Step 1/1]      /opt/tcagent/syncthing-3-work/19a8ed19ae4d1032/app/src/main/res/values/strings.xml:32:5-759: AAPT: error: unescaped apostrophe in string
```

AudriusButkevicius:
Will do in the evening.

app/src/main/java/com/nutomic/syncthingandroid/util/FileUtils.java
```diff
@@ -137,6 +136,19 @@ private static String getVolumePath(final String volumeId, Context context) {
...
+    private static String volumeToPath(Object storageVolumeElement, Class<?> storageVolumeClazz) throws Exception {
+        try {
+            Method getDir = storageVolumeClazz.getMethod("getDirectory");
+            File file = (File) getDir.invoke(storageVolumeElement);
+            return file.getPath();
+        } catch (NoSuchMethodException e) {
+            Method getPath = storageVolumeClazz.getMethod("getPath");
+            return (String) getPath.invoke(storageVolumeElement);
+        }
+    }
```
AudriusButkevicius, line 150: Would be useful to add a comment which one targets which API.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingService.java
```diff
@@ -546,6 +549,14 @@ private void shutdown(State newState, SyncthingRunnable.OnSyncthingKilled onKill
...
             mSyncthingRunnable = null;
         }
+        if (startupTaskIsRunning()) {
+            mStartupTask.cancel(true);
+            Log.v(TAG, "Waiting for mStartupTask to finish after cancelling ...");
+            try {
+                mStartupTask.get();
+            } catch (Exception e) { }
+            mStartupTask = null;
+        }
```
AudriusButkevicius, line 559: I assume this is a bug fix for rapid stop/start leading to some bad state?

app/src/main/java/com/nutomic/syncthingandroid/activities/MainActivity.java
```diff
@@ -258,20 +260,27 @@ public void onCreate(Bundle savedInstanceState) {
...
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
             startForegroundService(serviceIntent);
         } else {
             startService(serviceIntent);
         }
 
+        // In case start_into_web_gui option is enabled, immediately open that.
+        if (mPreferences.getBoolean(Constants.PREF_START_INTO_WEB_GUI, false)) {
+            startActivity(new Intent(this, WebGuiActivity.class));
+        }
```
AudriusButkevicius, line 275: Is this also a bug fix?

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
```diff
@@ -33,28 +39,46 @@
...
+            Slide.STORAGE,
+            Slide.LOCATION,
+            Slide.API_LEVEL_30,
+    };
+    private static int[] mLayouts = new int[]{
+            R.layout.activity_firststart_slide_intro,
+            R.layout.activity_firststart_slide_storage,
+            R.layout.activity_firststart_slide_location,
+            R.layout.activity_firststart_slide_api_level_30,
+    };
```
AudriusButkevicius, line 66: This is a bit grim.

You could rely on the enum definition order for the ordering, and then do the following for the layout ids:

https://www.baeldung.com/java-enum-values

app/src/main/res/values/strings.xml
```diff
@@ -21,11 +21,19 @@ Please report any problems you encounter via Github.</string>
...
 
+    <!-- Slide 4 -->
+    <string name="api_level_30_title">Upgrade Information</string>
+    <string name="api_level_30_desc">With the latest update syncthing-android is targeting API level 30, which changes the behaviour of Android\'s filesystem. Good news: On Android 11 and newer it is now able to write to an external SD card. Bad news: It may be slower and some paths will be hidden. Due to these hidden paths, the database needs to be reset to prevent data loss. This will cause all your data to be hashed again, but it doesn\'t need to be transferred from remote devices. It also may create conflicts and re-create recently deleted files if you aren\'t fully in sync right now. Depending on how much data you have, this process can take quite a long time.\n\n Please press the button below to reset the database and start Syncthing.</string>
```
AudriusButkevicius, line 32: Worth noting that it will be an intensive process that will make the phone hot/drain battery?
Wouldn't want to press that button while on my last 3% while waiting for uber

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
```diff
@@ -136,58 +159,105 @@ public void onBtnBackClick() {
...
-    private void addBottomDots(int currentPage) {
-        mDots = new TextView[mLayouts.length];
+    private boolean isFirstStart() {
+        return mPreferences.getBoolean(Constants.PREF_FIRST_START, true);
+    }
 
-        int[] colorsActive = getResources().getIntArray(R.array.array_dot_active);
-        int[] colorsInactive = getResources().getIntArray(R.array.array_dot_inactive);
+    private boolean upgradedToApiLevel30() {
+        boolean upgraded = mPreferences.getBoolean(Constants.PREF_UPGRADED_TO_API_LEVEL_30,false);
```
AudriusButkevicius, line 205: missing space after `,`, also could be folded into the if, given it's never reused.

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
```diff
@@ -136,58 +159,105 @@ public void onBtnBackClick() {
...
 
-    private int getItem(int i) {
-        return mViewPager.getCurrentItem() + i;
+    private void setActiveBottomDot(int currentPage) {
+        int[] colorsActive = getResources().getIntArray(R.array.array_dot_active);
+        int[] colorsInactive = getResources().getIntArray(R.array.array_dot_inactive);
+        for (int i = 0; i < mDots.length; i++) {
+            mDots[i].setTextColor(colorsInactive[currentPage]);
+        }
+        mDots[currentPage].setTextColor(colorsActive[currentPage]);
```
AudriusButkevicius, line 252: fuck ui development.

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
```diff
@@ -230,26 +300,42 @@ public Object instantiateItem(ViewGroup container, int position) {
...
+
+                case API_LEVEL_30:
+                    Button btnResetDatabase = (Button) view.findViewById(R.id.btnResetDatabase);
+                    btnResetDatabase.setOnClickListener(new View.OnClickListener() {
+                        @Override
+                        public void onClick(View v) {
+                            // The main-activity will actually do the db reset.
+                            mResetDatabase = true;
+                            mPreferences.edit().putBoolean(Constants.PREF_UPGRADED_TO_API_LEVEL_30, true).apply();
+                            onBtnNextClick();
```
AudriusButkevicius, line 335: What if I kill the app after this slide, or it ends up with syncthing being disabled due not being connected to wifi or something, etc, I exit, come back, come back later, it then it wouldn't set the flag and potentially blow up in my face.

Is there any harm in calling `syncthing -reset-database` or something of that like right here?
Also wonder if it's not just easier to `rm -rf` the database directory oppose to relying on actually launching the binary etc.

Feels like the flag should only be set after the actual launch that did the reset.

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
```diff
@@ -136,58 +159,105 @@ public void onBtnBackClick() {
...
+            case INTRO:
+                return !isFirstStart();
+            case STORAGE:
+                return PermissionUtil.haveStoragePermission(this);
+            case LOCATION:
+                return hasLocationPermission();
+            case API_LEVEL_30:
+                // Skip if running as root, as that circumvents any Android FS restrictions.
+                return upgradedToApiLevel30()
+                        || PreferenceManager.getDefaultSharedPreferences(this).getBoolean(Constants.PREF_USE_ROOT, false);
```
AudriusButkevicius, line 231: What about a case where I skip over this as I am running as root, then disable root, and everything potentially gets nuked?

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
```diff
@@ -281,61 +367,80 @@ private void startApp() {
...
+            if (componentName != null) {
+                // Launch "Allow all files access?" dialog.
+                startActivity(intent);
+                return;
+            }
+            Log.w(TAG, "Request all files access not supported");
+        } catch (ActivityNotFoundException e) {
+            Log.w(TAG, "Request all files access not supported", e);
+        }
+        Toast.makeText(this, R.string.dialog_all_files_access_not_supported, Toast.LENGTH_LONG).show();
```
AudriusButkevicius, line 418: Might be interesting which branch we've hit?

Also, I guess the user can just press cancel or something, but I guess that wouldn't allow them to skip to the next slide, as it checks.

app/src/main/java/com/nutomic/syncthingandroid/util/FileUtils.java
imsodin: Added some comments, not entirely sure on it. I just debugged a weird issue to finde `getPath` missing. `getDirectory` is documented as added in api level 30, but there's no information about `getPath`. I just kept that in case it's still the one present on some older android versions.

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
imsodin: That's an experimental setting - to me using root means you are ok with anything. I really don't want to handle edge cases around root.

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
imsodin: There was some problem with starting syncthing here, and not in the main activity. Already forgot again what it was. I'll set the flag only after that ran successfully.

app/src/main/java/com/nutomic/syncthingandroid/service/SyncthingService.java
imsodin: Yes. I got into such a state frequently with the db reset on startup, which causes a restart.

app/src/main/java/com/nutomic/syncthingandroid/activities/MainActivity.java
imsodin: No, moved from first start wizard. Which actually is not only unnecessary, but worse: This should only happen when newly opening the app, not when e.g. coming back to it through the android app switcher.

app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
imsodin: It's already discernable in the logs - I wouldn't display any such low-level info in a toast. And generally it's mainly a very defensive measure, not likely that either case is hit.

And yes it does check, display a message and prevent continuation, if the user tries to continue without having approved the access.

imsodin:
@AudriusButkevicius Thanks a lot for the review, I think I addressed all the points.

AudriusButkevicius (approved):

PR #1785, opened 2022-05-07
build: Use python as command (fixes #1783)

### LINKED ISSUE #1783: gradlew should use `python` instead of `python3` when compiling in Windows
I'm not sure if this is something that can be controlled within the repository or rather something that needs to be fixed in Gradle itself, but currently, when trying to use `gradlew` under Windows, it tries to execute the `python3` command, which isn't used at all by the Windows release of Python, which has always used just `python` (or `python.exe` specifically).

A clunky workaround is to create a symlink or hardlink named `python3.exe` linked to the proper `python.exe` executable. The caveat here is that the path changes upon each Python version update, so it needs to be created dynamically, e.g. by using a batch script as follows.

```
for %%a in (python.exe) do mklink /h "python3.exe" "%%~$path:a"
```

**Edit**:

I think the culprit is https://github.com/syncthing/syncthing-android/commit/c505c1b5257b70eb8938839c0d44d72ec2c82806.

imsodin:
requires https://github.com/syncthing/syncthing-android/pull/1786

### DIFF
--- a/syncthing/build.gradle
+++ b/syncthing/build.gradle
@@ -2,7 +2,7 @@ task buildNative(type: Exec) {
     environment "NDK_VERSION", "$ndkVersionShared"
     inputs.dir("$projectDir/src/")
     outputs.dir("$projectDir/../app/src/main/jniLibs/")
-    executable = 'python3'
+    executable = 'python'
     args = ['-u', './build-syncthing.py']
 }
 


imsodin:
Closing in favor of https://github.com/syncthing/syncthing-android/pull/1795

PR #1800, opened 2022-07-03
Update strings.xml

efraletti:
added some missing spaces


AudriusButkevicius:
Translations are managed via transifex, these files have no meaning and get overwritten from transifex.

PR #1803, opened 2022-07-10
Don't call mediastore on file changes on android >=10 (fixes #1801)

### LINKED ISSUE #1801: Spurious file deletion due to MediaStore/ContentResolver delete race
Syncthing Version v1.20.3
Syncthing-Android Version 1.20.3
Android 12, Pixel 6 Pro (SP2A.220505.002)

This morning I discovered an issue where some files of mine, stored in a SyncThing folder, went mysteriously missing. I believe this _could_ also be the cause of some issues I've seen unsolved in the forums, such as:
https://forum.syncthing.net/t/renaming-caused-file-deletion-bug/17591 (both the original issue, and most suggestively, the similar issue described by the last poster, whose setup is identical to mine.)
https://forum.syncthing.net/t/syncthing-destroyed-at-least-two-photos/16242/8

The issue I observe: When a file is deleted and then rapidly recreated, if it is being synced to a Syncthing Android client, sometimes (maybe half the time), the Android client will spuriously delete the file within a few seconds to a minute or so afterwards. Example:

Screen Shot 2022-07-09 at 6 06 22 PM

The cause I suspect: When Syncthing Android finishes receiving a sync event, if it is a deletion, it calls out to MediaStore/ContentResolver, to cause the Android media system to de-index the file.

https://github.com/syncthing/syncthing-android/blob/d376ed793839983446e6b2a2c9a5f372706119cb/app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java#L144

HOWEVER, as far as I can tell, it is _not_ safe to use ContentResolver.delete() in this way, because it will not only delete the file from the media database, it will also _delete the file itself_ if it's still on the filesystem. And it looks like (for whatever reason -- I don't know if the delay is in Syncthing Android or in MediaStore somewhere), the call to delete the item from the MediaStore can take over 10 seconds to complete. Which means that, if the file gets recreated as quickly as possible, there is a race, and the ContentResolver.delete() can finally finish processing _after_ Syncthing has recreated the file, and re-delete it. (At which point Syncthing will detect this as a deletion and remove the file on both ends.)

To reproduce easily:
- Create a test file (from either end) and let it sync between the Android client and another client (the "desktop").
- Delete the file on the desktop client. Wait for the delete to propagate to the Android client.
- Recreate the file manually _on the Android side_ as soon as it's deleted. (This is faster than waiting for it to re-sync from the desktop side, so it's easier to win the race this way.)
- Wait 1-15 seconds.
- You will find that the file has disappeared again.

To fix:
... honestly I don't really understand the Android ContentProvider/MediaStore/MediaProvider/ContentResolver machinery or how it's supposed to work. I think this issue _may_ be new with Android 10/11, which changed how external storage access is managed, but I'm really not sure. The documentation is all uselessly vague. 

I assume the delete() call was added to fix a problem like the one described in https://stackoverflow.com/questions/29875744/deleted-file-resurrects-after-action-media-scanner-scan-file-intent/ , which is linked in the source. I am not an Android dev, and I don't know what a better fix is. But if you're with me this far, let me know if you'd like me to try to help figure that part out too.

Thanks!

imsodin:
Also use `=` not `LIKE` because that's what we really want to check. Likely still wrong, depending on if the mediastore provider is case sensitive or not (couldn't find any info on it). However given there's been no complaints about this before, I am inclined to leave it as is for Android < 10.

I found some unofficial ramblings (SO) about the content-resolver only handling database contests pre scoped storage, and now also deleting the underlying files. The latter clearly is very bad, we don't want that. The intention of this code apparently is to make media appear and disappear immediately in apps (e.g. gallery) without a rescan. Maybe the disappearing doesn't work anymore now - it didn't work for me before either. And even if it did, that's a much better deficiency than spurious deletions.

I could repro case-renaming a new file leading to it getting deleted (`echo this > file.txt; mv file.txt File.txt`). With the SQL change it still happened, when skipping the delete thing it didn't happen anymore.

Anyone wanting to test is very welcome to, apk can be found here (click login as guest to get in without an account): https://build.syncthing.net/buildConfiguration/SyncthingAndroid_Build/123628?buildTab=artifacts#%2Fapk

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java
@@ -7,6 +7,7 @@ import android.content.Intent;
 import android.content.SharedPreferences;
 import android.media.MediaScannerConnection;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Handler;
 import android.os.Looper;
 import android.provider.MediaStore;
@@ -134,11 +135,23 @@ public class EventProcessor implements  Runnable, RestApi.OnReceiveEventListener
                     MediaScannerConnection.scanFile(mContext, new String[]{updatedFile.getPath()},
                             null, null);
                 } else {
+                    // Starting with Android 10/Q and targeting API level 29/removing legacy storage flag,
+                    // reports of files being spuriously deleted came up.
+                    // Best guess is that Syncthing directly interacted with the filesystem before,
+                    // and there's a virtualisation layer there now. Also there's reports this API
+                    // changed behaviour with scoped storage. In any case it now does not only
+                    // update the media db, but actually delete the file on disk. Which is bad,
+                    // as it can race with the creation of the same file and thus delete it. See:
+                    // https://github.com/syncthing/syncthing-android/issues/1801
+                    // https://github.com/syncthing/syncthing/issues/7974
+                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+                        break;
+                    }
                     // https://stackoverflow.com/a/29881556/1837158
                     Log.i(TAG, "Deleted file from MediaStore: " + updatedFile.toString());
                     Uri contentUri = MediaStore.Files.getContentUri("external");
                     ContentResolver resolver = mContext.getContentResolver();
-                    resolver.delete(contentUri, MediaStore.Images.ImageColumns.DATA + " LIKE ?",
+                    resolver.delete(contentUri, MediaStore.Images.ImageColumns.DATA + " = ?",
                             new String[]{updatedFile.getPath()});
                 }
                 break;


AudriusButkevicius (approved):

bt90:
@Catfriend1 does this affect the fork aswell?

Catfriend1:
> @Catfriend1 does this affect the fork aswell?

Maybe... So I need to check with the reproducer mentioned. Code is like this atm : https://github.com/Catfriend1/syncthing-android/blob/5dd054a643ef2cdf07faa6a51e542401a87214fe/app/src/main/java/com/nutomic/syncthingandroid/service/EventProcessor.java#L418

Catfriend1:
Related: https://stackoverflow.com/questions/57093479/get-real-path-from-uri-data-is-deprecated-in-android-q

Catfriend1:
@imsodin Yes, it affects the fork as well. Log:
```
2022-07-13 14:24:06.406 3439-3439/com.github.catfriend1.syncthingandroid.debug I/EventProcessor: onItemFinished: MediaScanner, Rescanning file: /storage/emulated/0/Pictures/file.txt
2022-07-13 14:24:26.420 3439-3439/com.github.catfriend1.syncthingandroid.debug I/EventProcessor: onItemFinished: MediaStore, Deleting file: /storage/emulated/0/Pictures/file.txt
2022-07-13 14:24:26.425 3439-3439/com.github.catfriend1.syncthingandroid.debug I/EventProcessor: onItemFinished: MediaScanner, Rescanning file: /storage/emulated/0/Pictures/File.txt
```

Catfriend1:
I've changed the behaviour in the fork to the following - regarding the above mentioned reproducer:
```
2022-07-13 14:30:43.091 5352-5352/com.github.catfriend1.syncthingandroid.debug I/EventProcessor: onItemFinished: MediaStore, Skip file deletion because file exists: /storage/emulated/0/Pictures/file.txt
2022-07-13 14:30:43.102 5352-5352/com.github.catfriend1.syncthingandroid.debug I/EventProcessor: onItemFinished: MediaScanner, Rescanning file: /storage/emulated/0/Pictures/File.txt
```

imsodin:
I am leaving as is here. I read statements indicating that the delete is potentially deferred for a while, so there's still a risk of a race: Syncthing deletes -> mediastore initiates deletion -> Syncthing creates -> mediastore deletes. Result is the file is gone. I don't want the wrapper to delete stuff (or generally modify actual files). If there's a way to update the mediastore without changing underlying files, I am happy to accept a contribution doing that.

PR #1817, opened 2022-08-04
Add delay for charging detection (fixes #1595)

### LINKED ISSUE #1595: Unreliable charging detection
My "Run Conditions": run on any Wi-Fi + only on AC.
Tested on: Galaxy S10+ on stock Android 11, Galaxy S5 on Android 10 LineageOS.

On both devices AC vs battery detection logic is extremely unreliable, every now and then it goes to inverted state, i.e. syncthing service is being started when disconnected from power and being killed when connecting to AC.

Validated by running "logcat | grep -E 'RunConditionMonitor|SyncthingService|BatteryService'" in wireless ADB shell and plugging AC powersupply in and out repeatedly.

Looking at the RunConditionMonitor.java and logcat output, here is my guess: ACTION_POWER_CONNECTED and ACTION_POWER_DISCONNECTED are being received just fine and then code is trying to process BatteryManager.EXTRA_PLUGGED value, which is a bad idea to do instantly, since a lot of times battery status hasn't updated yet.

Quick google returned similar issue: https://stackoverflow.com/a/11868558

Perhaps this call to updateShouldRunDecision() should be fired async, schedule it for execution in a few seconds, let the battery status settle first.
https://github.com/syncthing/syncthing-android/blob/00edf5f868512301ec770caeb5d07081dba3f1d1/app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java#L121

nmaggioni:
Fix for #1595 pulled in from [syncthing-fork](https://github.com/Catfriend1/syncthing-android/commit/0cd20ff4de68fac9511eb0670eb858007ba89723) as proposed in the issue.

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java
@@ -14,6 +14,7 @@ import android.net.wifi.WifiInfo;
 import android.net.wifi.WifiManager;
 import android.os.BatteryManager;
 import android.os.Build;
+import android.os.Handler;
 import android.os.PowerManager;
 import androidx.annotation.Nullable;
 import android.util.Log;
@@ -118,8 +119,13 @@ public class RunConditionMonitor {
         public void onReceive(Context context, Intent intent) {
             if (Intent.ACTION_POWER_CONNECTED.equals(intent.getAction())
                     || Intent.ACTION_POWER_DISCONNECTED.equals(intent.getAction())) {
-                SystemClock.sleep(5000);
-                updateShouldRunDecision();
+                Handler handler = new Handler();
+                handler.postDelayed(new Runnable() {
+                    @Override
+                    public void run() {
+                        updateShouldRunDecision();
+                    }
+                }, 5000);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java
```diff
@@ -118,6 +118,7 @@ private class BatteryReceiver extends BroadcastReceiver {
         public void onReceive(Context context, Intent intent) {
             if (Intent.ACTION_POWER_CONNECTED.equals(intent.getAction())
                     || Intent.ACTION_POWER_DISCONNECTED.equals(intent.getAction())) {
+                SystemClock.sleep(5000);
```
imsodin, line 121: I believe this is running on the main/UI thread, i.e. freezes the app for 5s - that's not acceptable. Kicking off a timer (assume the decision doesn't touch UI) or using some handler should work.

nmaggioni: Done, thanks for the reminder. Build environment is broken atm though, I can't test this change on a real device.

imsodin (approved):

Issue #1694, opened 2021-09-12
"Syncthing is disabled: Not connected to wifi" - I AM!


rmst:
Same problem as in https://github.com/syncthing/syncthing-android/issues/1280 remains for me. Would have reopened the issue but it's locked. The problem seems to go away when I disable Netguard VPN even though Syncthing is whitelisted and sometimes works even with Netguard activated

AudriusButkevicius:
This looks more like a support request than a clear bug or feature request. Please use the forum - https://forum.syncthing.net/ - for support, first by searching for any similar issues and then by posting a new one when required. The forum has a lot more eyes and friendly people on it than the issue tracker.

Thanks!

AudriusButkevicius:
Also see: https://github.com/syncthing/syncthing-android/issues/1568

Issue #1280, opened 2019-02-24
"Syncthing is disabled: Not connected to wifi" - I AM!


snow-frog:
I am connected on Wifi and no settings have been changed, but syncthing refuses to exit this condition. It was working fine until tonight. Fully stock phone, Android 8.0.0, Syncthing 0.10.17

capi:
I'm sorry to hear this. I've reworked the decission logic somehow to get the indication to the UI why we are not connected. Can you provide an `adb` log so that we can debug why it is considering your status as "not connected"?

mlda065:
I just had the same issue.

I'm using a VPN client. ([Private Internet Access](https://play.google.com/store/apps/details?id=com.privateinternetaccess.android))
When I disable the VPN, Syncthing starts.

I have already whitelisted Syncthing in my VPN app, so supposedly Syncthing shouldn't notice the VPN. But clearly there's still something different to no-VPN. Not sure what.

I haven't had time to check my adb logs. I'll do that later.

capi:
When connected to a WiFi, Android reports the WiFi as not connected. I know that there is a way to detect that the VPN is connected via a certain WiFi, because Tasker can do it ("Active" setting "Yes", "No", "Any", you need the "any" setting). But I need to have a look how they do it.

Catfriend1:
This is interesting, I guess it depends which vpn app you're using. I'm running NetGuard VPN here and this doesn't cause problems in STA's WiFi detection.

mlda065:
I think this bug is fixed. About a week after posting this an update fixed the problem. I'm assuming that was a deliberate patch.

shawn1thompson:
I am having this issue with ProtonVPN. I have enabled split tunnelling in the Proton app to no avail.

PR #1821, opened 2022-08-12
Run without (detected) network

fps:
There have been issues like https://github.com/syncthing/syncthing-android/issues/1694 or https://github.com/syncthing/syncthing-android/issues/1280 (and just very recently: https://github.com/syncthing/syncthing-android/issues/1822)

I ran into the same issue on my device (lineageos on sony xa2.) In my case syncthing refuses to run as soon as I start a wireguard tunnel. But instead of opening another ticket i figured I could try adding a workaround myself.

So this pull request adds another run condition entry: "Run without (detected) network" which short circuits the connectivity check.

This has not been tested yet, as I can't find the damn USB cable right now. It compiles and runs on the emulator though. Before going forward with this I wanted to test the waters and see if there is any interest for this at all..

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
@@ -104,6 +104,7 @@ public class SettingsActivity extends SyncthingActivity {
         private Preference         mCategoryRunConditions;
         private CheckBoxPreference mStartServiceOnBoot;
         private ListPreference     mPowerSource;
+        private CheckBoxPreference mRunOnNoNetwork;
         private CheckBoxPreference mRunOnMobileData;
         private CheckBoxPreference mRunOnWifi;
         private CheckBoxPreference mRunOnMeteredWifi;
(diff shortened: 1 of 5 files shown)


fps:
Thinking about it a little bit more. Maybe this is too specific and another approach would be more general (and possibly generally useful): Add a preference to enable/disable the run conditions all together. This would make running and exiting the app completely manually controlled, which would 1. workaround the issue in my case as well 2. be possibly useful for working around other run condition issues. What do you think?

imsodin:
I think a global (temporary?) disable run-conditions switch would be clearer. The option right now is more like a "don't care about networking conditions" switch, right?  

Generally of course I'd prefer if runtime-conditions detection would be improved to cover your use case, but a clean and simple workaround like completely disabling run-conditions, while preserving the underlying conditions for when it's enabled again, seems reasonable.

tomasz1986:
Doesn't the already existing condition to "run when device is in flight mode" technically cover this case too? This would need to be properly tested, but in my experience, it seems to apply even when there's no actual flight mode enabled on the device.

I say this, because I used to experience problems when the phone was switching between different networks (e.g. 2G, 3G and LTE). In the vert short period during the switching, there was no network connectivity at all, which caused Syncthing to stop, and then immediately restart a second later when the network connection was back. I then enabled the "run when device is in flight mode" condition, and the stop/restart cycle was no more.

fps:
> Doesn't the already existing condition to "run when device is in flight mode" technically cover this case too? This would need to be properly tested, but in my experience, it seems to apply even when there's no actual flight mode enabled on the device.

It sadly hasn't helped in my case.

fps:
I'll close this for now in favor of the other PR: https://github.com/syncthing/syncthing-android/pull/1823

PR #1823, opened 2022-08-13
add preference to run always (ignore run conditions)

fps:
As I wrote in the comment https://github.com/syncthing/syncthing-android/pull/1821#issuecomment-1213453291 for https://github.com/syncthing/syncthing-android/pull/1821 it might be useful to have a preference to ignore all run conditions. This PR implements this. Tested in the emulator and seems to work
run_conditions_enabled
run_conditions_disabled
.

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
@@ -11,6 +11,7 @@ import android.preference.CheckBoxPreference;
 import android.preference.EditTextPreference;
 import android.preference.ListPreference;
 import android.preference.Preference;
+import android.preference.PreferenceGroup;
 import android.preference.PreferenceFragment;
 import android.preference.PreferenceManager;
 import android.preference.PreferenceScreen;
@@ -101,7 +102,8 @@ public class SettingsActivity extends SyncthingActivity {
         @Inject NotificationHandler mNotificationHandler;
         @Inject SharedPreferences mPreferences;
 
-        private Preference         mCategoryRunConditions;
+        private PreferenceGroup    mCategoryRunConditions;
+        private CheckBoxPreference mRunConditions;
         private CheckBoxPreference mStartServiceOnBoot;
         private ListPreference     mPowerSource;
         private CheckBoxPreference mRunOnMobileData;
@@ -167,6 +169,8 @@ public class SettingsActivity extends SyncthingActivity {
 
             addPreferencesFromResource(R.xml.app_settings);
             PreferenceScreen screen = getPreferenceScreen();
+            mRunConditions =
+                    (CheckBoxPreference) findPreference(Constants.PREF_RUN_CONDITIONS);
             mStartServiceOnBoot =
                     (CheckBoxPreference) findPreference(Constants.PREF_START_SERVICE_ON_BOOT);
             mPowerSource =
@@ -234,8 +238,14 @@ public class SettingsActivity extends SyncthingActivity {
 
             mCategorySyncthingOptions = findPreference("category_syncthing_options");
             setPreferenceCategoryChangeListener(mCategorySyncthingOptions, this::onSyncthingPreferenceChange);
-            mCategoryRunConditions = findPreference("category_run_conditions");
+            mCategoryRunConditions = (PreferenceGroup) findPreference("category_run_conditions");
             setPreferenceCategoryChangeListener(mCategoryRunConditions, this::onRunConditionPreferenceChange);
+
+            if (!mRunConditions.isChecked()) {
+                for (int index = 1; index < mCategoryRunConditions.getPreferenceCount(); ++index) {
+                    mCategoryRunConditions.getPreference(index).setEnabled(false);
+                }
+            }
 
             exportConfig.setOnPreferenceClickListener(this);
             importConfig.setOnPreferenceClickListener(this);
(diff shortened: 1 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -318,6 +318,9 @@ Please report any problems you encounter via Github.</string>
     <string name="run_conditions_title">Run Conditions</string>
     <string name="run_conditions_summary">Use the following options to decide when Syncthing will run.</string>
 
+    <string name="run_always_title">Run always</string>
+    <string name="run_always_summary">Ignore run conditions and run always. Use this if you want to manually control running of the app.</string>
```
imsodin, line 322: How does this give manual control? It means it runs all the time, no control whatsover, right? Seems like just the first sentence would cut it.

imsodin (commented):
Could you please disable (grey out) all the other settings when this one is active to make it visually clear it overrides everything.

app/src/main/res/values/strings.xml
fps: The manual control is then exercised by starting and exiting the app. I agree that this option is badly worded all together. I guess the clearest labeling would be "Enable Run Conditions" as title with the default being true (to keep the behavior from before this PR unless explicitly changed). If set to false all run condition options are greyed out (like you mentioned in your other comment) and the run conditions are ignored What do you think? Thanks for the review!

imsodin: makes sense

fps:
Hi, I reworked the code a little bit to implement the previously discussed naming changes and the greying out of the other preferences.

I am not too happy with the duplication of the logic to enable/disable the WIFI related preferences. Also I'm not an adroid developer by any stretch so I don't know if the iteration over the preferences by index in the run condition category is the smartest thing to do there.. Happy to get more feedback on this.. Thanks!

fps:
Oh, and I guess that's a little embarrassing: When I actually installed the apk on my device I noticed that the new version had working connectivity detection even with wireguard enabled (and with run conditions enabled.) Possibly the f-droid version I had installed previously (v1.20.4) was just too old. I'll leave this PR open anyways because disabling run conditions might be useful for future problems..

fps:
Actually: Ignore my previous comment. Sometimes the connectivity check just takes a while to kick in. So I actually still do need this preference to run syncthing with wireguard enabled. Sorry for the noise ;)

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
```diff
@@ -363,6 +373,17 @@ private void setPreferenceCategoryChangeListener(
 
         public boolean onRunConditionPreferenceChange(Preference preference, Object o) {
             switch (preference.getKey()) {
+                case Constants.PREF_RUN_CONDITIONS:
+                    boolean enabled = (Boolean) o;
+                    for (int index = 1; index < mCategoryRunConditions.getPreferenceCount(); ++index) {
+                        mCategoryRunConditions.getPreference(index).setEnabled(enabled);
+                        // Log.d(TAG, "setEnabled");
```
imsodin, line 380: Forgotten debug log?

imsodin (commented):
Looks good. I assume you tested it and it works as expected? Screenshot for UI changes is always appreciated (though here it's pretty obvious).

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
fps: Yes, thanks. Removed that line. In my tests it works fine. Of course I'm a bit wary to actually have my amateurish code in such a widely used project without further testing by other people. So if anyone else wants to give this a spin they would make me very happy..

fps:
I attached two screenshots as well. I'm not yet convinced this is the cleanest UI, so I'm open to suggestions. I'm not too well versed in the android UI guidelines (not at all) and I'm not too much of an android dev (or even user), so no offence taken if I missed an obviously better way to present this..

tomasz1986:
What do you think about making the explanation a little clearer, e.g. by changing

```
<string name="run_conditions_summary">Use the following options to decide when Syncthing will run.</string>
```

to

```
<string name="run_conditions_summary">Use the following options to decide when Syncthing will run. When disabled, Syncthing will run all the time.</string>
```

?

app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
imsodin: There's going to be a testing phase when the RC is released to the play beta channel. This seems safe enough (hey, the mainly active maintainer here isn't exactly an android specialist either ;) ).

imsodin:
I personally don't like longer strings, I think that just means they are more likely to be ignored. And I don't think the second sentence adds any crucial new info. And it means translation effort. So I'd tend against doing the change, but no very strong opinion.

imsodin (approved):
lgtm. The branch needs a rebasing though.

imsodin:
That makes no sense though: What's the purpose of something that is running but also not? Then again I am obviously biased as I know how it works. Anyway I'll defer to @fps or anyone else to chime in, I really don't care much.

fps:
I see @tomasz1986's point of some more clarification needed. But I'm not sure on how to reword that to make it more clear. Your suggestion, @tomasz1986, can also be a source of confusion (as "always run" can be misunderstood to really "always run", i.e. there's no more way to turn syncthing off at all, which is not true either.) 

I'd leave it as is right.. Thanks for the suggestion though :)

imsodin (approved):

tomasz1986:
The PR is already merged, but I'd like to ask one more question, as I can't really test this at the moment.

What happens when you upgrade from the older version of the app? Do the current settings persist (i.e. so that run conditions are enabled) or does it disable them?

fps:
The new preference defaults to <code>true</code>, so the original behaviour is preserved on an upgrade.

PR #1828, opened 2022-08-24
Rework the StateDialogActivity (fixes #1814) 

### LINKED ISSUE #1814: Not able to quit Syncthing Android when no Wi-Fi
Steps to reproduce:

1. Disconnect Wi-Fi
2. Open Syncthing Android
3. Syncthing disabled message is shown because there is no Wi-Fi
4. Clicking Exit
5. Syncthing window closes but Syncthing Android is still running and is not possible to quit.

I must connect to Wi-Fi to quit Syncthing Android.

fps:
Possible fix for https://github.com/syncthing/syncthing-android/issues/1814

* Add a neutral button to simply dismiss the dialog and close the app
(but leave it in the tasks list)
* Make the exit button stop the service and exit the app, removing it
from the task list

Capture

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/StateDialogActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/StateDialogActivity.java
@@ -2,17 +2,22 @@ package com.nutomic.syncthingandroid.activities;
 
 import android.content.Intent;
 import androidx.databinding.DataBindingUtil;
+
+import android.content.SharedPreferences;
 import android.os.Bundle;
 import android.os.Handler;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.core.app.ActivityCompat;
 import androidx.appcompat.app.AlertDialog;
+
+import android.preference.PreferenceManager;
 import android.view.View;
 
 import com.nutomic.syncthingandroid.R;
 import com.nutomic.syncthingandroid.databinding.DialogLoadingBinding;
 import com.nutomic.syncthingandroid.model.RunConditionCheckResult;
+import com.nutomic.syncthingandroid.service.Constants;
 import com.nutomic.syncthingandroid.service.SyncthingService;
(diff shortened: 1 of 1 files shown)


imsodin (changes requested):
The exist button has to be conditional on `PREF_START_SERVICE_ON_BOOT       = "always_run_in_background"` (https://github.com/syncthing/syncthing-android/blob/892fdea2ea359830980d598b4fd2a016829906d4/app/src/main/java/com/nutomic/syncthingandroid/service/Constants.java#L16) just like in the main/sidebar menu. Respectively it could also be disabled (with something like a tooltip if that exists on android) instead of hidden. Generally I do prefer if UI elements don't dis/reappear but witout an explanation of why it's disable it might be better if it's not there at all.

fps:
Thanks for the review! Yeah, you're right about it being conditional on that preference. I'll have to ponder this for a little while..

fps:
I decided to remove the "dismiss" button again and basically make the exit button in the dialog work like the exit button in the DrawerFragment (i.e. stop the service only when it's not set to start at boot and remove the task from the tasklist in either case.)

imsodin:
There's no exit button at all in the drawer menu when set to start at boot. And generally I don't think it's a good idea having a button (not deactivated) which does nothing.

fps:
Hi again, the exit button does exist in the StateDialogActivity dialog though, even when the service is set to start at boot. So do you argue the exit button should not exist in that dialog either in case the service is set to start at boot?

imsodin:
Doesn't exist for me:

Screenshot of Syncthing's menu without an exit entry

Sorry again for the super long reaction time, I'll need to make a point of dedicating some fixed time to Syncthing - I really like doing it, it's just hard to allocate the time (that sounds like BS even to myself :P ).

fps:
Hi! No problem at all. There is no hurry :) I guess it's me running an older version again.. I will think about the dialog a bit more then...

PR #1854, opened 2022-11-14
Remove copied device ID toast on android 13+

o13e:
>  In Android 12L (API level 32) and lower, we suggest alerting users that they’ve successfully copied by issuing a visual pop-up in-app widget (like Toasts or Snackbars) after copying.
> 
> To avoid duplicate displays of information, we strongly recommend removing any pop-up widget shown after an in-app copy for Android 13 and higher.
> If you show a copy confirmation snackbar in Android 13, the user sees duplicate messages.
> If you show a copy confirmation toast in Android 13, the user sees duplicate messages.
> 

https://developer.android.com/develop/ui/views/touch-and-input/copy-paste#duplicate-notifications

it show duplicate notification on android 13 at the moment:

Screenshot_20221114-030710_1

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/util/Util.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/util/Util.java
@@ -42,8 +42,10 @@ public class Util {
                 context.getSystemService(Context.CLIPBOARD_SERVICE);
         ClipData clip = ClipData.newPlainText(context.getString(R.string.device_id), id);
         clipboard.setPrimaryClip(clip);
-        Toast.makeText(context, R.string.device_id_copied_to_clipboard, Toast.LENGTH_SHORT)
+        if (android.os.Build.VERSION.SDK_INT < 33) {
+            Toast.makeText(context, R.string.device_id_copied_to_clipboard, Toast.LENGTH_SHORT)
                 .show();
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/nutomic/syncthingandroid/util/Util.java
```diff
@@ -42,8 +42,10 @@ public static void copyDeviceId(Context context, String id) {
                 context.getSystemService(Context.CLIPBOARD_SERVICE);
         ClipData clip = ClipData.newPlainText(context.getString(R.string.device_id), id);
         clipboard.setPrimaryClip(clip);
-        Toast.makeText(context, R.string.device_id_copied_to_clipboard, Toast.LENGTH_SHORT)
+        if (android.os.Build.VERSION.SDK_INT < 33) {
+            Toast.makeText(context, R.string.device_id_copied_to_clipboard, Toast.LENGTH_SHORT)
                 .show();
```
imsodin, line 47:
```suggestion
                    .show();
```

imsodin (commented):
Thanks for the contribution! Change looks good besides a minor indent nit. And just to check my understanding:  
Do I see that correctly that the blob with the thumbnail stile inner blobs is some android notification regarding copying the ID to clipboard? And somehow they decided to write out the copied string in that large white blob? Strange...

imsodin (approved):

PR #1865, opened 2022-11-26
readme: Update and add detailed steps to build instructions

tomasz1986:
Currently, the build instructions are very brief and don't explain the whole process of setting up the build environment in detail. Thus, add explicit steps on how to download required tools and prepare then for compilation.

Signed-off-by: Tomasz Wilczyński <twilczynski@naver.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,8 @@ Docker, you don't need to set them up separately.
         sure that the version numbers are up-to-date.
 2. Go (see https://docs.syncthing.net/dev/building#prerequisites for the
    required version)
-3. Java version 11
+3. Java version 11 (if not present in ``$PATH``, you might need to set
+   ``$JAVA_HOME`` accordingly)
 4. Python version 3
 
 ## Build instructions
--- a/README.md
+++ b/README.md
@@ -20,20 +20,62 @@ Language mappings are defined in `.tx/config`, with the second code being the on
 
 # Building
 
-### Dependencies
-- Android SDK, with `$ANDROID_HOME` pointing to it (you can skip this if you are using Android Studio)
-- Android NDK (you should install the required version (`ext.ndkVersionShared` in `./build.gradle`) with the usual tools, such that it's located within `$ANDROID_HOME/ndk/`).
-- Go (see [here](https://docs.syncthing.net/dev/building.html#prerequisites) for the required version)
-- Java Version 11 (you scan skip this if you are using Android Studio, otherwise you might need to set `$JAVA_HOME` accordingly)
+## Dependencies
 
-### Build instructions
+These are necessary for building from the command line. If you build using
+Docker, you don't need to set them up separately.
 
-Make sure you clone the project with
-`git clone https://github.com/syncthing/syncthing-android.git --recursive`. Alternatively, run
-`git submodule init && git submodule update` in the project folder.
+1. Android SDK and NDK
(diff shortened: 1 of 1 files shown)


tomasz1986:
I've updated the PR to match the current state of the code. Please review when you've got some time 🙂.

Just a comment about Java, as the original instructions talked about `$JAVA_HOME` home. I don't think `$JAVA_HOME` is required, just that Java has to be installed and present in `$PATH`.

README.md
```diff
@@ -20,20 +20,62 @@ Language mappings are defined in `.tx/config`, with the second code being the on
...
+       folder hierarchy looks as follows.
+       ```
+       cmdline-tools/latest/bin
+       cmdline-tools/latest/lib
+       cmdline-tools/latest/source.properties
+       cmdline-tools/latest/NOTICE.txt
+       ```
+    4. Navigate inside `cmdline-tools/latest/bin`, then execute
+       ```
+       ./sdkmanager "platform-tools" "build-tools;31.0.0" "platforms;android-31" "extras;android;m2repository" "ndk;25.1.8937393"
```
imsodin, line 43: I don't like having yet another place that these versions need to be updated - I'd prefer just referencing the dockerfile (and stubbing out the versions here).

README.md
```diff
@@ -20,20 +20,62 @@ Language mappings are defined in `.tx/config`, with the second code being the on
...
+        **NOTE:** You can double-check [Dockerfile](docker/Dockerfile) to make
+        sure that the version numbers are up-to-date.
+2. Go (see https://docs.syncthing.net/dev/building#prerequisites for the
+   required version)
+3. Java version 11
+4. Python version 3
+
+## Build instructions
+
+### Command line
```
imsodin, line 56: There's no other build instructions, so could just drop this sub-header.

README.md
```diff
@@ -20,20 +20,62 @@ Language mappings are defined in `.tx/config`, with the second code being the on
...
-### Dependencies
-- Android SDK, with `$ANDROID_HOME` pointing to it (you can skip this if you are using Android Studio)
-- Android NDK (you should install the required version (`ext.ndkVersionShared` in `./build.gradle`) with the usual tools, such that it's located within `$ANDROID_HOME/ndk/`).
-- Go (see [here](https://docs.syncthing.net/dev/building.html#prerequisites) for the required version)
-- Java Version 11 (you scan skip this if you are using Android Studio, otherwise you might need to set `$JAVA_HOME` accordingly)
+## Dependencies
 
-### Build instructions
+These are necessary for building from the command line. If you build using
+Docker, you don't need to set them up separately.
```
imsodin, line 26: Should probably be moved directly below `# Building`, and also mention using android studio as an option. In both cases all the sections aren't relevant.

imsodin (commented):
Thanks!

I'd prefer to keep the `JAVA_HOME` part - it's not necessary if it's the only java installation in the default place the platform is looking for. And that's not a given and setting it to the correct java location can't hurt.

README.md
tomasz1986: Done.

README.md
tomasz1986: Done.

README.md
tomasz1986: Done.

tomasz1986:
Thanks for having a look at this and please check the current iteration 🙂.

imsodin (approved):

PR #1866, opened 2022-11-27
Use local dir state for prebuilding in Dockerfile

imsodin:
Currently the script checks out the main branch from the repo. Meaning you can never test local changes or build from a different branch. Use the local dir instead copying that into the docker image instead of checking out from the repo.

This also requires a small change to CI: Change docker context from `./docker` to `./` (repo root) - I am testing that right now by building this branch (worked fine locally).

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,13 +20,18 @@ Language mappings are defined in `.tx/config`, with the second code being the on
 
 # Building
 
-### Dependencies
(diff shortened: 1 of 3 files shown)


imsodin:
Teamcity is being clever with git. Working around that cleverness isn't easy -> keep the current main only way.

PR #1916, opened 2023-06-13
Fix contrast of top app bar

### LINKED ISSUE #1915: Latest update makes headers nearly unreadable when using dark mode
Please see the top of this screenshot:
Screenshot_20230613-012529297

Previously, the header text was not white-on-light. (EDITED.)

### Version Information

    App Version: 1.23.5
    Syncthing Version: 1.23.5
    Android Version: Android 13

Bnyro:
closes #1915

### DIFF
--- a/app/src/main/res/drawable/ic_add_24dp.xml
+++ b/app/src/main/res/drawable/ic_add_24dp.xml
@@ -1,6 +1,7 @@
 <vector xmlns:android="http://schemas.android.com/apk/res/android"
         android:width="24dp"
         android:height="24dp"
+        android:tint="?colorControlNormal"
         android:viewportWidth="24.0"
         android:viewportHeight="24.0">
     <path
--- a/app/src/main/res/layout/activity_main.xml
+++ b/app/src/main/res/layout/activity_main.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <androidx.drawerlayout.widget.DrawerLayout xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:app="http://schemas.android.com/apk/res-auto"
     android:id="@+id/drawer_layout"
     android:layout_width="match_parent"
     android:layout_height="match_parent">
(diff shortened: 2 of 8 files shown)


imsodin:
Thanks a lot for the super speedy reaction!  
Would you mind posting a screenshot?

Bnyro:
Unfortunately, I currently don't have a development setup for syncthing, so the app crashes because the go code is not included. However, I can make screenshots from the startup. 
I just noticed there are issues in light mode now, I'll force push the branch in a sec and send screenshots then.

Bnyro:
Screenshot_20230613-104734_Syncthing
Screenshot_20230613-104742_Syncthing

Bnyro:
I had to make some other adjustments now to fix the icon colors in the top bar and prevent different elevations of the toolbar and the tab layout.

Bnyro:
Yes, we don't apply any custom style to dialogs, we use the stock ones from Material Design 3. The same behavior is visible in other apps using Material Design 3.
(That's why I always use the dark mode, it just looks way better :p)

imsodin (approved):

LinuxOnTheDesktop:
It seems to me - if I may say so - that in both screenshots above there should be a little more contrast.

Also - if this is a separate point - in the second screenshot, the background to the rounded box, and/or the text within that box, are a little too yellow (as if some sort of eyecare filter has been applied; such filters should be left to the OS or to the user).

Still, the dark mode is now _legible_, which is the main thing!

imsodin:
As far as I understand we now don't apply any custom theming, but use some material dark google pre-defined stuff. Which I like on one hand, as it's less stuff to maintain and get wrong, and dislike on the other hand, for (some of) the reasons you outline. However the maintenance part wins big time - so yeah, maybe you can convince google to stop that low-contrast, flatness craze (or it will blow over eventually) :)

LinuxOnTheDesktop:
@imsodin: thanks. It's no problem. I'll just have a word with Ms Google - via her search bar - and ask her to improve things. Give it a day or two. :)

PR #1917, opened 2023-06-13
Remove incorrect assumption that event data is always a map (fixes #1884)

### LINKED ISSUE #1884: Crashes due to syncthing event parsing failure
Hi,

Not sure when this started, but syncthing no longer runs on my pixel 4a running android 13. I have the 4th of Jan version of syncthing-android installed from google play.

The app starts, and my folders start to scan and sync, but after about 10 seconds the app just disappears and the "syncthing is running" notification goes away too. I can restart syncthing, but it always dies shortly after.

I can't run the app long enough to prod around in settings or look at logs. I've tried clearing the cache, no joy.

I have 3 folders shared, about 10,000 files in total. Sharing to OpenBSD nodes.

Is there anything I can try? Thanks

### Version Information

    App Version: 1.23.0
    Syncthing Version: 1.22.3
    Android Version: Stock Android 13

imsodin:


imsodin:
Maybe you could have a look, @AudriusButkevicius ?

calmh:
Maybe we should conform to this assumption on the Syncthing side as well, though. It would make things simpler. I took a quick glance and I think the main culprit is the `Failure` event which uses just a string, and which we could easily change to be an object type.

imsodin:
Maybe, probably...
  
However I'd like to stop needlessly crashing for quite a few users asap - that's why I am hoping to merge this and release it in the current RC. So maybe @calmh or @Catfriend1 can give me a review on this regardless, please? :) 

And this isn't a new thing either, so I'd expect most active consumers to already have adjusted to it. Of course we can still narrow the type in syncthing, as it's a backwards compatible change, I just don't see it as particularly important.

calmh (approved):
LGTM, except for a whole bunch of Java

imsodin:
> LGTM, except for a whole bunch of Java

Thank you - and couldn't agree more!

PR #1939, opened 2023-07-15
feat: :sparkles: add syncthing uri

LBF38:
Hi there, I'm currently working on a [Syncthing integration for Obsidian](https://github.com/LBF38/obsidian-syncthing-integration) and I am having trouble to launch the Syncthing app from within [Obsidian](https://obsidian.md/).

Therefore, I would love to add some syncthing URIs to launch certain activities of the app.

In this PR, I'm adding a URI to open the app from the URI `syncthing://open`.
If possible, it would be awesome to also have `syncthing://start` and `syncthing://stop` to be able to start or stop Syncthing services from the Obsidian plugin I'm developing.

Purpose of this PR : 
- add a `syncthing://open` uri to open the syncthing app from another app.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -14,11 +14,11 @@
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
     <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="android.permission.READ_SYNC_SETTINGS" />
-    <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS"/>
+    <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" />
     <!-- ACCESS_COARSE_LOCATION is required to get WiFi's SSID on 8.1 -->
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
     <!-- FOREGROUND_SERVICE is required to create a foreground service on 9 -->
-    <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
+    <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <!-- ACCESS_FINE_LOCATION is required to get WiFi's SSID on 10 "Q" -->
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <!-- ACCESS_BACKGROUND_LOCATION is required to get WiFi's SSID on 10 "Q" -->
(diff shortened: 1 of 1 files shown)


imsodin:
Why do you need to add a data object to open the syncthing app with this intent? It should already allow doing that as is as far as I understand.

LBF38:
> Why do you need to add a data object to open the syncthing app with this intent? It should already allow doing that as is as far as I understand.

Oh my bad, then how could I do it ? 

I've tried launching it programmatically by using the package name or other things but didn't make it. Most of it might be concerning the Obsidian app, it might not allow to launch another app. However, I'm able to launch other apps via URIs or specific URLs like `github://...` will open the GitHub app, and `https://play.store.app/...` will open the browser or redirect to the app page on PlayStore.

These are the workaround I've found to make it happen. If you have a better way to do it, I'll be happy to know more about.

LBF38:
Moreover, I think that the URI support could be useful for launching the Syncthing app in the background.
For example, the user would launch the Obsidian app and, if toggle in the plugin's settings, the Syncthing service/app would automatically start in the background for syncing files and folders.

AudriusButkevicius:
Seems like intents should be enough for that?

LBF38:
> Seems like intents should be enough for that?

Then, could you show me a code snippet to do that please ? 

As the Obsidian app is using the [CapacitorJS lib](https://capacitorjs.com/) (see the [Obsidian's forum post](https://forum.obsidian.md/t/what-technology-obsidian-mobile-is-developed-with/40125)), I don't know how to make it happen or how to call an intent from the plugin.

acolomb:
Have you seen this?
https://github.com/Catfriend1/syncthing-android/wiki/Remote-Control-by-Broadcast-Intents
It should work just the same in the official app AFAIK.

LBF38:
> Have you seen this? https://github.com/Catfriend1/syncthing-android/wiki/Remote-Control-by-Broadcast-Intents It should work just the same in the official app AFAIK.

Thanks for the reference. I haven't seen it. Well, I'll try it again using the method but, the last tests I've made with intents in Obsidian, didn't work at all.

imsodin:
Pretty sure adding a restriction to the intent won't help with running it - if it's not working as is something is wrong with the state now (either on your apps side or ours).

LBF38:
Hi, thanks for your comments, maybe the URI isn't the best solution, that's how I imagined it to be the simplest way to call the app. 

Obsidian is using [CapacitorJS](https://capacitorjs.com/) on the mobile app, you might know how to launch an app from there ?

Another way to interact with the Syncthing instance is through the web browser, at the `https://localhost:8384` address. However, it is protected for security purpose. Would it be possible to use it for knowing if the Syncthing instance is active ?

If you have code examples or snippets that I could use to make it work, I'll thank you a lot.

acolomb:
Just to check if Syncthing is running should be easy with that URL. You either get an HTTP 401 error response or you get a connection abort because nothing is listening on that port.

LBF38:
@acolomb that could do the trick, and I also should be able to interact with the Syncthing instance with the API Key no ? Like on desktop ?

LBF38:
Another question that comes to mind is why does the syncthing uri could not be the easiest solution ? I mean, does it add some security issue or something unwanted ?

imsodin:
There's no data to be acted upon in this intent, I don't see the point of adding that. I don't pretend to understand the ins and outs of intents and their attributes, but there weren't any pointers or explanations in this PR that would signify otherwise. Also what acolomb said regarding features -> closing this. Please use the forum - https://forum.syncthing.net - for further discussions/help with that integration.

PR #1940, opened 2023-07-16
Publish as bundle to google play

imsodin:


.github/workflows/build-app.yaml
```diff
@@ -44,7 +44,7 @@ jobs:
 
   release:
     name: Release Build and Publish
-    if: github.event_name == 'push'
+    if: github.event_name == 'push' || github.event_name == 'pull_request'
```
AudriusButkevicius, line 47: Publish to play on pull request?

imsodin: This is just so I can debug on the PR and don't need to keep committing to main/release - will be gone before merge.

PR #1949, opened 2023-07-20
Use normal color attrib. instead of white (fixes #1947)

### LINKED ISSUE #1947: Android will not Add Device.

Pr0pHesyer:
fix #1947


imsodin (approved):
Thank you

PR #1962, opened 2023-08-16
Remove non-functional restartOnWakeup from app settings (fixes #1961)

### LINKED ISSUE #1961: Settings | Syncthing Options | Restart on Wakeup
This option apparently defaults to true, but as long as I've used Syncthing it has never stayed checked when I have explicitly checked it, and has never been checked, as is the default, on a new installation or when I've returned to this page after checking this box on.

Please indicate that it is on by marking the option checked, if it on or unchecked If it is off.

The reason for logging this issue is that I am not convinced that it is ever on, or it is switched off as you leave this page..

If your issue is a bug report, replace this boilerplate with a description of the problem, being sure to include at least:

Restate:
* What happened:  checkbix off always when you open this settings page item whether you check the box before leaving or not. Sb checked as the default is on.
* What you expected to happen instead:
-  the checkbox defaults to on and is checkmarked on when it is on.
- please confirm that this option does what it says; restarts Syncthing when the mobile is woken.
* any steps to reproduce the problem.: Observe that the checkbox is never checked regardless of manual attempts to check when returning to that options page.
 - unable to confirm that the Syncthing is restarted when the mobile is allowed to time out and awoken.

### Version Information

    App Version: 1.23.7
    Syncthing Version: 1.23.7
    Android Version: Android 13. Security update 5 August 23 
Kernel 5.10.157-android13-4-00003-g776d0a76f6aa-ab10208116
#1 Thu May 25 12:11:12 UTC 2023

tomasz1986:
Remove non-functional restartOnWakeup from app settings (fixes #1961)

The option "restartOnWakeup" was removed in Syncthing v1.21.0 [1]. Thus,
remove it from the Android app as well, since the option does not
perform any function anymore. In addition, update the Docs also [2].

[1] https://github.com/syncthing/syncthing/issues/8448
[2] https://github.com/syncthing/docs/pull/815

Signed-off-by: Tomasz Wilczyński <twilczynski@naver.com>

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/SettingsActivity.java
@@ -126,7 +126,6 @@ public class SettingsActivity extends SyncthingActivity {
         private CheckBoxPreference mRelaysEnabled;
         private EditTextPreference mGlobalAnnounceServers;
         private EditTextPreference mAddress;
-        private CheckBoxPreference mRestartOnWakeup;
         private CheckBoxPreference mUrAccepted;
 
         private Preference mCategoryBackup;
@@ -214,7 +213,6 @@ public class SettingsActivity extends SyncthingActivity {
             mRelaysEnabled          = (CheckBoxPreference) findPreference("relaysEnabled");
             mGlobalAnnounceServers  = (EditTextPreference) findPreference("globalAnnounceServers");
             mAddress                = (EditTextPreference) findPreference("address");
-            mRestartOnWakeup        = (CheckBoxPreference) findPreference("restartOnWakeup");
             mUrAccepted             = (CheckBoxPreference) findPreference("urAccepted");
 
             mCategoryBackup         = findPreference("category_backup");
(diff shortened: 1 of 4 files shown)


tomasz1986:
I've never really done a proper pull request to the Android app repo before, but when it comes to translations, is it enough to remove the English strings only (which I'm doing here)? Will Transifex take care of the other languages automatically?

imsodin:
Thanks for taking care of this.

Maybe a tx push/pull cycle would resolve this, however the following command works definitely: `sed -i -e '/restart_on_wakeup/d' app/src/main/res/values*/strings.xml`. Could you run and push that to make the build/CI pass please.

imsodin:
I applied that myself and PRed here - thanks for fixing the actual issue (you'll still be the committer): https://github.com/syncthing/syncthing-android/pull/1968

tomasz1986:
I'm sorry for the delay. I would have come back to this eventually, but I sometimes tend to be rather slow 😳.

imsodin:
Oh no worries at all, I am as well - I just now had time/inclination to finally (as in late) release the latest RC on android, and I wanted to get this in.

Issue #1778, opened 2022-04-14
SyncThing is always disabled.


iwilliamson:
Newly configured setup.  

What happened:
First sync worked, but now SyncThing is showing as "disabled".  When I open the App, I get that my WiFi isn't "whitelisted.  Change Settings?".  Disabling and re-enabling the selected WiFi SSID makes it run, but then it goes to "disabled" again.

Expected operation:
I would expect SyncThing to be running when I'm connected to my WiFi.

Steps to reproduce:
Just be on my Wifi and try and get things to sync.

### Version Information

    App Version: 1.19.2
    Syncthing Version: 1,19.2
    Android Version: 12 - stock.

AudriusButkevicius:
You need to grant the app location permissions most likely. Wifi ssid is assumed to give away location position.

iwilliamson:
Thanks!  

That does seem to have fixed the issue, so I guess the problem is really the error message that the WiFi "isn't whitelisted".  

That's a little misleading.  Maybe the error message could be a little more specific.  Just a thought.

AudriusButkevicius:
I think we explain that when we ask for location permissions.

iwilliamson:
Yeah - I think I may have missed the fact that the service running does count as "While the app is running".  That explains why when I start the actual "app", it seems to work.  

Anyway - thanks for the help!

PR #1974, opened 2023-09-18
Clarified error message for restricted location permissions

zanovis:
This error message is slightly misleading (related to #1778) currently, the user will get this error even if they have whitelisted the SSID.


imsodin:
Makes sense. Can't merge this though at the moment, as our translation system is broken.

Catfriend1:
@imsodin I also have problems with the Transifex api. They changed something but the required changes on my side are unclear. gave up after 2 hours trying to fix it. :-(

imsodin (approved):

acolomb (approved):

PR #1978, opened 2023-09-29
Set a custom build user

bt90:
We need a dedicated build user to be able to distinguish the distribution channel. See https://github.com/syncthing/syncthing/issues/9140


calmh (approved):
LGTM

PR #2000, opened 2023-11-11
Update dependency resolution management

adamszewe:
## Changes
* get rid of the JCenter repository
* centralize repository declaration for all projects in project settings
* refactor `build.gradle` and `settings.gradle` from Groovy syntax to Kotlin DSL (thus replacing those files with `build.gradle.kts` and `settings.gradle.kts` respectively)

The remaining `build.gradle` files are refactored in: https://github.com/syncthing/syncthing-android/pull/2022


imsodin:
What's the reason for switching the style/syntax? Not that I have any strong preferences, but the existing one I have already seen/used and the new one not at all, so I'd like to hear about some incentives for the switch :)

There are also a few other build.gradle files, and I'd like to have a single syntax/style everywhere.

adamszewe:
> What's the reason for switching the style/syntax? Not that I have any strong preferences, but the existing one I have already seen/used and the new one not at all, so I'd like to hear about some incentives for the switch :)
> 
> There are also a few other build.gradle files, and I'd like to have a single syntax/style everywhere.

Those files were using Groovy language and they were refactored to use Kotlin DSL.
The main advantage of using Kotlin instead of Groovy is the IDE support. With Groovy it is extremely limited, since it is dynamically typed, while Kotlin is statically typed.
Kotlin is also the default language when it comes to developing Android applications, thus using the same language for both build logic and the app itself is also quite helpful.
See also: [Kotlin DSL is Now the Default for New Gradle Builds](https://blog.gradle.org/kotlin-dsl-is-now-the-default-for-new-gradle-builds).

I'd happily refactor the other remaining `build.gradle` files from Groovy to Kotlin 😄

imsodin:
Thanks for explaining - this being the new default is particularly convincing to me :)

Please do change all of them, I'd like to use one style/DSL only throughout the project.

adamszewe:
I opened a new PR that completes this refactor: https://github.com/syncthing/syncthing-android/pull/2022
I preferred to open another one to make it easier to review.

adamszewe:
@imsodin I rebased the PR to get latest updates and this PR is ready to review

build.gradle.kts
```diff
@@ -1,32 +1,27 @@
...
+        set("ndkVersionShared", "25.2.9519653")
+
     }
+
     repositories {
+        gradlePluginPortal()
         google()
-        jcenter()
+        mavenCentral()
     }
```
imsodin, line 15: Just to double check: This block needs to be present both here and in settings.gradle.kts?

adamszewe: Yes, it still has to be present because the `buildscript` block is used. 
If it's removed the build fails as it cannot resolve those dependencies in the `buildscript`.

image

It should be possible to get rid of this, I'll try to do that in a future PR.

imsodin (approved):

build.gradle.kts
imsodin: Oh no worries, just wanted to make sure this wasn't a simple oversight

PR #2002, opened 2023-11-12
Use StrictMode VmPolicy only in debug build type

adamszewe:
## Description 
`StrictMode` makes sense mainly in for debug build types, enabling it for release does not provide any value as it can add additional overhead and its logs going to be removed (because R8 strips them).

## Changes
* Enable the StrictMode VM policy only for debug build type

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/SyncthingApp.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/SyncthingApp.java
@@ -25,6 +25,12 @@ public class SyncthingApp extends Application {
 
         new Languages(this).setLanguage(this);
 
+        if (BuildConfig.DEBUG) {
+            setStrictMode();
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


imsodin:
I am not not that convinced by the benefits of of the content provider. Aren't we abusing a it, as it's intended to do something else? And those other checks you mentioned, couldn't we do that in `Application` as well?

adamszewe:
> I am not not that convinced by the benefits of of the content provider. Aren't we abusing a it, as it's intended to do something else? And those other checks you mentioned, couldn't we do that in `Application` as well?

Hello 👋 
Using a `ContentProvider` is a "trick" other libraries, like Firebase uses among many, use.
I think I over-engineered a bit the solution initially thus I rewrote the PR and just moved the code that sets the VMPolicy to a dedicated function which is invoked only when the build type is `debug`.

imsodin (approved):

PR #2004, opened 2023-11-12
Remove outdated drawer entry for CPU usage.

acolomb:
The code to calculate this metric has long been removed from the backend upstream, in commit https://github.com/syncthing/syncthing/commit/c7d6a6d78042876d9a27b13df2ae45783dca7c37.

Thus the GUI element should go away as well, currently always displaying 0 percent anyways.

Ref https://github.com/syncthing/syncthing/pull/6393

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/fragments/DrawerFragment.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/fragments/DrawerFragment.java
@@ -43,7 +43,6 @@ public class DrawerFragment extends Fragment implements View.OnClickListener {
 
     private static final String TAG = "DrawerFragment";
 
-    private TextView mCpuUsage;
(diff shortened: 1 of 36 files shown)


imsodin:
The ram next to the cpu stuff was also removed, right? Could you remove both at once?

tomasz1986:
> The ram next to the cpu stuff was also removed, right? Could you remove both at once?

The ram usage reporting still works, it was just hidden from the Web GUI…

acolomb:
Correct, the RAM reporting still works.  I was about to remove it as well, but noticed it had valid numbers.  On Android, I think that's a pretty useful piece of information.  In contrast to the CPU stats, it didn't require a whole bunch of code to work and it was deemed more accurate back then.

imsodin (approved):

PR #2020, opened 2023-12-05
i18n: Weblate updates (fixed merge errors)

acolomb:
Unfortunately, because of recent troubles with Weblate's internal Git repository, some changes were lost on the pushed branch.  This current state now contains all changes up-to-date.

Please merge soonish so that Weblate can pick up the updated upstream branch and continue working normally.

### DIFF
--- a/app/src/main/res/values-cs/strings.xml
+++ b/app/src/main/res/values-cs/strings.xml
@@ -223,7 +223,7 @@
     <string name="respect_battery_saving_title">Dodržovat nastavení úspory baterie z Androidu</string>
     <string name="respect_battery_saving_summary">Vypnout Syncthing pokud je aktivní šetření energie.</string>
     <string name="respect_master_sync_title">Dodržovat nastavení automatické synchronizace dat z Androidu</string>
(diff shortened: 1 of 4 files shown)


imsodin:
I hope merging (well fast-forward here) was the right thing to do, and it didn't need a squash. If you have permissions on github, I'd be ok with you fixing up weblate issues directly as well. If you don't have access to do that and think it would help, let me know and I'll see what I need to do to get you the necessary permissions.

acolomb:
Oh dear... Yes, it would really help to have direct push access to the main branch here (which I don't have yet).  Problem is that new translations are coming in continuously, so whenever I try to push merge conflicts into the main branch for Weblate to finally reach a conflict-free state, they often end up causing new conflicts before the PR is reviewed and merged.  All because of an internal Weblate bug which I reported and their attempt to fix it on our repository actually discarded some changes.

I hope when we do reach a clean state again, such troubles will not occur during normal interaction with Weblate.

acolomb:
In the meantime, if you get to it first:  Merge the `weblate` branch into `main` again, there are two more translation commits by now.

imsodin:
You should now have access and I just merged weblate again.

PR #2021, opened 2023-12-05
Upgrade Gradle wrapper to version `8.5`

adamszewe:
## Description
Dependabot does not support (yet) upgrading the Gradle wrapper thus it needs to be updated/upgraded manually.

## Changes:
* Upgrade gradle wrapper to version `8.5`

See also: https://github.com/dependabot/dependabot-core/issues/2223

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -1,3 +1,5 @@
+import org.gradle.configurationcache.extensions.capitalized
+
 plugins {
     id("com.android.application")
     id("com.github.ben-manes.versions")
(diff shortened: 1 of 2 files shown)


app/build.gradle.kts
```diff
@@ -108,3 +108,9 @@ tasks.register<Delete>("deleteUnsupportedPlayTranslations") {
         "src/main/play/listings/ta/",
     )
 }
+
+project.afterEvaluate {
+    tasks.named("mergeReleaseJniLibFolders") {
+        dependsOn(":syncthing:buildNative")
+    }
+}
```
adamszewe, line 116: Newer versions of Gradle check for [implicit dependencies between tasks](https://docs.gradle.org/8.5/userguide/validation_problems.html#implicit_dependency).

With an implicit dependency the build fails as follows:

Screenshot 2024-01-05 at 22 02 06-updated

The above change makes the dependency explicit.

imsodin (approved):

PR #2022, opened 2023-12-06
Refactor groovy gradle files to kotlin dsl

adamszewe:
## Description:
Refactor gradle build files to use Kotlin DSL instead of Groovy.
There were also a bash script and a python script that needed to be updated because they relied on parsing the `build.gradle` files written in Groovy.

This PR completes the work started in: https://github.com/syncthing/syncthing-android/pull/2000


syncthing/build-syncthing.py
```diff
@@ -47,11 +47,11 @@ def fail(message, *args, **kwargs):
...
 
 def get_min_sdk(project_dir):
-    with open(os.path.join(project_dir, 'app', 'build.gradle')) as file_handle:
+    with open(os.path.join(project_dir, 'app', 'build.gradle.kts')) as file_handle:
         for line in file_handle:
             tokens = list(filter(None, line.split()))
-            if len(tokens) == 2 and tokens[0] == 'minSdkVersion':
-                return int(tokens[1])
+            if len(tokens) == 3 and tokens[0] == 'minSdk':
+                return int(tokens[2])
```
adamszewe, line 54: The line that contains the `minSdk` has now three tokens.
Before:
```
minSdkVersion 21
```
Now:
```
minSdk = 21
```

scripts/bump-version.bash
```diff
@@ -3,7 +3,7 @@
 set -e
 
 NEW_VERSION_NAME=$1
```
adamszewe, line 5: awk now needs to print the 3rd element as there is now an equals sign in the middle for `versionName` and `versionCode`, like:
```
        versionCode = 4372
        versionName = "1.27.0-rc.2"
```

app/build.gradle.kts
```diff
@@ -0,0 +1,108 @@
...
+        targetSdk = 33
+        versionCode = 4372
+        versionName = "1.27.0-rc.2"
+        testApplicationId = "com.nutomic.syncthingandroid.test"
+        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+    }
+
+    signingConfigs {
+        create("release") {
+            storeFile = file(System.getenv("SYNCTHING_RELEASE_STORE_FILE") ?: "release.keystore")
```
imsodin, line 52: That was `null` before instead of "release.keystore" - is that equivalent?

app/build.gradle.kts
```diff
@@ -0,0 +1,108 @@
...
+    buildTypes {
+        getByName("debug") {
+            applicationIdSuffix = ".debug"
+            isDebuggable = true
+            isJniDebuggable = true
+            isRenderscriptDebuggable = true
+            isMinifyEnabled = false
+        }
+        getByName("release") {
+            signingConfig = signingConfigs.getByName("release")
```
imsodin, line 68: Previously this conditionally pointed to either the store-file within the release config, or the release config itself. Previous state is not at all self-explanatory, so if you know/think the new way is right I am willing to just take that at face value and try.

imsodin:
That was quite annoying to review with all the lines "changed", thus no usable diff, and needed to be checked that they actually didn't change semantically xD

One question about the signing stuff, otherwise lgtm.

app/build.gradle.kts
adamszewe: you're correct, I referenced my own config while writing this 😓 
I updated the PR to set `storeFile` to `null` instead if the `SYNCTHING_RELEASE_STORE_FILE` var is not set

scripts/bump-version.bash
adamszewe: Bumped to version `1.27.2-rc.2` now

app/build.gradle.kts
adamszewe: I made the wrong assumption that the `signingConfig` should never be null for the `release` build type.
I updated the logic to result in a null `signingConfig` in case there is no `storeFile` available to be the same as in previous config.

adamszewe:
> That was quite annoying to review with all the lines "changed", thus no usable diff, and needed to be checked that they actually didn't change semantically xD
> 

Sorry for any confusion and thanks for the in-depth review 😉 
The python scripts were also a bit of a pain to update. They are a hacky solution and can be converted to dedicated gradle tasks in order to avoid relying on parsing gradle config files manually.

imsodin (approved):

imsodin:
> Sorry for any confusion and thanks for the in-depth review 😉

Oh nothing to be sorry for, I didn't mean to complain about anything you have done. It was simply the very nature of the change that made it hard to review.

PR #2023, opened 2023-12-07
Simplify conditions that check for lollipop, api 21, version

adamszewe:
Lollipop, API 21, has been the min sdk version for over a year in this project.
There were still some conditions in the code, which checked for api 21, that can be removed or simplified.

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java
@@ -294,28 +294,13 @@ public class RunConditionMonitor {
      * Functions for run condition information retrieval.
      */
     private boolean isCharging() {
-        // API level >= 21
-        return isCharging_API17();
-    }
-
-    @TargetApi(16)
-    private boolean isCharging_API16() {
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/nutomic/syncthingandroid/service/RunConditionMonitor.java
```diff
@@ -298,13 +294,8 @@ private boolean wifiWhitelistConditionMet(boolean prefWifiWhitelistEnabled,
...
-            return isCharging_API16();
-        } else {
-            // API level >= 21
-            return isCharging_API17();
-        }
+        // API level >= 21
+        return isCharging_API17();
     }
 
     @TargetApi(16)
```
imsodin, line 301: Looks like there's something that can go too then? :)

adamszewe: 👍 , updated the PR 
I removed the code in the method below (annotated with TargetApi 16) and moved the the code from `isCharging_API17()` method directly to `isCharging()` one.

imsodin (approved):
Thanks!

PR #2036, opened 2024-01-05
Remove PRNGFixes file

adamszewe:
The security workarounds contained in the `PRNGFixes` class were needed for devices older Android APIs (16, 17, 18) while the current min sdk API is 21.
Therefore this workaround is no longer needed.


app/src/main/java/com/android/PRNGFixes.java
```diff
@@ -1,338 +0,0 @@
...
-
-    /**
-     * Applies all fixes.
-     *
-     * @throws SecurityException if a fix is needed but could not be applied.
-     */
-    public static void apply() {
-        applyOpenSSLFix();
-        installLinuxPRNGSecureRandom();
-    }
```
adamszewe, line 59: The first method checks whether the SDK version is lower than 16 or higher than 18.
The second method checks only whether it's higher than 18.
The current min sdk version for the app is 21, thus the rest of the class is never executed.

imsodin (approved):

PR #2037, opened 2024-01-05
Use view binding

adamszewe:
Enable the [view binding feature](https://developer.android.com/topic/libraries/view-binding#kts) and replace explicit view binding with `binding.xyz` in `DeviceActivity`.


app/src/main/java/com/nutomic/syncthingandroid/activities/DeviceActivity.java
```diff
@@ -65,27 +64,7 @@ public class DeviceActivity extends SyncthingActivity implements View.OnClickLis
 
     private Device mDevice;
 
-    private View mIdContainer;
```
adamszewe, line 68: explicit declaration for each view reference is no longer needed

imsodin (approved):

PR #2040, opened 2024-01-13
Fix incorrect initialization of Info notification channel

the-eclectic-dyslexic:
I was using syncthing as a reference for handling notifications in my app, when I noticed that the "Info" channel appears to be incorrectly initialized.

It looks like it was intended for the Info channel not to vibrate, not to make sound, but only to show a badge. This might be redundant as it's possible every notification pushed to this channel already follows those standards, I am unsure. However, the code used to initialize the Info channel was actually modifying the already created Persistent channel, and thus did nothing. 

This pull request fixes, what I think are typos which originated from a copy and paste of the code used to create the Persistent channel.

I think these lines should either be changed as per this pull request, or removed entirely.


imsodin (approved):
That indeed looks like a copy-paste mistake - thanks!

PR #2049, opened 2024-02-04
Upgrade compile sdk and tools

adamszewe:
Upgrade compile sdk and tools from 33 to 34

This is needed to update Android Gradle Plugin (AGP) and other dependencies.


imsodin (approved):
I usually update this together with the target SDK. However everything I read indicates that it's fine to just update compile sdk, i.e. none of the runtime behaviour changes that e.g. come with bumping the target. Lets hope that's the case :)

Issue #2062, opened 2024-02-23
"Override Changes" button MUST WARN PEOPLE WHAT IT DOES.


ZacharyACoon:
The "Override Changes" button in android provides zero information about what it does. 

Override WHAT?
Override WHERE?
What files are being changed how?

This is EXTREMELY dangerous and just nearly cost my partner years of baby photos, I'm shaking.

tomasz1986:
Duplicate of https://github.com/syncthing/syncthing-android/issues/1769.

We are aware that this can pose a serious usability issue, as a single miss click can lead to file deletion on the other side, but the app is in maintenance mode, so unless a volunteer steps in and improves the experience, the current state likely won't change in the near future…

There is a confirmation dialog in the Web GUI, but not the app itself.

PR #2063, opened 2024-02-26
Greatly improve "Override Changes" button and UX.

### LINKED ISSUE #1769: Use 'Review and Override' instead of the imperative 'Override Changes'
(After an Android 11 forced reset of the DB), I see this:

Screenshot_20220307-190054_Syncthing_unclear_button__shadow

I don't dare press it, because it's written in the [imperative mood](https://en.wikipedia.org/wiki/Imperative_mood). I want to do know what it's going to do. Is there a review screen? If so, call the button 'Review and Override'.

ZacharyACoon:
Fixes #1769, #2062

# Description
The override changes button does not clearly indicate its function, and performs a very dangerous action that by design wipes files across the entire cluster.  It has caused numerous users to lose files unexpectedly and scared the hell out of me the other day, so I'm fixing it so that no user can be confused as to what it will do.

# Changes
- Renames "Override Changes" to "Override Remotes"
- Moves button to folder settings to avoid annoying users and accidental presses.
- Provides detailed description of function before pressing.
- Provides serious warning of danger and requires confirmation when pressed.
- Button fits into Syncthing's theme nicely rather than looking weird in the main screen.

# Screenshots

| Before | After |
| ------ | ----- |
|| (Not enabled) Screenshot_20240226-032518 |
| Screenshot_20240226-034947 | Screenshot_20240226-032540 |
||(Enabled) Screenshot_20240226-032545|
||Screenshot_20240226-032549|


tomasz1986:
Can you change the look and wording to match what is used in the Web GUI?

image

app/src/main/java/com/nutomic/syncthingandroid/activities/DeviceActivity.java
```diff
@@ -57,7 +57,7 @@ public class DeviceActivity extends SyncthingActivity implements View.OnClickLis
     private static final String TAG = "DeviceSettingsFragment";
     private static final String IS_SHOWING_DISCARD_DIALOG = "DISCARD_FOLDER_DIALOG_STATE";
     private static final String IS_SHOWING_COMPRESSION_DIALOG = "COMPRESSION_FOLDER_DIALOG_STATE";
-    private static final String IS_SHOWING_DELETE_DIALOG = "DELETE_FOLDER_DIALOG_STATE";
+    private static final String IS_SHOW_DELETE_DIALOG = "DELETE_FOLDER_DIALOG_STATE";
```
tomasz1986, line 60: Why change the name? If not absolutely necessarily, I would suggest to limit code changes to minimum, so that the actual improvements can be reviewed more easily.

app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
```diff
@@ -304,12 +324,18 @@ public void onPause() {
     @Override
     protected void onSaveInstanceState(Bundle outState) {
         super.onSaveInstanceState(outState);
-        outState.putBoolean(IS_SHOWING_DELETE_DIALOG, mDeleteDialog != null && mDeleteDialog.isShowing());
+        outState.putBoolean(IS_SHOW_DELETE_DIALOG, mDeleteDialog != null && mDeleteDialog.isShowing());
         Util.dismissDialogSafe(mDeleteDialog, this);
 
+        outState.putBoolean(IS_SHOW_OVERRIDE_REMOTES_DIALOG, mOverrideChangesDialog != null && mOverrideChangesDialog.isShowing());
```
tomasz1986, line 330:
```suggestion
        outState.putBoolean(IS_SHOW_OVERRIDE_CHANGES_DIALOG, mOverrideChangesDialog != null && mOverrideChangesDialog.isShowing());
```

Please change in all places. If unclear, it could also be `OVERRIDE_REMOTE_CHANGES` (but not just `REMOTES`).

app/src/main/res/values-fr/strings.xml
```diff
@@ -74,7 +74,7 @@
         <item quantity="many">%1$d / %2$d Fichiers</item>
         <item quantity="other">%1$d / %2$d Fichiers</item>
     </plurals>
-    <string name="override_changes">Ignorer les modifications</string>
+    <string name="override_changes">Écraser les changements distants</string>
```
tomasz1986, line 77: Translations are handled at https://hosted.weblate.org/projects/syncthing/android, so I would suggest to remove the change from here and fix the text there.

app/src/main/res/values-fr/strings.xml
```diff
@@ -553,4 +553,4 @@
     <string name="api_level_30_button">Réinitialiser la base de données</string>
     <string name="notification_permission_title">Permission de notification</string>
     <string name="require_notification_permission_desc">Nécessaire pour afficher les notifications concernant les nouvelles connexions d\'appareils, les demandes de partage de dossiers, etc.</string>
-</resources>
\ No newline at end of file
+</resources>
```
tomasz1986, line 556: Please remove unrelated newline changes.

app/src/main/res/values/strings.xml
```diff
@@ -116,7 +116,9 @@
         <item quantity="other">%1$d / %2$d Files</item>
     </plurals>
 
-    <string name="override_changes">Override changes</string>
+    <string name="override_changes">Override Changes</string>
+    <string name="override_changes_description">When configured as \"send only\" and changes have been made remotely, you may click here to overwrite the folder content on other devices.</string>
```
tomasz1986, line 120:
```suggestion
    <string name="override_changes_description">When configured as \"Send Only\" and changes have been made remotely, you may click here to overwrite the folder content on other devices.</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -116,7 +116,9 @@
...
+    <string name="override_changes_are_you_sure"><b>Warning!</b>\n\nThe folder content on other devices will be overwritten to become identical with this device.  Files not present here will be deleted on other devices.\n\nAre you sure you want to continue?</string>
```
tomasz1986, line 121:
```suggestion
    <string name="override_changes_are_you_sure"><b>Warning!</b>\n\nThe folder content on other devices will be overwritten to become identical with this device. Files not present here will be deleted on other devices.\n\nAre you sure you want to continue?</string>
```

ZacharyACoon: I made these changes in the latest force push to this PR. :)

app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
ZacharyACoon: > Please change in all places. If unclear, it could also be `OVERRIDE_REMOTE_CHANGES` (but not just `REMOTES`).

I missed these, reverted them as well in the latest push.

tomasz1986:
> > Thank you for working on this. I'm going do to some testing on a real device soon.
> 
> Bump, is there anything I can do to help further? I'm hoping to see this!

No need to bump 😅. When it comes to Syncthing, PR reviews can take a while. I promised to give this a try on a real device, and I will do this, but "soon" for me is more like in 1-2 weeks when I manage to find some free time.

Also, I can't really review the Java code itself. Someone more knowledgeable would need to have a look at that part.

imsodin:
Thanks a lot for tackling this - it's easily the most important thing to be improved in this app!

Some high-level points first:

The confirmation dialog is definitely needed and great!

My initial reaction to moving the button into the edit dialog isn't entirely positive. Feels a bit surprising as it's not an operation to edit the folder settings. And in the syncthing web UI the button is on the folder overview (which admittedly is more extensive). So not sure a user would find the button? Then again the dialog opens by just tapping the folder, there's no indication of that being only for settings, so maybe it's fine.  
I assume you clearly prefer it inside the dialog. @tomasz1986 any input/opinion on that?  
Just to be clear, I am not asking you to change it at this point, just bringing it up.

I also first wanted to ask you to make it look more like a button. Then I realized there's a bunch of "buttons"/menu opening entries there and they all look flat. I guess that's just what's considered normal/good under material design... If you have a good idea how to improve it, that'd be great, but if not it's fine as is clearly.

Regarding review "speed": Sorry for the long wait. And after review it might be something between a long while and never until this reaches the majority of our users due to google, see https://github.com/syncthing/syncthing-android/issues/2064

ZacharyACoon:
I actually started by keeping the original button and just adding a
warning, it was certainly "easier".

I thought a little while about how I feel general user experience might
best be represented, and I came to the conclusion that the existing folder
actions interface in the web gui and android UI feels a bit out of place.

---

I do strongly prefer to consider what I am labeling "folder actions" a
subset of modifying the folder settings.

If that feels confusing, we might relabel "Folder Settings" to folder
details.

We could easily add a switch/checkbox to keep/reenable the original
override button.

---

Also.  I'm happy to pull request this change in the Web UI as well.  I
don't like the button being so prominent there either.

I've installed Android Studio now, so I'm happy to make more modifications.

app/src/main/java/com/nutomic/syncthingandroid/fragments/FolderListFragment.java
```diff
@@ -101,9 +102,15 @@ private void updateList() {
 
     @Override
     public void onItemClick(AdapterView<?> adapterView, View view, int i, long l) {
+        Folder folder = mAdapter.getItem(i);
+        FolderStatus folderStatus = mAdapter.mLocalFolderStatuses.get(folder.id);
+        boolean overridableChanges = folder.type.equals(Constants.FOLDER_TYPE_SEND_ONLY)
+            && folderStatus.isNeedsItems()
+            && folderStatus.isOutOfSync();
```
imsodin, line 109: The previous condition checked for folder idleness. That's because you can't revert if a scan or sync operation is in progress. This is a bit of a problem with the menu, as the time when you open the menu is not necessarily the time the you click override. So the enabledness of the button might not match reality. Which means the operation will not work (likely hang and time out eventually).

imsodin: Except I just saw the condition for `isOutOfSync` further down - it checks both `isNeedsItems` and for idleness. So the `isNeedsItems` check here is redundant. 

The timing problem in the menu still is a thing. Ideally that would somehow be hooked to folder status, but that's probably not super simple? I may be able to be convinced to just close my eyes and pretend I didn't know about this.

app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
```diff
@@ -40,7 +40,7 @@ public class FoldersAdapter extends ArrayAdapter<Folder> {
 
     private static final String TAG = "FoldersAdapter";
 
-    private final HashMap<String, FolderStatus> mLocalFolderStatuses = new HashMap<>();
+    public final HashMap<String, FolderStatus> mLocalFolderStatuses = new HashMap<>();
```
imsodin, line 43: The `m` prefix is used for private members. I think the javay thing to do is to keep it private and add a getter. Maybe a useful interface would be `getLocalFolderStatus(String folderId)` as the use in this PR only cares about a single folder.

imsodin (commented):
Finally got around to an actual review, sorry for the long wait.

imsodin:
> If that feels confusing, we might relabel "Folder Settings" to folder
details.

That kinda makes sense. Or even just "Folder", or no title at all. I am ok with leaving as is as well though.

> We could easily add a switch/checkbox to keep/reenable the original
override button.

Please not - I don't want more moving pieces for something like this.

> Also.  I'm happy to pull request this change in the Web UI as well.  I
don't like the button being so prominent there either.

You could do that, clearly not required for this PR. And I am not sure in the web UI that would go through. There's a lot more activity there, and opinions involved. So at least I'd start by explaining and proposing to get feedback/consensus, before investing much time into it there.

app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
```diff
@@ -162,26 +168,40 @@ public void onCreate(Bundle savedInstanceState) {
...
             // Prepare edit mode.
             binding.id.clearFocus();
             binding.id.setFocusable(false);
             binding.id.setEnabled(false);
             binding.directoryTextView.setEnabled(false);
+
+            // overridable remotes button
+            setOverrideChangesContainerEnabled(
+                getIntent().hasExtra(EXTRA_FOLDER_OVERRIDABLE_CHANGES)
+                && getIntent().getBooleanExtra(EXTRA_FOLDER_OVERRIDABLE_CHANGES, false)
```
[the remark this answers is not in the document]
ZacharyACoon: I agree, was simply following the existing code... will remove

app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
```diff
@@ -304,12 +324,18 @@ public void onPause() {
...
 
+        outState.putBoolean(IS_SHOW_OVERRIDE_CHANGES_DIALOG, mOverrideChangesDialog != null && mOverrideChangesDialog.isShowing());
+        Util.dismissDialogSafe(mOverrideChangesDialog, this);
+
         if (mIsCreateMode){
             outState.putBoolean(IS_SHOW_DISCARD_DIALOG, mDiscardDialog != null && mDiscardDialog.isShowing());
             Util.dismissDialogSafe(mDiscardDialog, this);
+
+            outState.putBoolean(IS_SHOW_OVERRIDE_CHANGES_DIALOG, mOverrideChangesDialog != null && mOverrideChangesDialog.isShowing());
+            Util.dismissDialogSafe(mOverrideChangesDialog, this);
```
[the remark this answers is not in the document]
ZacharyACoon: I agree, was simply following the existing code... will remove, again.

app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
```diff
@@ -778,4 +804,34 @@ private void setVersioningDescription(String type, String description) {
         binding.versioningType.setText(type);
         binding.versioningDescription.setText(description);
     }
+
+    private void setOverrideChangesContainerEnabled(boolean state) {
+        LinearLayout container = (LinearLayout) findViewById(R.id.overrideChangesContainer);
+        binding.overrideChangesContainer.setEnabled(state);
```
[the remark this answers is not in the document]
ZacharyACoon: I agree, was simply following the existing code... will remove, again, again

app/src/main/java/com/nutomic/syncthingandroid/views/FoldersAdapter.java
ZacharyACoon: implementing that public interface...

app/src/main/java/com/nutomic/syncthingandroid/fragments/FolderListFragment.java
ZacharyACoon: I removed the isNeedsItems as redundant.

app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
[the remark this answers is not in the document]
imsodin: Ah didn't notice that. Just in case: Following existing code is a good argument to me. As in if in the future you intentionally dont change something and don't want to change it, and I miss it's preexisting, feel free to just tell me that without changing anything.

imsodin:
Sorry for updating the branch and potentially interfering with your branch... I only wanted to approve the CI run and misclicked (need to find a way to make that permanent for you, and potentially also find a way to make it stick in general per PR).

app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
```diff
@@ -162,26 +168,39 @@ public void onCreate(Bundle savedInstanceState) {
...
                 showDeleteDialog();
             }
         }
 
         if (savedInstanceState != null){
-            if (savedInstanceState.getBoolean(IS_SHOWING_DELETE_DIALOG)){
+            if (savedInstanceState.getBoolean(IS_SHOW_DELETE_DIALOG)){
                 showDeleteDialog();
             }
         }
```
imsodin, line 196: Entirely optional, just suggesting it as you are renaming the constant which is also a somewhat unrelated cleanup: You could delete one of the two redundant blobs of code.

app/src/main/java/com/nutomic/syncthingandroid/activities/FolderActivity.java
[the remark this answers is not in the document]
imsodin: Just in case as you removed in the other case: It's not removed yet here.  
(Also not that it matters, but here I don't see what the existing code is that is doing the same/something similar.)

app/src/main/java/com/nutomic/syncthingandroid/model/FolderStatus.java
```diff
@@ -27,4 +27,12 @@ public class FolderStatus {
...
     public String error;
     public String watchError;
+
+    public boolean isNeedsItems() {
+        return needFiles + needDirectories + needSymlinks + needDeletes > 0;
+    }
+
+    public boolean isOutOfSync() {
+        return state.equals("idle") && isNeedsItems();
+    }
```
imsodin, line 37: Given you aren't using `isNeedsItems` anymore, I'd prefer if you removed that and folded it into `isOutOfSync`.

imsodin:
Few minor points/leftover from laster review, otherwise this looks good to merge to me.

imsodin:
@ZacharyACoon Do you have time/inclination to address the small remaining review comments above? Otherwise I can also do that myself and then merge - I'd really like to get this in (thanks again).

PR #2072, opened 2024-03-23
Update screenshots for play store listing

realeatham:
# Description
Updated the screenshots that appear in Google play 

# Changes
What changes are made in this PR
* Changed the screenshots for new ones


tomasz1986:
The old ones are outdated, however I think the font here is just too small. Also, the old screenshots were using light mode while these are in dark mode. Android, at least the stock version, by default comes in light mode.

What device has the screenshots been taken on?

realeatham:
My font is at the minimum. I can replace all these with normal font size , and light mode. The device was an S21.

realeatham:
I have updated the screenshots

tomasz1986:
I think they look much better now! There can probably be some objections that it is Samsung's skin and not stock Android, but the current ones are really old and outdated, so updating them is a good thing regardless.

One thing that it may be worth it to check is whether the actual image size and dimensions are fine for Google Play, as they may have some rules about that.

realeatham:
The image meets google play image sizes. The app design isn't changed by the Samsung skin, so it should be fine. Just the navbar is different

imsodin (approved):
Thanks!

PR #2073, opened 2024-04-03
Prefer injecting preferences instead of using PreferenceManager directly

adamszewe:
# Description
Prefer injecting preferences instead of using `PreferenceManager.getDefaultSharedPreferences(context)` directly.

# Changes
* replace `PreferenceManager.getDefaultSharedPreferences(context)` by injecting SharedPreferences instead

### DIFF
--- a/app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
+++ b/app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
@@ -21,7 +21,6 @@ import androidx.core.content.ContextCompat;
 import androidx.viewpager.widget.PagerAdapter;
 import androidx.viewpager.widget.ViewPager;
 
-import android.preference.PreferenceManager;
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/nutomic/syncthingandroid/activities/FirstStartActivity.java
```diff
@@ -240,7 +240,7 @@ private boolean shouldSkipSlide(Slide slide) {
             case API_LEVEL_30:
                 // Skip if running as root, as that circumvents any Android FS restrictions.
                 return upgradedToApiLevel30()
-                        || PreferenceManager.getDefaultSharedPreferences(this).getBoolean(Constants.PREF_USE_ROOT, false);
```
imsodin, line 243: Shouldn't there now be an unused import for `PreferenceManager`? Same on all other files.

imsodin (approved):

PR #2089, opened 2024-07-15
Add Norwegian (nb) translation template for Play listings

acolomb:
Based on user request from Weblate, user `@Njardarheim-1337`.

Looks promising based on the profile:
https://hosted.weblate.org/user/Njardarheim-1337/


imsodin (changes requested):
Google play has a limited list of supported languages/codes, nb-no is not one of them, but simply no: https://support.google.com/googleplay/android-developer/table/4419860?hl=en

Also this suggests that using just "no" is appropriate unless one provides both nn and nb variants. Where nn is not widely used, thus likely we don't want both of them and only using no is appropriate: https://linguistics.stackexchange.com/a/36794

imsodin (approved):

PR #2093, opened 2024-07-29
fix typo

breakTBB:
Description
This PR fixes a typo in the Chinese (Simplified) translation of the app.
Changes

In app/src/main/res/values-zh-rCN/strings.xml, corrected a typo in the api_level_30_desc string:

Changed "速度可能斌满" to "速度可能变慢"
This corrects the phrase from an incorrect character combination to the proper Chinese phrase meaning "speed may slow down"

Screenshots
Not applicable for this change as it only affects text content and not UI layout.


imsodin:
We use weblate for translation, please update the string there: https://github.com/syncthing/syncthing-android?tab=readme-ov-file#translations

PR #2106, opened 2024-09-24
Fix material switch padding on item_device_form layout

Divinelink:
# Description
This pull request addresses the missing padding in the `item_device_form.xml`, which caused the Material Switch to appear too close to the edge of the screen. To resolve this, I added the same padding used by the other Material Switch components through their style. 

Let me know if this works for you!

# Changes
What changes are made in this PR
* Add `android:paddingEnd="@dimen/abc_action_bar_content_inset_material"` on `item_device_form.xml` layout

# Screenshots
In case of UI changes, add before and after screenshots

| Before | After |
| ------ | ----- |
| Screenshot 2024-09-24 at 5 17 00 PM | image |


imsodin:
Thanks for contributing - it's on my radar (and looks good/unproblematic). Just other/unrelated (procedural) things going on at the moment, thus not merging right away.

